PR #33, opened 2016-09-29
generic.MeshScript rewritten to work under Windows

goulu:
Windows does not allow another process (Blender) to open files that are open.
I rewrote generic.MeshScript.__enter__to close files after creation and store file names that are deleted in __exit__

Also had to put a "r" qualifier in blender.py.template, but now trimesh can perform boolean operations under Windows !


goulu:
strange... it goes through my tests https://travis-ci.org/goulu/trimesh but fails 3.4 on yours ...

mikedh:
Thanks! I have no easy way of testing this so apologies if I break it again accidentally.

PR #53, opened 2017-04-24
windows compatible script_text

dajusc:
Dear Michael, :)

when I apply boolean operations (python 2.7.10, scad engine, win10, 64bit), execution fails with message:
> WARNING: Can't open import file '"c:usersschaeferappdatalocal\temp\tmprzu3wa.STL"'.

I figured out, that it is due to the temporary generated scad-script containing single backslashes. This commit fixes the issue by replacing the single backslashes with escaped backslashes in the scad-script .

Successfully tested with scad- and blender-engine on Windows.
Not tested for Linux version, but should also work.

Implementing the fix via "self.replacement['\\'] = '\\\\'" would be cleaner, but unfortunately does not seem to work.

Best regards
David

### DIFF
--- a/trimesh/interfaces/generic.py
+++ b/trimesh/interfaces/generic.py
@@ -38,6 +38,7 @@ class MeshScript:
         self.replacement['script'] = self.script_out.name
 
         script_text = Template(self.script).substitute(self.replacement)
+        script_text = script_text.replace("\\", "\\\\")
         self.script_out.write(script_text.encode('utf-8'))
 
         # close all temporary files


mikedh:
Thanks for the PR, and good find! When I was testing this I noticed that super minor changes to trimesh's interfaces cause it to do crazy, crazy things. 

I wrapped your change in a platform check, and added it to https://github.com/mikedh/trimesh/commit/4ed67b044071026a07dc7240cb868095fc49a9a5. I also added some search paths for openscad. At some point I'll add openscad to the appveyor tests. 

Thanks!

dajusc:
Big thanks to you for this great library! :)

PR #59, opened 2017-06-02
Adds face groups to the mesh when wavefront loaders is used.

hoosha:


mikedh:
Thanks for the PR!

* would you be able to provide an example file that has these keys so I can add it to tests?

* could these 'groups' exist as a metadata property rather than an attribute? it seems like the vast majority of the time (for other file types) it won't be set. Something like:
```
loaded = {'vertices': data[vid].astype(float),
'vertex_normals': data[nid].astype(float),
'faces': faces, 
'metadata': {'face_group' : groups}}
```

hoosha:
Hi 
Sorry for my late response. An example file is attached.
on the second point it would make sense to add it to the meta data as well.
Best,
Hooman

mikedh:
No worries! I don't think that file made it, could you send it directly to my gmail (mik3dh) or link it?

Thanks!
Mike

mikedh:
Thanks for the model! I added it to tests, and those groups properties should be loaded as `mesh.metadata['face_groups']` when available. I included your group loading code in https://github.com/mikedh/trimesh/commit/83aa8f00958a6860a33f084981e69ed82948fdd6

hoosha:
thanks. will fetch your code and test if it works for me in the new way and let you know.
Best,
Hooman

PR #64, opened 2017-07-04
Add export functionality for OBJ files

mmatl:
The functionality here is really similar to the export feature for OFF files. The one exception is that OBJ files are 1-indexed for vertices. I tested this on several OBJ files and it works well.

### DIFF
--- a/trimesh/io/export.py
+++ b/trimesh/io/export.py
@@ -70,6 +70,30 @@ def export_off(mesh):
     return export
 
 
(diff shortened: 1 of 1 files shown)


mikedh:
This looks great, thanks! And so easy! It looks like this exposed a bug with the importer too which is a plus.

mmatl:
No problem! Trimesh is really useful, thanks for open-sourcing it and keeping it maintained.

PR #77, opened 2017-09-06
+ Trimesh.vertex_neighbor, + vertex_adjacency_graph to graph.py, + tests

Marviel:
-added vertex_neighbor function to Trimesh base class and tests
-added vertex_adjacency_graph property to Trimesh base class 
-added vertex_adjacency_graph to graph.py and tests


mikedh:
LGTM, thanks!

PR #86, opened 2017-09-27
Fix bug in the ray tracer

mmatl:
Basically, the ray tracer assumed that the planes given to the function
planes_lines() all intersected the ray in the positive direction.
This assumption was made because planes intersecting the ray in the
negative direction should have been pruned via the rtree. However, if
the ray's origin is too close to any triangle, the triangle won't be
pruned even if intersects the ray in the negative direction.

In order to fix this, this commit adds a new function, planes_rays(),
that performs the check explicitly. The ray tracers now call this
function instead of planes_lines() by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,108 @@
-*.py[cod]
+# Temporary files
 *~
 *.swp
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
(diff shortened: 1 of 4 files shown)


mmatl:
We discovered this bug because we were shooting rays away from the surface of meshes but starting near the surface (i.e. distances of 1e-8 to 1e-12). This patch should fix that without changing any other behavior.

mmatl:
Oh, also just added github's Python .gitignore (helps keep things clean and tidy when developing)

mikedh:
Thanks for the PR! BTW don't worry about the build failing, it looks like freaking networkx 2.0 just got released and changed their return types again, I'll fix it shortly. 

One question before I merge, is this case not handled by https://github.com/mikedh/trimesh/blob/master/trimesh/ray/ray_triangle.py#L224-L226? I added that forward check fairly recently after seeing something similar.

mmatl:
Oh, that looks like it should fix the issue :) Should have read through the most recent commits before issuing the PR, sorry! I'll close it. Good luck with the networkx stuff, changing interfaces is a pain. Also, thanks so much for the library! It's awesome and we're hoping we'll have some stuff to contribute in the coming weeks.

mikedh:
No worries, contributions are greatly appreciated!  Yeah I noticed the same ray problem when I wrote some unit tests that I thought should work but didn't 😛 

I saw you guys had some cool stuff in your branch (like that shiny URDF IO), thanks for taking the time to PR!

mmatl:
No worries! :) I'm currently cleaning up some of that stuff (URDF I/O, quasi-static stable pose analysis, and a few other bells and whistles), and I'll hopefully have some tests and such ready soon. I'll submit a PR for whatever you're interested in mainlining once it's ready!

Issue #82, opened 2017-09-13
Convert voxels to mesh object


Emrys-Lee:
Hi, I have a 3d voxel matrix now. Is it possible to convert it to a mesh object so that I can generate a .ply file? Thank you!

Marviel:
This is an existing function, but it may not perform as you'd wish.

try:
```python
vox = mesh.voxelized(pitch)
voxel_mesh = vox.mesh
voxel_mesh.show()
```
the difference between this and the ideal version is that each voxel column is essentially represented as a single cuboid, rather than the whole voxel volume being a unified single trimesh object.

Marviel:
reference: https://github.com/mikedh/trimesh/blob/master/trimesh/voxel.py

mikedh:
Yeah- that function does exist, and as Luke mentioned you probably don't want it. What you probably want is marching cubes.

Marviel:
@mikedh on that note, I've found that the marching cubes algorithm in scikit's skimage.measure [1] does a pretty good job, once I convert the voxel runs to a binary numpy array.

I'm not sure if you'd be interested in including scikit as a dependency, but this has been working very well for me thus far.

[1] http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.marching_cubes_lewiner

Emrys-Lee:
Thank you. Maybe I'll try some other ways.

PR #87, opened 2017-10-01
add marching cubes mesh generator to Voxel

Marviel:
Does what it says on the tin, addressing concerns of:  #82 

The only question marks here might be the way I installed the scikit-image library -- it is pippable, so I think it should be ok to install the way I did.

Let me know what you think!


Marviel:
It looks like the failures I'm seeing in tests are unrelated import errors

Marviel:
@mikedh if you get a sec, I could use your help understanding why the changes this introduces would cause the massive test failures I'm seeing on Travis

mikedh:
Thanks for the PR! skimage.measure has some cool stuff in there, I was going to benchmark their unique_rows against mine at some point. Pretty nice they did (what looks like) a good marching cubes implementation!

I think the build is failing because travis needs to build some of those dependencies. Since scikit-image is only used in this one function and would add a good number of new dependencies, I would rather do the import inside the marching cubes function so we don't force everyone to install scikit-image. Also it looks like the conda packaging for it is pretty great, so my preference before merging would be to:

* move the scikit-image install from setup.py to the advanced install page
  * looks like the conda-forge builds are quite nice:
  *  `conda install -c conda-forge scikit-image `
* move the import `from skimage import measure` into the top of the `marching_cubes_mesh` function
* add  `conda install -c conda-forge scikit-image ` to the .travis.yml

Marviel:
@mikedh I've tried to execute your wishes, let me know if i've done anything wrong! Stoked to have this easily available :)

mikedh:
Looks great thanks! The build was complaining in early python 3 versions for some reason- in https://github.com/mikedh/trimesh/commit/09a1c349bc0cd98ff7bf45ecd7e5dc53b5413c55 I changed to the marching cubes algorithm their [docs are pushing](http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.marching_cubes_lewiner) and added a level kwarg (maybe its not optional in older versions?). We'll see if the tests can make it through haha.

PR #88, opened 2017-10-02
Initial PR for some of our changes

mmatl:
Feel free to pick and choose what you'd like to include in the mainline :)

Big components here:
* Quasi-static stable pose analysis for meshes (poses.py)
* Bug-Fix for assimp loaders
* Update to native .obj loader to enable reading multiple meshes from one .obj file
* A convex decomposition engine (depends on V-HACD backend being installed)
  * _This one relies on the prior two commits to work cleanly_
* Enable users to manually specify mesh density and center of mass
  * _We find this useful when using meshes in practice_
* URDF exporter (exports convex decomposition approximation of mesh as URDF directory)

There are probably cleaner ways to incorporate the URDF exporter, but I left it as a stand-alone for the moment.


mmatl:
Can also cherry-pick some of these to get rid of that ray tracer bugfix/reversion commit.

mmatl:
Got some CI build errors due to networkx changes, will patch up my code real fast.

mmatl:
I think this is now passing everything -- the error we're seeing in Travis is from a problem with the test_fill_holes test in test_mesh.py, which fails every so often and is unrelated to my changes.

mikedh:
Hey this looks great, thanks! I reran the test that failed and it was fine- gotta love tests with random data :roll_eyes:. It looks like [appveyor found a missing import](https://ci.appveyor.com/project/mikedh/trimesh/build/1.0.214/job/pxn724d0wqayg16t) but I'll fix that in a sec. 

I just went through the PR, and have only minor thoughts, I'll merge and then go back and make some small changes (things like kwargs `com` -> `center_mass`, to be consistent, etc). Couple q's though: 

* vhacd- what's the recommended way to install this?
  * It looks like in the main repo (https://github.com/kmammou/v-hacd/tree/master/bin) he has windows/OSX binaries named `testVHACD`, but your interface is looking for `vhacd`. I tried this on Linux and Windows, and it looks like testVHACD is probably the right binary?
* for the density/COM overrides
  * if the COM is overridden, does it make the inertia matrix trimesh returns wrong? I guess its just a translation so probably not?

mmatl:
Thanks so much! Happy to help with any cleanup as needed.

With regard to your questions:

* vhacd -- Yep, testVHACD is the correct binary :) We can modify the backend to look for testVHACD as well, we'd just installed it as vhacd. I compiled it from source, but I think all the recent pre-compiled binaries should work fine.
* overrides -- If the COM is overridden, it automatically removes mass_properties from the cache, so when moment_inertia is queried, it recomputes the mass_properties with the new COM, which should return a valid inertial matrix (assuming the mesh is watertight, so the volume computation works). Ditto for density. I assumed that if you overrode those properties, you wanted mass_properties to use the overridden values.

mikedh:
No worries, vhacd actually builds without much of a fight haha. It wasn't working in trimesh on Linux for some reason but I think I got it in https://github.com/mikedh/trimesh/commit/991edc30babb691ea8d193d8a53a8d5ea4891596. I also added the search for `testVHACD` as well as `vhacd`.  Let me know if I broke it for you :D.

I'm pretty sure I was able to build vhacd statically for linux and will PR a binary back into the main repo once I test it on a few machines (https://github.com/mikedh/v-hacd-1/tree/master/bin/linux). It produces surprisingly reasonable looking results too. 

Thanks again- changes look great!

mmatl:
Looks awesome! :) I'll test the changes soon, but they look great. And thanks for looking into getting a static linux binary for V-HACD -- I was really surprised they didn't have one yet haha!

PR #91, opened 2017-10-06
Add mesh-mesh collision checking

mmatl:
This collision checking depends on python-fcl, a python wrapper
for FCL 0.5.0. Not having python-fcl will not break things.
python-fcl is pip installable and supports Linux at the moment,
although I'll add support for OS X and Windows soon.

### DIFF
--- a/tests/test_collision.py
+++ b/tests/test_collision.py
@@ -1,0 +1,76 @@
+import generic as g
+
+
+class CollisionTest(g.unittest.TestCase):
+
+    def test_collision(self):
+        # Ensure that FCL is importable
+        try:
+            g.trimesh.collision.CollisionManager()
+        except ValueError:
+            g.log.warning('skipping collision tests, no FCL installed', engine)
+            return
+
+        cube = g.get_mesh('unit_cube.STL')
+
+        tf1 = g.np.eye(4)
+        tf1[:3,3] = g.np.array([5, 0, 0])
+
+        tf2 = g.np.eye(4)
+        tf2[:3,3] = g.np.array([-5, 0, 0])
+
+        # Test one-to-many collision checking
+        m = g.trimesh.collision.CollisionManager()
+        m.add_object('cube0', cube)
+        m.add_object('cube1', cube, tf1)
+
+        ret = m.in_collision_single(cube)
+        self.assertTrue(ret == True)
+
+        ret = m.in_collision_single(cube, tf1)
+        self.assertTrue(ret == True)
+
+        ret = m.in_collision_single(cube, tf2)
+        self.assertTrue(ret == False)
+
+        # Test internal collision checking and object addition/removal/modification
+        ret = m.in_collision_internal()
+        self.assertTrue(ret == False)
+
+        m.add_object('cube2', cube, tf1)
+        ret = m.in_collision_internal()
+        self.assertTrue(ret == True)
+
+        m.set_transform('cube2', tf2)
+        ret = m.in_collision_internal()
+        self.assertTrue(ret == False)
+
+        m.set_transform('cube2', tf1)
+        ret = m.in_collision_internal()
+        self.assertTrue(ret == True)
+
+        m.remove_object('cube2')
+        ret = m.in_collision_internal()
+        self.assertTrue(ret == False)
+
+        # Test manager-to-manager collision checking
+        m = g.trimesh.collision.CollisionManager()
+        m.add_object('cube0', cube)
+        m.add_object('cube1', cube, tf1)
+
+        n = g.trimesh.collision.CollisionManager()
+        n.add_object('cube0', cube, tf2)
+
+        ret = m.in_collision_other(n)
+        self.assertTrue(ret == False)
+
+        n.add_object('cube1', cube, tf1)
+
+        ret = m.in_collision_other(n)
+        self.assertTrue(ret == True)
+
+
+if __name__ == '__main__':
+    g.trimesh.util.attach_to_log()
+    g.unittest.main()
+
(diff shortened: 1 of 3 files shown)


mmatl:
Adding a quick fix in the test case now.

mmatl:
Looks like the Travis checks are the same ol' random failures from before, nothing too nefarious.

mikedh:
This looks awesome, thanks! I've had my eye on that JF branch for a while, cool that you guys are updating it! The python side looks like a great start and one could imagine adding collision managers to scene / primitive objects eventually too. 

So I'm having a bit of trouble building python-fcl: it looks like FCL changed the name of pretty much every file around https://github.com/flexible-collision-library/fcl/commit/4b68f59db166280f3e46bda0cec83123acdcdc5b so if I have the latest fcl python-fcl doesn't build for header-files-are-all-renamed reasons. 

I took a stab at updating the header files locations in https://github.com/mikedh/python-fcl/commit/258025147e23706ede746ccad72879cefab043e4 but it didn't magically work immediately, so I checked out progressively older versions of FCL, and was testing mostly with  https://github.com/flexible-collision-library/fcl/commit/b5b9a7017291ad56b11a756af376babe4fd1fbe2

Its pretty likely I'm doing something dumb, but when I build python-fcl I see some variant of a bunch of these:
```fcl/fcl.cpp: In function ‘int __pyx_pf_3fcl_3fcl_9Transform___cinit__(__pyx_obj_3fcl_3fcl_Transform*, PyObject*)’:
fcl/fcl.cpp:3039:146: error: no matching function for call to ‘Eigen::Transform<double, 3, 1>::Transform(fcl::Matrix3f, fcl::Vec3f)’
 _f_3fcl_3fcl_numpy_to_mat3f(__pyx_v_rot), __pyx_f_3fcl_3fcl_numpy_to_vec3f(__pyx_v_trans));
                                                                                          ^
In file included from /usr/include/eigen3/Eigen/Geometry:44:0,
                 from /usr/include/eigen3/Eigen/Dense:6,
                 from /usr/local/include/fcl/math/vec_3f.h:48,
                 from fcl/fcl.cpp:544:
```

What version of FCL, eigen, and libccd are you building with? I wonder if including the right versions as submodules and ignoring the system ones might make the build more reproducible. 

Thanks!
Mike

mmatl:
I'm using FCL 0.5.0, Eigen 2.92, and libccd 2.0-1. I'll upgrade my FCL and see if I can get python-fcl working with it -- I'll let you know how it goes ASAP.

mmatl:
Or rather, what do you think the best option is for making this build reproducible? If you're able to build cleanly against FCL 0.5.0, it might be fine as-is (aside from needing to make python-fcl Windows/OS X compatible), since it'll be tough to support python-fcl against an ever-changing development branch of FCL. If you're not able to, I can check and see what the problem might be -- could be something with Eigen and/or libccd, but I don't think that'll be the case.

mikedh:
Cool, I'll try building with those versions thanks! I'm travelling for the next week but as soon as I get it built once and can test I'll merge this. 

Yeah I don't know what the right answer build- wise is: maybe going full conda-forge? They already got [eigen](https://anaconda.org/conda-forge/eigen) but getting fcl/python-fcl/libccd on it would be a ton of work. Though as one data point pyembree seems pretty reproducibly installable. 

I think just some reproducible instructions for just Linux would be plenty, though brace yourselves for the people trying to compile it on their OSX-running toaster 😛

mikedh:
Built eventually with those versions thanks! Test passes for me and looks pretty reasonable so I'm merging. 

By the way what are your thoughts on adding hooks into your CollisionManager into the main objects? I was thinking one thing that could be nice is adding it to scene objects and then adding a `scene.in_collision` property or something. Also mesh objects could cache a CollisionManager and then have `mesh.collides_with(other_mesh)`.  I might look into that at some point unless you guys were planning to do it. 


Also, this is what I ended up using to build python-fcl, probably some of the stuff in there isn't necessary:
```
export LD_LIBRARY_PATH=`pwd`:$LD_LIBRARY_PATH:/usr/local/lib/x86_64-linux-gnu

git clone https://github.com/danfis/libccd.git
cd libccd
git checkout 64f02f741ac94fccd0fb660a5bffcbe6d01d9939
cd src
make -j4
sudo make install

git clone https://github.com/flexible-collision-library/fcl.git
cd fcl
git checkout 22f375f333beccc10c527974cef96784f0841649
cmake .
make -j4
sudo make install

git clone https://github.com/BerkeleyAutomation/python-fcl.git
cd python-fcl
export NP_INC=`python -c "import numpy; print(numpy.get_include())"`
python setup.py build_ext -v --include-dirs $NP_INC:../fcl/include --library-dirs ../fcl/lib

```

mmatl:
Thanks! Sorry for the delayed reply, but yeah, I'm definitely up for putting some hooks into the scene object for collision checking :) I can take care of that later this week.

For building python-fcl, that looks pretty good. Generally, a `sudo apt-get install libccd-dev` on Ubuntu seems to get things working fine on 14.04 and 16.04 for us. Python FCL is also pip-installable, so that should make things a bit cleaner as well :) For FCL, I do think it's probably still best to build from source.

mikedh:
Cool! Actually after looking around I think wheels may be the way to go here, check out [cibuildwheel](https://github.com/joerick/cibuildwheel) which looks pretty nice. It would make it so a `pip install python-fcl` doesn't require any building. 

I've been messing around with getting it to build on [my branch](https://github.com/mikedh/python-fcl) of python-fcl, if I succeed and you're interested I'm happy to PR.

mmatl:
Please do, that sounds great! If you don't have time, just let me know and I'm happy to do it -- having a wheel for this would be awesome, and I could probably fix the setup.py to work with Windows and OS X as well as Linux.

PR #97, opened 2017-10-24
Fix bug when updating vertices with a mask

mmatl:
Basically, the given code in update_vertices() is designed to update the
vertex list by filtering it out with a mask. In addition, it attempts
to filter the vertex normals by the same mask. However, the given
code will never accomplish this -- the newly-masked vertex normals
won't be the same shape as the old vertex list, so setting
vertex_normals will result in a no-op, leaving the vertex normals
unchanged. Assigning to the vertices first, then to the vertex normals
fixes this bug.


mmatl:
I discovered this small bug when trying to render some meshes with opengl. The process() function would merge a lot of the vertices together, but the length of the vertex_normals would stay the same. This simple patch should fix that problem.

mikedh:
LGTM, thanks!

PR #99, opened 2017-11-06
Allow collision checking to return object names

mmatl:
The prior interface to the collision checker only returned boolean
results for collisions. This update allows the checker to also return
the names of objects that were in collision.


mmatl:
Test checks out on my machine, and I think the interface update could be valuable.

Also, FYI, will be trying to build an OS X wheel for python-fcl soon!

mikedh:
LGTM thanks! Also re: OSX build, nice! It'll be interesting to see how easy/painful that is haha.

PR #101, opened 2017-11-07
fix missing module reference in docs

codingcatgirl:
This is needed for the module reference to be generated. otherwise, the .rst/.html files will be missing. hopefully, this fixes your readthedocs problem as well :)


mikedh:
Great find, thanks!

PR #110, opened 2017-12-03
decomposition and URDF changes

mikedh:
Hey @gizatt, I was about to add the same kwarg passthrough to VHACD/convex decomposition, and I noticed on the forks page you had already added it (as well as the switch to lxml which I'm a fan of as well). If it's OK with you I'd like to merge your changes upstream. 

Thanks!
Mike

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -1681,7 +1681,7 @@ class Trimesh(object):
         result = self.export(file_type='dict')
         return result
 
-    def convex_decomposition(self, engine=None):
+    def convex_decomposition(self, engine=None, **kwargs):
         '''
         Compute an approximate convex decomposition of a mesh.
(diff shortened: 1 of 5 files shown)


gizatt:
I'd be happy to have this stuff upstream-- thanks for taking the initiative! Looks like it might need some tweaking to pass CI -- is that something you're willing to take care of, or would you like me to try to resolve it? (My knowledge of the broader codebase is pretty limited, so it'll probably take me a bit to get it resolved.)

mikedh:
Cool thanks! No worries, I'll merge it into devel and fix the CI issue before merging into master. I think it the issue is related to the interfaces.generic to_dict change. Your change makes sense- I'll see why I was doing it the other way.

Issue #107, opened 2017-11-21
'Compressed' obj format loaded wrong


dvicini:
Hi,
It seems that Trimesh does not parse OBJ files correctly if they are in the 'compressed' format. Compressed in this case means that vertices along hard edges are not duplicated. An example of this i the OBJ file representing a cube:

```
o Cube
v 1.000000 -1.000000 -1.000000
v 1.000000 -1.000000 1.000000
v -1.000000 -1.000000 1.000000
v -1.000000 -1.000000 -1.000000
v 1.000000 1.000000 -0.999999
v 0.999999 1.000000 1.000001
v -1.000000 1.000000 1.000000
v -1.000000 1.000000 -1.000000
vn 0.0000 -1.0000 0.0000
vn 0.0000 1.0000 0.0000
vn 1.0000 -0.0000 0.0000
vn 0.0000 -0.0000 1.0000
vn -1.0000 -0.0000 -0.0000
vn 0.0000 0.0000 -1.0000
s off
f 2//1 4//1 1//1
f 8//2 6//2 5//2
f 5//3 2//3 1//3
f 6//4 3//4 2//4
f 3//5 8//5 4//5
f 1//6 8//6 5//6
f 2//1 3//1 4//1
f 8//2 7//2 6//2
f 5//3 6//3 2//3
f 6//4 7//4 3//4
f 3//5 7//5 8//5
f 1//6 4//6 8//6
```
If I load this file using Trimesh, the loaded mesh does not have the correct normals from the file. Instead, it recomputes vertex normals as average face normals, which is not what we want here. For reference, the loaded 'vertex_normals' then are:
```
[[-0.40824829 -0.81649658 -0.40824829]
 [-0.66666672 -0.33333344  0.66666656]
 [-0.57735012  0.57735041 -0.57735027]
 [-0.57735046  0.57735017  0.57735017]
 [ 0.40824849 -0.40824809 -0.81649658]
 [ 0.66666663 -0.66666663  0.33333348]
 [ 0.81649658  0.4082486  -0.40824798]
 [ 0.33333315  0.66666663  0.6666668 ]]
```
The problem seems to be that the OBJ format allows a vertex position to be used with different normals in different faces. To replicate this correctly inside Trimesh, Trimesh should probably just duplicate these vertices when loading such a file.

Thanks
Delio

mikedh:
Hey, I think this should be fixed by Pyry's PR which was just merged, if you load OBJ files without processing:
[...]

PR #117, opened 2017-12-15
improved obj loader: handle separately indexed pos/tex/normal

PyryM:
This reworks the wavefront obj loader in order to handle files where faces specify different indices for positions, texcoords, and normals.

The previous loader only worked correctly on files where each face looked like:
```
f 1/1/1 20/20/20 9/9/9
```
e.g., where each vertex of each face references a position, texcoord, and normal with the same numeric indices. 

However, most obj files will have faces with different indices, for example if you create a cube in blender, set shading to smooth, and export it, you'll see:
```
f 2//1 4//2 1//3
f 8//4 6//5 5//6
f 5//6 2//1 1//3
```

If the same position appears multiple times with different texcoords or normals, then multiple output vertices will be generated for each unique combination. For example, `2//1` and `2//3` will map to distinct vertices in the output. (However, unless you explicitly specify `process=False` in the loader, the trimesh constructor will just merge those vertices back together again).

Texture coordinates are treated like normals, so vertices are correctly split based on texcoord as well, in case trimesh more fully supports texcoords in the future.

The loader is written using a `for line: line.split()` approach instead of using numpy indexing tricks and/or regex findalls. Performance wise, compared to a numpy/regex approach it is about 10% slower (in python3) on files with only positions, but it is almost twice as fast on files where vertices need to be split/remapped because it can parse and remap a file in a single pass.


trimesh/io/misc.py
```diff
@@ -61,129 +62,83 @@ def load_wavefront(file_obj, file_type=None):
...
+                    gf = f.split('/')
+                    out_v.append(attribs['v'][int(gf[0])-1])
+                    if len(gf) > 1 and gf[1] != '':
+                        out_vt.append(attribs['vt'][int(gf[1])-1])
+                    if len(gf) > 2:
+                        out_vn.append(attribs['vn'][int(gf[2])-1])
+                out_f.append(remap_table[f])
+        elif gps[0] == 'o':
+            _append_mesh(out_v, out_vt, out_vn, out_f, g_list)
+            out_v = []
```
mikedh: After appending a mesh and starting a new one, we're resetting `out_v`, `out_vn`, `out_f`, but not `out_vt`- is that intentional?

mikedh:
Awesome thanks this is great! This addresses https://github.com/mikedh/trimesh/issues/107 (if called like `mesh.load(file_name, process=False)`)and hopefully is more reliable in general. I’m going through and had minor questions (the not clearing ‘vt’ on a new object mostly). Also what are your thoughts on the group handling? In an ideal world would a group be handled the same as an object?


As a quick performance check on a 37 mb OBJ with face colors:

Current OBJ loader:
5.82 s ± 509 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Your loader:
5.52 s ± 135 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

My slight modifications are basically identical:
5.48 s ± 34.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

trimesh/io/misc.py
PyryM: Oh no, that's a mistake alright.

PyryM:
I think it depends on what people are using groups for.

trimesh/io/misc.py
mikedh: cool fixed in https://github.com/mikedh/trimesh/commit/df89a38d58574a930f7819a486fbe4f133ff1d9e

PyryM:
If loaders could take arguments beyond the file name (could the process argument perhaps be extended beyond True/False?), it would be easy enough to have the option to not do any vertex remapping (and then throw an error if you try to do so with a file that would require remapping).

mikedh:
I think additional kwargs for loaders are fine as long as the loader has the required ones, but it does make a solution a little less generally useful (as someone then has to call the loader directly and know details about their file and the loading process). 

I wonder if there is a way to make the remap table maintain order by default, except on files that have the split indices (where there should be no expectation of ordering).

PyryM:
Yeah that's the lexicographical sorting option (which you may not have seen because I snuck it in in an edit rather than a new comment).

mikedh:
Cool yeah the lexsort option seems like a great option. My understanding:

- for people who just want to load an OBJ, it doesn't require a bool flag or any understanding of the format/loading process
- for people who have a file WITHOUT split indices and want to reference things by index it works as expected (Matt)
- for people who have a file WITH split indices it still does the right thing
- people who have a file with split indices and want to reference things by index are SOL, but this is not a thing that makes much sense. 

Anyone want to take a stab at a PR? I can probably look into it in a few days if not.

PyryM:
Oh, hmm, I realized there is another case: what happens if you have a .obj file with vertices that are never references by any faces? As an extreme case, imagine a file with vertex positions but no actual faces defined.

A way around this is to first push all the original vertex attributes, and then afterwords append remapped vertices as needed. The downside is that the resulting loaded mesh can potentially have a lot of extra vertices that are never used.

For example, consider:
```
v 1 2 3
v 4 5 6
v 7 8 9
vn 1 0 0
vn 0 1 0
vn 0 0 1
f 3//2 1//3 2//1 
```
This would result in a vertex buffer that has extra vertices never used in the index buffer:
```
vertices
[0] v = [1, 2, 3], n = [1, 0, 0]
[1] v = [4, 5, 6], n = [0, 1, 0]
[2] v = [7, 8, 9], n = [0, 0, 1]
[3] v = [7, 8, 9], n = [0, 1, 0]
[4] v = [1, 2, 3], n = [0, 0, 1]
[5] v = [4, 5, 6], n = [1, 0, 0]

indices
[[3, 4, 5]]
```

And while you can imagine another level of bookkeeping that would solve this, fundamentally any mapping is going to have cases that break just because the internal representation that trimesh uses is more limited than that of obj files.

mikedh:
Yeah, I think sorting by remap key is a pretty easy/cheap way to solve the basic complaint if not all cases. I just pushed in https://github.com/mikedh/trimesh/commit/c019a25874fec33ca6ec3094621743711a29623f

PR #119, opened 2017-12-16
obj: export normals and texcoords when present

PyryM:
Of course most trimesh operations won't keep texture coordinates or manually set vertex normals in sync, but baby steps.

Perhaps the trimesh constructor should default to process=False when either vertex normals or texture coordinates are provided?


mikedh:
Great thanks! I switched the vertex_normal check to check that normals exist in the cache rather than checking the `mesh.vertex_normals` property shape (which is autogenerated to always be correct). I also changed to format call from a list comprehension to numpy wangling followed by a single format call which actually does speed things up here. On that 500K face mesh:

Original exporter:
1.51 s ± 50.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

PR with normals:
6.21 s ± 356 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Post mods, with single format call no vertex normals:
1.6 s ± 56.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Post mods, with a single format call and vertex normals:
3.02 s ± 60.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Thanks for the PR!

PyryM:
Creating a giant format string the length of all the faces seemed to me like it couldn't possibly be the fastest way of doing this, and indeed unrolling some number of loops (but not all of them), is faster:

```
dt (np):  2.4073729515075684
dt (fmt_unroll1):  2.811537027359009
dt (fmt_unroll5):  2.295804977416992
dt (fmt_unroll10):  2.263214111328125
dt (fmt_unroll20):  2.185274124145508
dt (fmt_unroll60):  2.1641438007354736
dt (fmt_unroll100):  2.197834014892578
dt (fmt_unroll200):  2.2189409732818604
```

where

```python
def unrollx(ff, x):
    return "\n".join("f " + " ".join(ff.format('{' + str(i+(j*3)) + '}') 
                                     for i in range(3)) 
                     for j in range(x))
    
def format_unroll(faces, face_type = 'v', unroll = 5):
    face_formats = {'v': '{0}', 'vt': '{0}/{0}', 'vn': '{0}//{0}', 'vtn': '{0}/{0}/{0}'}
    rowf = unrollx(face_formats[face_type], unroll)
    out_str = "\n".join(rowf.format(*row) for row in faces.reshape((-1,unroll*3)) + 1)
    return out_str
```

mikedh:
I wasn't quite following your timing test so I made the simplest example I could think of for the two ways of formatting rows. When I run it, the large format string with a single call is faster under all the face counts I tried:

```
import timeit
import numpy as np
    
def format_loop(array):
    '''
    Loop through each row and run a format operation per- row
    '''
    result = '\n'.join('{} {} {}'.format(*row) for row in array)
    return result
    
def format_once(array):
    '''
    Create a giant string and run a single format operation
    '''
    result = '{} {} {}\n' * len(array)
    result = result.format(*array.reshape(-1))[:-1]
    return result
    
if __name__ == '__main__':
    # check to make sure functions are doing the same thing
    data = np.arange(1000*3).reshape((-1,3))
    a = format_loop(data)
    b = format_once(data)
    assert a == b
    
    # the order of magnitude is that a large/midsize mesh has roughly 1 million faces
    counts = np.linspace(0.0, 1e6/3, 10)[1:].astype(int) * 3

    # how many runs per function
    test_runs = 10
    result = [['counts', 'loop', 'once']]
    
    for count in counts:
        t_loop = min(timeit.repeat(stmt='f(d)', 
                          setup='d=np.arange(count).reshape((-1,3))', 
                          globals={'count' : count,
                                 'f' : format_loop,
                                 'np' :np},
                          number=test_runs,
                          repeat=3)) / test_runs
                          
        t_once = min(timeit.repeat(stmt='f(d)', 
                                  setup='d=np.arange(count).reshape((-1,3))', 
                                  globals={'count' : count,
                                         'f' : format_once,
                                         'np' :np},
                                  number=test_runs,
                                  repeat=3)) / test_runs
        result.append([count, t_loop, t_once])
        print(result[-1])
```

Results:
```
count         looping method       run once method
[111111, 0.07480875927552019, 0.04058497228393274]
[222222, 0.15056783151575753, 0.08190685538707268]
[333333, 0.22456352666495097, 0.12398028881371773]
[444444, 0.3037991605011115, 0.16574710130843684]
[555555, 0.37661803469813504, 0.21271855037118853]
[666666, 0.4511606896126722, 0.250817539631926]
[777777, 0.5297424813597289, 0.2929469935616851]
[888888, 0.6049035368205835, 0.33063671805179096]
[999999, 0.6794521590759217, 0.3798115658776311]
```

PyryM:
Right, one giant string (effectively unrolling the format loop completely) is faster than per-row, but if you unroll instead by ten rows at a time it's faster than either of those two extremes:
```
f {0}/{0}/{0} {1}/{1}/{1} {2}/{2}/{2}
f {3}/{3}/{3} {4}/{4}/{4} {5}/{5}/{5}
f {6}/{6}/{6} {7}/{7}/{7} {8}/{8}/{8}
f {9}/{9}/{9} {10}/{10}/{10} {11}/{11}/{11}
f {12}/{12}/{12} {13}/{13}/{13} {14}/{14}/{14}
f {15}/{15}/{15} {16}/{16}/{16} {17}/{17}/{17}
f {18}/{18}/{18} {19}/{19}/{19} {20}/{20}/{20}
f {21}/{21}/{21} {22}/{22}/{22} {23}/{23}/{23}
f {24}/{24}/{24} {25}/{25}/{25} {26}/{26}/{26}
f {27}/{27}/{27} {28}/{28}/{28} {29}/{29}/{29}
```

PyryM:
For example, in your example timing framework above, you can define
```python
def format_unroll(faces, unroll = 300):
    rowf = "\n".join(["{} {} {}"] * unroll)
    out_str = "\n".join(rowf.format(*row) for row in faces.reshape((-1,unroll*3)))
    return out_str
```
and then I get:
```
count    per_row             giant format string   unrolled 300
[180000, 0.13243745490035508, 0.07367865519918268, 0.0706970024999464]
[360000, 0.26479814189951867, 0.15050376970029902, 0.14491130009992048]
[540000, 0.4024645936995512, 0.22882544200110716, 0.21426412779983367]
[720000, 0.5492700888993568, 0.31801500169967767, 0.292010567500256]
[900000, 0.6768803890998243, 0.380749787199602, 0.358476088300813]
[1080000, 0.8354325858002994, 0.4570282450993545, 0.4339503370996681]
```

which is only a slight advantage in this case (where there are only vertex positions), but with normals and textures the advantage is larger.

mikedh:
Interesting how close it is, I guess its the slowness of python function calls vs. allocation of larger memory block. 

It occurred to me that this logic should really be in `util.array_to_string` so I added a `value_format` parameter and reformatted the OBJ exporter to use the logic there, so any optimizations to that function would benefit the rest of the code base. The giant string is a little inelegant but it is fast and simple, for a 5% speedup unrolling is a tough sell for me. 

That being said I didn't test on things like `value_format='{}/{}/{}'` where the speedup might be a lot more than 5%.

PyryM:
Yeah, the minor performance improvement is not worth the more complicated code. It's surprising, and not to python's credit, that the fast way to do fairly simple text formatting is to use 4x+ as much memory (3x to tile the face indices, and then another 1x at least to produce the giant format string).

PR #120, opened 2017-12-19
Small bugfixes and upgrades

mmatl:
This PR has a few bugfixes and improvements:

- Pre-set center of mass is preserved across copies and moved appropriately during transformations.
- Allows for returning sampled triangle indices in surface sampler.
- Allows default ray tracer to return only first hit.
- Fixes small potential error in vhacd wrapper.

Let me know if you have any questions!

Best,
Matt

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -1671,7 +1671,6 @@ class Trimesh(object):
             mass['mass'] = -mass['mass']
         if mass['volume'] < 0:
             mass['volume'] = -mass['volume']            
-        
         return mass
 
     def scene(self):
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -524,8 +524,6 @@ class Trimesh(object):
         -----------
         center_mass: (3,) float array, volumetric center of mass of the mesh
         '''
-        if not self.is_watertight and len(self._data['center_mass']) != 3:
-            log.warning('Center of mass requested for non- watertight mesh!')
         center_mass = self.mass_properties['center_mass']
         return center_mass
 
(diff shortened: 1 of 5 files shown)


mmatl:
P.S. VirtualGL works well for using a viewer remotely over SSH. If you want, I can link you to some of our instructions for getting it working.

mikedh:
LGTM thanks! Couple thoughts:

* I think I will change the defaults on `remove_close` and `sample_surface` to always return the face_ids, and move the `return_index` flag up to `Trimesh.sample(count, return_index=False)` to avoid some boolean flags

* for the negative volume check, have you seen negative volume on watertight meshes with correct winding (normals facing out)? It seems like usually the negative volume is indicating that something is screwey with the mesh, is simply negating the mass/inertia values the right thing to do? I could imagine putting something like this into `mesh.process`:
```
if self.is_watertight and self.is_winding_consistent and mass['volume'] < 0.0:
    with self._cache:
        self.face_normals *= -1
        self.faces = np.fliplr(self.faces)
```
Having a positive volume is looked at in `mesh.is_volume` and some of the tests as an indicator that the mesh is generally OK.

mikedh:
BTW thanks for the VirtualGL lead! I actually got it mostly working in a docker container by installing `apt-get install xvfb libgl1-mesa-swx11`, running an `Xvfb :0 -screen 0 1600x1200x24 &`, then just using my normal pyglet openGL stuff. It would probably be better with llvmpipe (which you can get I think by instead installing `xvfb libgl1-mesa-dri libgl1-mesa-glx`) but that was giving me weird draw ordering artifacts for some reason.

PR #130, opened 2018-01-30
Updates to collision manager and stable pose computation

mmatl:
This PR has two commits:

- First, a simple update to stable pose computation to prevent infinite loops if a cycle is formed in the toppling digraph due to FP error.
- Second, an addition to the collision manager code to allow for returning more detailed collision and distance response data.


mikedh:
Looks great, thanks!

PR #133, opened 2018-02-05
Change the way mesh normals are validated

mmatl:


trimesh/triangles.py
```diff
@@ -253,32 +253,25 @@ def bounds_tree(triangles):
...
     """
     triangles = np.asanyarray(triangles, dtype=np.float64)
     if not util.is_shape(triangles, (-1, 3, 3)):
         raise ValueError('Triangles must be (n,3,3)!')
 
-    # if both edges of the triangles OBB are longer than tol.merge
-    # we declare them to be nondegenerate
-    ok = (extents(triangles=triangles,
-                  areas=areas) > tol.merge).all(axis=1)
+    # Option 1: Minimum angle threshold
```
mmatl: This change is optional, if you don't like it we can get rid of it :) Basically, just uses an angle-based validation for collinear points and assumes that any colocated vertices will have already been merged.

trimesh/util.py
```diff
@@ -65,7 +65,7 @@ def unitize(points, check_valid=False):
         length[np.isnan(length)] = 0.0
 
     if check_valid:
-        valid = np.greater(length, _TOL_ZERO)
+        valid = np.not_equal(length, 0.0)
```
mmatl: This change is interesting and I think for the best -- we might want to just trust face normals unless they're actually zero.

mikedh: Since these are floats, shouldn't it really be `np.logical_not(np.isclose(length, 0.0))`

mikedh: But yeah 👍 on using numpy compare functions

mmatl: Oh right, excellent point. I'll update that now.

mmatl: Actually, looks like behavior should be the same assuming you want rtol = atol = 0.0 in the np.isclose() call. That is, assuming you want to only mark normals that are actually zero-length as invalid and are okay with very small faces.

trimesh/triangles.py
```diff
@@ -253,32 +253,26 @@ def bounds_tree(triangles):
...
-    # if both edges of the triangles OBB are longer than tol.merge
-    # we declare them to be nondegenerate
-    ok = (extents(triangles=triangles,
-                  areas=areas) > tol.merge).all(axis=1)
+    # Minimum angle threshold
+    a, valid = util.unitize(triangles[:, 1] - triangles[:, 0], check_valid=True)
+    b, valid = util.unitize(triangles[:, 2] - triangles[:, 1], check_valid=True)
+    dots = np.abs(np.einsum('ij,ij->i', a[valid], b[valid]))
+    ok = valid
+    ok[valid] = (1.0 - dots > tol.zero)
```
mmatl: New change here for you to review -- basically does a double filter for truly zero-area faces and faces whose points are collinear by some threshold.

mmatl: Could also re-do this to just explictly compute normals and then mark any faces with zero-area for deletion.

trimesh/util.py
mikedh: I think we want a tolerance on the float comparison, comparing to exactly 0.0 seems likely to be squirrely. The real question is how small to go- currently tol.merge connects vertices roughly closer than 1e-8. So if we have a face with two 1e-8 edges the magnitude of the cross product would be 1e-16. What about a comparison ~= that?

trimesh/triangles.py
mikedh: Gotcha, so this formulation of degenerate faces is saying, "both edges are longer than the tolerance inside unitize, and the dot product of one onto the other is larger than tol.zero." Seems reasonable- also I changed the threshold to be settable in unitize in a branch. 

Though doesn't this work out to the same thing as the original one which was "one edge and the height of the triangle are greater than tol.merge"? The dot product is the height, and the edge length is the width. BTW in the formulation above it should probably be
```
a, a_valid = util.unitize(triangles[:, 1] - triangles[:, 0], check_valid=True)
b, b_valid = util.unitize(triangles[:, 2] - triangles[:, 1], check_valid=True)
valid = np.logical_and(a_valid, b_valid)
```

Your point about preferring merged vertices over degenerate faces makes sense- so we should only be throwing away faces significantly smaller than tol.merge (tol_zero seems good).

mikedh: Oh nice einsum btw, I benchmarked to see if I should switch `trimesh.util.diagonal_dot` to use it, looks like the sum method is slightly faster albeit with more variance for some reason: 

```
In [11]: %timeit trimesh.util.diagonal_dot(a,b)
The slowest run took 14.29 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 22.3 µs per loop

In [12]: %timeit np.einsum('ij,ij->i', a,b)
The slowest run took 5.41 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 51.8 µs per loop


```

mikedh:
Sorry for the competing branches but I merged some of your changes into the magic branch because we got slightly out of sync:
https://github.com/mikedh/trimesh/tree/magic

The only major difference is that face_normals saves its valid mask to the cache, and then by default `remove_degenerate_faces` just uses that mask, IE "do the two edges of the triangle have a cross product with a length greater than tol.merge ** 2 (works out to 1e-16). Then the `triangles.nondegenerate` is only called if you specify a minimum height.

trimesh/util.py
mmatl: That sounds like a great idea! The cross product magnitude also has a sin(theta) term though, so we'll need to incorporate allowable angles somehow -- if you allow angles as small as 1e-12 rad, suddenly the floor becomes 1e-28. What do you think a reasonable threshold on the angle is?

trimesh/triangles.py
mmatl: Good catch on the formulation bug, I'll fix it and update :)

I don’t think this is the same thing as the original one. Here’s my thinking:

There are two types of degenerate faces:
- Faces with co-located vertices (an edge less than `tol.merge` in length)
- Faces with unique vertices, but the vertices are collinear

The first case is taken care of by `merge_vertices`, so any triangle passed to this function is guaranteed to have three edges with length at least `tol.merge`.

The second case should be addressed by this function. In the original implementation, the check was dependent on the area of the triangle, but when checking for collinearity, we probably don’t want to give different results if the mesh is simply scaled up. In the new implementation, I explicitly unitize the vectors to factor out scale, and then threshold the angle between them (the dot product). I think that this is probably cleaner, as it creates a clearer criterion for rejecting faces. What do you think?

An example of where the new function would differ from the old is with an equilateral triangle with side length `1.01e-8`. It seems that this should be allowed, but the old method would reject it, as the height of this triangle is less than `1e-8`. However, the new method would note that the angles are fine (60 degrees) and not flag the triangle as bad.

Actually, I probably don’t even need the validity check for `unitize` here — all of the edges must be greater than `1e-8` in length, since they were already fixed up by merging vertices. I’m thinking it can’t hurt to keep it, though :)

Let me know what you think! Sorry this is turning into a long PR comment chain haha, these sort of thresholding things tend to be messy. Definitely don’t change anything just for my use case!!! Only accept them if you think they make good sense in the general case.

Best,
Matt

mmatl: Actually, my method does have a failure case -- very long, skinny triangles :( I think your method might actually be better with some re-jiggering... going to think about this tonight while I sleep haha.

mmatl:
Awesome, your branch looks good to go :) I like your fix there a lot! I'll close this PR and delete my branch. Thanks for dealing with the thresholding struggles with me haha, really appreciate it!

mikedh:
No worries, thanks for thinking about this! The magical face delete always bugged me. I'll version bump and merge `magic`.

trimesh/util.py
mikedh: Ah good point on the sin term I forgot about. So `tol.zero` in trimesh is 1e12, and looks like [machine epsilon](https://en.wikipedia.org/wiki/Machine_epsilon) for float64 is ~1e-16. So I should probably make that normal magnitude check some number larger than machine epsilon- 1e-15?

mikedh: Numpy has a nice function for machine epsilon actually, might just make everything use tol.zero:
`tol.zero = np.finfo(np.float64).resolution * 100`

mmatl:
Awesome, thanks a million!

Issue #138, opened 2018-02-21
int8 and uint8 "in the wild" in ply files


vlimit:
Loading a ply file with the following header causes an exception in read_ply_header:

"""
ply
format ascii 1.0
element vertex  6449
property float32 x
property float32 y
property float32 z
element face 12894
property list uint8 int32 vertex_index
end_header
"""

C:\Users\vlimit>python test.py
Traceback (most recent call last):
  File "convert.py", line 3, in <module>
    mesh = trimesh.load('outdata.ply')
  File "c:\python27\lib\site-packages\trimesh\io\load.py", line 86, in load
    **kwargs)
  File "c:\python27\lib\site-packages\trimesh\constants.py", line 126, in timed
    result = method(*args, **kwargs)
  File "c:\python27\lib\site-packages\trimesh\io\load.py", line 129, in load_mesh
    file_type=file_type)
  File "c:\python27\lib\site-packages\trimesh\io\ply.py", line 48, in load_ply
    elements, is_ascii = read_ply_header(file_obj)
  File "c:\python27\lib\site-packages\trimesh\io\ply.py", line 165, in read_ply_header
    endian +
KeyError: u'uint8'

The problem is caused by the uint8 in the property list:

property list uint8 int32 vertex_index

Simple fix:

"""
--- a/trimesh/io/ply.py
+++ b/trimesh/io/ply.py
@@ -19,9 +19,11 @@ ply_dtypes = {'char': 'i1',
               'short': 'i2',
               'ushort': 'u2',
               'int': 'i4',
+              'int8': 'i1',
               'int16': 'i2',
               'int32': 'i4',
               'uint': 'u4',
+              'uint8': 'u1',
               'uint16': 'u2',
               'uint32': 'u4',
               'float': 'f4',
"""

mikedh:
Resolved by your PR- thanks!

PR #139, opened 2018-02-21
Issue #138: Added unrecognised int8 & uint8 types to ply_dtypes.

vlimit:
Simple fix to parse ply files that use uint8/int8, like "property list uint8 int32 vertex_index".


mikedh:
This looks great, thanks!

PR #141, opened 2018-02-22
Fix undefined variable name in ContactData

cyrilpic:
In the current code, requesting contact information for collisions raises an error as `result` is not defined in the initialiser.


mikedh:
Looks great, thanks!

PR #142, opened 2018-03-15
Fix matrix_to_marching_cubes offset

willmycroft:
The previous code produces an incorrect offset for the mesh produced by marching cubes, as illustrated by the below code.

import numpy as np
from trimesh.voxel import matrix_to_marching_cubes

voxels = np.ones((3,3,3), dtype=np.bool)
mesh = matrix_to_marching_cubes(voxels, 1.0, np.zeros(3))
print(mesh.bounds)
mesh = matrix_to_marching_cubes(voxels, 3.0, np.zeros(3))
print(mesh.bounds)


mikedh:
Good catch- thanks!

PR #148, opened 2018-04-12
Added a path for Blender search on MacOS.

Anthony-MostlyHarmless:
In my case on a fresh Blender install the blender.app folder was located in /Applications/Blender and not /Applications directly.


mikedh:
LGTM, thanks!

PR #150, opened 2018-04-12
Fixed examples/offscreen_render.py

Gjacquenot:
Due to a previous refactoring, this example failed while calling scene.save_image that does take a filename as input.


mikedh:
Looks great thanks!

PR #156, opened 2018-04-18
Bug fix in nearby_faces

willmycroft:
The AABB for each point must contain the sphere of radius distance to the nearest vertex. Therefore, our AABB must be a cube of twice this distance. The cuboid used before is insufficient for this purpose and can result in not finding the nearest point.

### DIFF
--- a/trimesh/proximity.py
+++ b/trimesh/proximity.py
@@ -16,7 +16,7 @@ from collections import deque
 
 def nearby_faces(mesh, points):
     """
-    For each point find nearby faces relativly quickly.
+    For each point find nearby faces relatively quickly.
 
     The closest point on the mesh to the queried point is guaranteed to be
     on one of the faces listed.
@@ -44,7 +44,7 @@ def nearby_faces(mesh, points):
     kdtree = mesh.kdtree
 
     # find the distance to each vertex to create an axis aligned bounding box
-    distance_vertex = np.abs(points - mesh.vertices[kdtree.query(points)[1]])
+    distance_vertex = kdtree.query(points)[0]
     distance_vertex += tol.merge
 
     # axis aligned bounds


mikedh:
Isn't it doubling that distance when it [constructs the query bounds](https://github.com/mikedh/trimesh/blob/master/trimesh/proximity.py#L51-L52 )?

Do you have a test where this logic fails? I'm a little hesitant to switch from cuboid to sphere as in a lot of meshes that will explode the number of faces it has to consider. Is it possible the cuboid just needs to be padded by a larger distance (currently it is tol.merge, 1e-8)?

Thanks!
Mike

willmycroft:
Hi Mike,

The following image should make it clear where the problem can arise.

image

My test arose from a rather complicated mesh, let me know if you want and I'll dig out the example in code.

Thanks

Will

mikedh:
Ahh, good point, thanks for the illustration. I think you were missing a `reshape` which is why the tests were failing, I added your change in https://github.com/mikedh/trimesh/commit/b9d3c30c1a630fb87f19fc756efeb5023d5dea4e

The performance is already pretty bad on `nearby_faces`, I'll play around with options that cover this edge case and hopefully are a little faster (maybe `sklearn.neighbors.BallTree` queries)

willmycroft:
Thanks, sorry for missing the reshape.

I'm writing some code for computing thickness (following
https://www.tandfonline.com/doi/pdf/10.1080/16864360.2015.1084186) and
computing curvature measures (following
http://www.cs.jhu.edu/~misha/Fall09/Steiner03.pdf). Is this something you'd
be interested in having merged into the main code base?

I'm fairly new to open source software collaboration. Is there anything I
should be aware of when writing the code beyond trying to follow the
conventions in the rest of the project?

Thanks
Will

On 18 April 2018 at 21:49, Michael Dawson-Haggerty <notifications@github.com

mikedh:
No worries and thanks for making the effort to PR! Both of those features (thickness metrics and curvature) would be awesome to get into the main codebase! I've computed thickness before in a pretty naive way but it would be great to have a nice implementation of it. 
- I think thickness probably would be a natural fit in the `trimesh.proximity` module
- curvature might deserve its own `trimesh.curvature` module

Some things I generally aim for with new code:
- docstrings (numpydoc format)
- if adding a dependency ideally only adding things that install cleanly on [Linux/OSX/Windows](https://github.com/geometries/trimesh_upstream) so someday conda won't be necessary
 - decent performance- it's pretty easy to write functions which are super slow, usually in Python you just need to be careful to never loop through the entire face/vertex array (vs vector operations)

Best,
Mike

PR #157, opened 2018-04-18
Fix bug in intersects_id

willmycroft:
When returning locations, we must check that the list is non-empty because other grouping.unique_rows throws an error.

Moreover, the user is expecting numpy arrays as a return type instead of python lists, so return these instead of [],[],[].


mikedh:
Looks good! I think `unique_rows` should be OK on empty data after https://github.com/mikedh/trimesh/commit/acc84fe802b74c1003a1c9bbdada9575a151580e but your change makes sense, as it gets the typing right and skips a function call.

Thanks for the PR!

PR #164, opened 2018-05-05
closest_point() fix added to select the better suitable face from two ambiguous faces.

dbukenberger:
A closest point on the surface is maybe on an edge between two adjacent faces and therefore they are ambiguous. However, it may happen that one of these faces is not even facing the querying point which results in a wrong signed distance to this point.
If have added a fix that first checks if there are two candidate faces with the same distance, then evaluates both their normals and compares them to the direction of closest-point and querying-point. Eventually select the index of the face that is actually facing the querying point.

### DIFF
--- a/tests/test_proximity.py
+++ b/tests/test_proximity.py
@@ -135,6 +135,26 @@ class NearestTest(g.unittest.TestCase):
         mesh = g.get_mesh('20mm-xyz-cube.stl')
         assert (mesh.nearest.signed_distance([[-51, 4.7, -20.6]]) < 0.0).all()
 
+    def test_acute_edge_case(self):
+        # acute tetrahedron with a sharp edge
+        vertices = [[-1, 0.5, 0], [1, 0.5, 0], [0, -1, -0.5], [0, -1, 0.5]]
+        faces = [[0,1,2],[0,2,3],[0,3,1],[3,2,1]]
+        mesh = g.trimesh.Trimesh(vertices, faces)
+
+        # a set of points on a line outside of the tetrahedron
+        # their closest surface point is [0, 0.5, 0] on the sharp edge
+        # for a point exactly in the middle a closest face is still ambiguous
+        # -> take an even number of points
+        n = 20
+        n += n%2
+        pts = g.np.transpose([g.np.zeros(n), g.np.ones(n), g.np.linspace(-1, 1, n)])
+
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! Looks good to me- though would you mind adding a test to [test_proximity](https://github.com/mikedh/trimesh/blob/master/tests/test_proximity.py) which demonstrates the edge case?

dbukenberger:
Thank you for this awesome project.
Yes, of course - I will look into that and construct a test.
Should I also add a model if necessary or try to hardcode a minimal example?

mikedh:
Hahah no worries!

Yeah feel free to do either, a small model to `models`, or if you can demonstrate it on a unit cube or something that would be good too. 

Thanks!

dbukenberger:
It had to be sharp edge between faces enclosing an acute angle - so a cube wouldn’t do the trick.
Therefore I constructed a small tetrahedron with a sharp edge and some points to demonstrate it in a test scenario.

However, on the workstation where I encountered the issue, the trimesh and scipy versions are a bit outdated. On another machine where I have recent versions installed I could not reproduce the "false sign" behavior. It had something to do with the order of faces in the trimesh.faces array.

Nevertheless, I edited my fix so that the returned face-index of the closest face is at least the one that is actually facing the querying point.

mikedh:
Thanks for the test! It looks like the test that's failing `test_sample` is indicating that points sampled on the surface of a mesh are not always returning distances less than `tol.merge` after this change. The sampler is reasonably well tested and hasn't changed.

I think it's related to your conditional check to make sure both points are on the surface, which is now checking that the two smallest distances are close to each other:
```
(abs(distance[idx0] - distance[idx1]) < tol.merge
```
I think what you want is instead a check that both have an absolute distance below `tol.merge`:
```
# check to make sure the two smallest distances are both effectively zero
if (np.abs(distance[[idx0, idx1]]) > tol.merge).any():
    idx = idx0
```

It might be worthwhile to unpack the conditional check on line `162` into a couple different statements to make it a little easier to understand.

Thanks for working through this!

mikedh:
Looks good to me, thanks! Loading those formats is probably failing due to missing `lxml` (`pip install lxml`).

dbukenberger:
Okay cool. Yes, that was the package that was missing, thank you. =)

PR #166, opened 2018-05-05
enhancement: solid voxelization

LinJiarui:
add methods for solid voxelization, the inner space of a trimesh can be filled by voxels, call show(solid_mode=True) for visualization


mikedh:
Thanks for the PR! I'll add some comments in line. 

How was performance? I worry a little bit about python iterating through every voxel being pretty slow- is there any way to vectorize that operation?

trimesh/voxel.py
```diff
@@ -266,6 +291,62 @@ def voxelize_subdivide(mesh, pitch, max_iter=10):
 
     return voxels_sparse, origin_position
 
+def voxelize_subdivide_solid(mesh,pitch,max_iter=10):
+    voxels_sparse,origin_position=voxelize_subdivide(mesh,pitch,max_iter)
```
mikedh: If we passed the sparse voxels, objects could re- use that calculation, something like:

```
def voxelize_subdivide_solid(sparse, sparse_origin):
```

trimesh/voxel.py
```diff
@@ -266,6 +291,62 @@ def voxelize_subdivide(mesh, pitch, max_iter=10):
...
+    #create grid and mark inner voxels
+    max=voxels_sparse.max()+1
```
mikedh: `max` is a builtin function- I might rename this to `max_value` or something

LinJiarui: Thanks for the improvement. I'm not sure if we can vectorize the operation or not, I'll check it later. 
Just for reference, I tested the method with bunny.obj and buddha.obj, for a 64x64x64 grid, the former takes 0.36s (20% of total time) and the later takes 0.56s (8% of total time) to fill the inner voxels, which I think is reasonable now.

LinJiarui:
Thanks for the improvement. I'm not sure if we can vectorize the operation or not, I'll check it later.
Just for reference, I tested the method with bunny.obj and buddha.obj, for a 64x64x64 grid, the former takes 0.36s (20% of total time) and the later takes 0.56s (8% of total time) to fill the inner voxels, which I think is reasonable now.

mikedh:
Awesome! I'll test with this tomorrow or possibly later this week. Thanks!

mikedh:
Merged into master in https://github.com/mikedh/trimesh/commit/258c05d068aa01a67f107595ecc942c4f2647074 

Thanks!

PR #169, opened 2018-05-17
Feature/added ctm support

lejafar:
I haven't really been able to fully test it, but I thought I would first get some feedback before going further


mikedh:
Looks great! That was fast hahah. Since its a ctypes thing that doesn't require building I think this could work nicely, as long as we keep it as a soft dependency by wrapping any import exceptions. 

The one thing that people are guaranteed to raise issues with is how to get the right shared library. 

Options:
- Set up an upstream python package that bundles the shared library (like we did with [python-fcl](https://github.com/BerkeleyAutomation/python-fcl)) by building it with cibuildwheel, that we can add to the `trimesh` dependency list. 
- (easier) Just add a shell script we can point people at, like the [draco one](https://github.com/mikedh/trimesh/blob/master/docker/builds/draco.bash).

lejafar:
I prefer the second one! And I'll also write the export method!

mikedh:
Cool! Just out of curiosity (still happy to merge), why OpenCTM over [Draco](https://github.com/google/draco)? OpenCTM looks basically abandoned on their Sourceforge and Github forks, while draco is quite active and backed by a large corporation.

lejafar:
Its just that I'm receiving the files mostly in a OpenCTM format (I'm working in additive), we do use draco but just for creating previews though!

mikedh:
Sounds good! Yeah if you add that shell script I'll merge. I may move the files around a little bit, I was thinking of making a directory `openctm` in io, and then do the try block in `__init__`:

```
try:
     from .ctm import load_ctm
     ctm_loaders = {'ctm': load_ctm}
except ImportError:
    ctm_loaders = {}
```

mikedh:
Oh- would you mind also adding a small (500kb- ish) test model in the openCTM format to /models?

lejafar:
I will get that done!

PR #172, opened 2018-06-04
Added support for thickness computations.

willmycroft:
Hi Mike,

I hope you are well. I've written some code for computing thickness using both ray casting and maximal tangential sphere methods which seems to be robust and reasonably efficient.

Let me know what you think.

Thanks
Will

### DIFF
--- a/tests/test_thickness.py
+++ b/tests/test_thickness.py
@@ -1,0 +1,67 @@
+import generic as g
+
+class ThicknessTest(g.unittest.TestCase):
+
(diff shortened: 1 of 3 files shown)


mikedh:
This looks great, thanks! 

I've been testing and I noticed the sphere method is returning the sphere radius, rather than the sphere diameter- I'll merge and change it to diameter.

willmycroft:
You're right, it makes more sense for thickness() to return the sphere diameter.

PR #175, opened 2018-06-06
Added functionality for curvature computations.

willmycroft:
Hi Mike,

Added some functions for calculating discrete curvature measures following this paper:

http://www.cs.jhu.edu/~misha/Fall09/Steiner03.pdf

Hope this is of interest to you.

Thanks
Will


mikedh:
This looks awesome, thanks! Based on a quick read of the [wikipedia article on Gaussian curvature](https://en.wikipedia.org/wiki/Gaussian_curvature), I'll try to check the following:

- verify toroids have negative curvature on the hole- side and positive curvature on the outer side
- verify spheres have Gaussian curvature everywhere of `1/r**2` for different radii
  - check with subdivisions and rigid transforms
- verify cylinders have Gaussian curvature of 0 everywhere

Are there any other easy ground truth examples you can think of before I merge?

Thanks!
Mike

willmycroft:
No problem, happy to help - this is a great project you have. Those test
cases seem sensible, I'd also check that a plane has curvature zero.

One thing to note is that these are "curvature measures" produced by
averaging a curvature tensor across a mesh. The curvatures of a triangular
mesh are zero within the faces, but infinite on the edges (and vertices).
However, the integral over any region of non-zero measure is finite and
well defined. The region I have been using is the intersection of the mesh
with a ball of a given radius. To determine the curvatures, you should thus
divide by the area of this region, as in examples/curvature.ipynb. For an
arbitrary mesh, computing the area of this region exactly is
computationally intensive.

For a less computationally intensive approach which I would be happy to add
is to take the region to be the n-ring of a face. However, this wouldn't
produce accurate computations for irregular meshes.

Sorry this explanation was a bit rushed, let me know if anything is unclear
and i'll try explain.

Thanks
Will

willmycroft:
Just added a small change to make the vertex defects computation much faster by looping over faces and summing the contributions rather than looping over all vertices.

mikedh:
Sounds good! I'm still going through, I'll try to get this in tomorrow.

Oh interesting RE: vertex defect, I wonder if there's a way to use the sparse face matrix to avoid the loop, like: `mesh.faces_sparse.dot(mesh.face_angles).sum(axis=1)`

That doesn't give an equivalent result and isn't quite right, but I wonder if something like that could work.

willmycroft:
I had a read about sparse matrices and managed to implement a solution. The speed up was remarkable. Vertex defects now computes about 10000x faster for a large mesh (icosphere subdivided twice). Thanks for the idea!

mikedh:
Nice! scipy.sparse is really quick, I had a similar experience with `mean_vertex_normals`.

Thanks for the PR!

aymanesouani:
I will do some comparaison between these two methods (the one I coded and the one in Trimesh).

jrdkr:
> I will do some comparaison between these two methods (the one I coded and the one in Trimesh).

Hey,
I am also interested in your method/discrete curvature estimation. Currently done few tests on a cylinder with different approaches. Problem is that all aproaches led to errors at the boundaries of the cylinders or plates. Eg. the gaussian curvature for a cylinder and plane should be zero at every vertex. Do you have any idea or are you familiar with an approach to solve this problem (expect of applying a filter to vertices close to the boundary)?

Thank you in advance!

PR #176, opened 2018-06-07
Add MANIFEST.in

pauldmccarthy:
Currently the trimesh source distribution (i.e. what you get from `python setup.py sdist`) doesn't contain the trimesh license or readme files. We can use a manifest to ensure that this happens.


mikedh:
LGTM- thanks!

PR #177, opened 2018-06-08
 Add ply export support for 1) generating ascii ply 2) including vertex normals

vlimit:
Hi,

This is a minor change to allow the generation of ascii ply files + the inclusion of vertex normals. There must be a better way to implement ndtoascii. Happy to learn (I'm not strong on numpy).

Hope the change is of value. Tested a combination of ply formats (vertex/face colours, ascii/binary, w/wo normals and all opened successfully in meshlab.

Thanks,
Stu

### DIFF
--- a/trimesh/io/ply.py
+++ b/trimesh/io/ply.py
@@ -58,21 +58,34 @@ def load_ply(file_obj, *args, **kwargs):
     return kwargs
 
 
-def export_ply(mesh):
+def export_ply(mesh, **kwargs):
     '''
     Export a mesh in the PLY format.
 
     Parameters
     ----------
     mesh : Trimesh object
+    encoding : ['ascii'|'binary_little_endian']
+    vertex_normal : include vertex normals
 
     Returns
     ----------
     export : bytes of result
     '''
+
+    encoding = 'binary_little_endian'
+    if kwargs.has_key('encoding'):
+        encoding = kwargs['encoding']
+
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome, thanks! 

1) You may want to replace the custom `ndtoascii` with [`trimesh.util.array_to_string`](https://github.com/mikedh/trimesh/blob/master/trimesh/util.py#L879-L946) as it *might* be faster and avoids implementing some logic in the exporter.
2) Would you mind adding a unit test (maybe in `test_export`) to make sure the vertex normals are surviving the round trip?

This looks like a solid improvement, thanks for the PR!

mikedh:
Actually no worries, I'll add some additional tests/logic. Thanks for the PR!

mikedh:
Included in master as of https://github.com/mikedh/trimesh/commit/f7595e89c2e216baf62e61101dc374fe84e642a2

vlimit:
Thanks for picking up the PR. I have been away from email over the
weekend, so just catching up on this.
Will do as you suggest wrt the stringification (thanks for the tip) and
the unit tests.

Guptajakala:
Hi, is the feature "including vertex normals" already existing in version 3.7.12?
I couldn't export ply file with normals by 
```
mesh = trimesh.Trimesh(vertices=pts, vertex_normals=normals, vertex_colors=colors)
```

PR #182, opened 2018-06-20
Add local voxelisation support

willmycroft:
Hi Mike,

I found myself needing to voxelise large meshes at small values of pitch, and running into memory errors. I've added some code to perform voxelisation in a local region of a point which was very helpful for my applications.

In doing so, I also found a few small quirks in the voxelisation / remesh code.

1) The loop in subdivide_to_size subdivides at the end of the loop, and so the range should be max_iter + 1. This gives the intuitive result that setting max_iter = 0 returns the original mesh (if max_edge is large).

2) The line `hit = np.vstack((np.ceil(hit), np.floor(hit))).astype(int)` seems strange to me. I think you are doing this so the voxelisation process preserves connectivity. However, when the edge_factor is larger than 1, the 6-connectivity of the voxels (i.e. adjacent up, down, left etc.) is already guaranteed. Moreover, the code is weirdly asymmetrical, and it would seem more natural to do something like scipy.ndimage.morphology.binary_dilation if you were to go down this route (which I think is unnecessary).

Additionally, I added support for setting max_iter = None in voxelize_subdivide where it determines the necessary value of max_iter based on the edge lengths.

Let me know if any of this is unclear.

Thanks
Will


mikedh:
LGTM- thanks! RE: `hit = np.vstack((np.ceil(hit), np.floor(hit))).astype(int)` yeah, I agree that's weird, I'll make a quick pass through and change it to `np.round(hit)`.

PR #186, opened 2018-06-27
Make it easy to align parallel cross sections of a mesh in 2D space

nick-parker:
Added an optional arg to Path3D.to_planar which specifies where the origin should be in the 2D output. This lets you take many parallel slices of a Trimesh and compare them simply in 2D.

The current recommended workflow as I understand it is:
* Put your first section thrrough to_planar and save to_3d from that
* Invert to_3d to get to_2d back
* Use that to_2d on all other sections.

That doesn't work well if you want valid to_3d transforms for returning the other sections to their original positions after some 2D modification though. to_3d is generated directly from to_2d in this function, so all the other sections will actually return the same to_3d matrix as the first one, and when you put your sections back into 3D they'll all be coplanar.


mikedh:
Thanks for the PR! I actually added a `section_multiplane` method in https://github.com/mikedh/trimesh/commit/07544492f4e8e8992ee4e65a22b2e06ebcfa760f which hopefully makes the multiple parallel slicing operation easier, and it should be faster as it's caching and offsetting the dot products: 
```
# list of Path2D objects
sections = mesh.section_multiplane(plane_origin=mesh.bounds[0], 
                                   plane_normal=[0,0,1], 
                                   heights=np.linspace(0,1,10))
# return them to Path3D objects
section_3D = [i.to_3D(i.metadata['to_3D']) for i in sections if i is not None]
```


As for the args in `to_planar`, yeah they are currently pretty weird: when you pass a normal, it was only using that for a sign hint (still fitting a plane). I'd like to make more explicit if possible. I'm a little confused by what's happening to the origin in this update:
`   C += (origin - C) - N * np.dot(N, origin - C) `

For updating the args to be clearer, these would be my preferences:

### Option 1:
- only allow explicit matrix (`to_2D`) to be passed; otherwise fit a plane

### Option 2:
- if `to_2D` is passed it is used and other args are discarded
- `normal` and `origin` must be passed together, and they define the plane directly using `plane_transform`
- if `normal_hint` is passed, it defines the sign of the fit plane

What do you think? 

Best,
Mike

nick-parker:
Hey Mike,

Sorry for the long delay, I'm the lone dev at a startup so this slid way down the queue. I was going to start using multiplane myself before making a choice here. I ran into some issues which I haven't quite finished hunting down, I'll make a ticket for that though.

I'd say option 1 is probably best, I can't think of a use case for option 2 that isn't better served by multiplane.

PR #188, opened 2018-07-03
Registration of point clouds to meshes

willmycroft:
Hi Mike,

Quick bit of code I wrote to align point clouds with meshes, thought it would fit nicely into the library.

Thanks
Will


mikedh:
Looks great, thanks! Some minor thoughts before I merge:

- unfortunately `@` is Python 3.5+ only, and `trimesh` targets 2.7+, `np.dot`
- `procrustes(X, Y...)`: X and Y are a little confusing as they are both (n, 3) XYZ point clouds: maybe change to `a` and `b`?
- procrustes replaces the Horn method (in https://github.com/mikedh/trimesh/blob/master/trimesh/points.py#L151): I'll benchmark and then either move absolute_orientation into registration (as like `horns_method` or something) or just remove it if it is much slower and worse
- I should really tune those nearest point on surface queries 😬

mikedh:
Looks great- thanks!

PR #194, opened 2018-07-19
Fixed a bug with multiplane slicing.

Artur-Sampaio:
Fixed a bug with multiplane slicing that caused the wrong transformations and offset.

1) The original base_transform was a transformation from the original bounds to the z = 0 plane ("3D -> 2D"). It is now its reverse, projecting a 2D slice on the z = 0 plane to the original bounds ("2D -> 3D"). This is required to produce the correct to_3D and to_2D matrices.

2) Moreover, the stored heights should be stored with respect to the first height value (since the inverse of the new base transform already moves that distance to the origin).

The bug is unnoticeable if the model begins at z = 0, the plane_normal equals [0,0,1], and the heights vector begins in 0.

### DIFF
--- a/trimesh/intersections.py
+++ b/trimesh/intersections.py
@@ -221,15 +221,15 @@ def mesh_multiplane(mesh,
     plane_normal = util.unitize(plane_normal)
     plane_origin = np.asanyarray(plane_origin,
                                  dtype=np.float64)
-    heights = np.asanyarray(heights, dtype=np.float64)
+    heights = np.asanyarray(heights - heights[0], dtype=np.float64)
 
     # dot product of every vertex with plane
     vertex_dots = np.dot(plane_normal,
                          (mesh.vertices - plane_origin).T)
 
-    # reconstruct transforms for each 2D section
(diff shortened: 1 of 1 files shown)


mikedh:
Awesome, good catch!

As I was adding additional tests that should fail in this case (https://github.com/mikedh/trimesh/commit/3d4d37a2b3afabbb4ac278e7659262e8d1a9e915), I totally get inverting `base_transform`, as the new tests (sectioning along non- XY parallel planes) fail in that case. 

I'm a little confused by:
>> Moreover, the stored heights should be stored with respect to the first height value (since the inverse of the new base transform already moves that distance to the origin).

The `plane_origin` isn't offset along `plane_normal` by `heights[0]`, so wouldn't this be odd in the case where I pass non monotonically increasing heights (ie `heights = [100.0, 0.0, 1.0]`)? The desired behavior would be in that case the second returned section is exactly on the plane specified by normal/origin.

Even though I'm re- ordering the `heights` in the new tests they all pass with or without the `heights -= heights[0]` offset which I don't fully understand (it seems like it should fail in one or the other case). 

Thanks for the PR!

PR #195, opened 2018-07-20
Fixed a bug with multiplane slicing on models that rest entirely on the negative half-space.

Artur-Sampaio:
The following call would previously produce only empty sections on the attached object:

mesh = trimesh.load('cube_negative_z.obj')
z_extents = mesh.bounds[:,2]
z_levels  = np.arange(*z_extents, step=1)
sections = mesh.section_multiplane(plane_origin=mesh.bounds[0], plane_normal=[0,0,1], heights=z_levels)

[cube_negative_z.zip](https://github.com/mikedh/trimesh/files/2214635/cube_negative_z.zip)

### DIFF
--- a/trimesh/intersections.py
+++ b/trimesh/intersections.py
@@ -12,6 +12,7 @@ from . import util
 from . import geometry
 from . import grouping
 from . import transformations
+from . import points
 
 
 def mesh_plane(mesh,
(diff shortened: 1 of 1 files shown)


mikedh:
Hey, thanks for the PR! Sorry for the delay. I think I've been thinking about `heights` as 'offset from passed plane' rather than 'absolute height in model frame.' In your example if you treat it that way it works as expected:

```
    mesh = trimesh.load('cube_negative_z.obj')

    # extents is the (3,) edge lengths of the AABB (always positive)              
    z_levels = np.arange(mesh.extents[2], step=1.0)

    # take sections offset from the plane lying with a normal               
    # along positive Z and an origin at the lowest point of the mesh        
    sections = mesh.section_multiplane(
        plane_origin=mesh.bounds[0],
        plane_normal=[0,0,1],
        heights=z_levels)


    # return the 2D sections to 3D to show together 
    # this was only added last week so you might have to update via pip                        
    trimesh.Scene([i.to_3D() for i in sections]).show()

```

I should possibly change `heights` to `offsets` for clarity here. The behavior I was after was to always return a section on the passed plane when a zero height/offset value is passed.

PR #204, opened 2018-08-14
Add (periodic) callback feature to the scene viewer

gibsjose:
This adds a callback feature in which a callback method can be specified
and will be called every time the scene object's `_update_meshes()` method
is called. The callback can either be periodic, with a default rate
of 60Hz, or if the callback period is specifically set to 0, the
callback method will only be called when the scene is manipulated
(e.g. by rotating the viewer or clicking in the scene).


mikedh:
Looks great! Don't worry about the CI, it looks like `svg.path` released an update, I'll fix that now then merge this shortly.

gibsjose:
Excellent, thanks!

PR #207, opened 2018-08-19
Allow Path3D export to .glb and display in notebook

### LINKED ISSUE #205: display Path3D in jupyter notebook
It's possible to .show() a Path3D object using pyglet, but if you're in the notebook you get:

`AttributeError: 'Path3D' object has no attribute 'visual'`

Presumably this is because the path needs to be "inflated" into a mesh in order to export it to gltf? I guess doing that would be an enhancement.

In the meantime, a better error might be nice -- something that explains that 3d paths can be shown using pyglet but not in a browser.

foobarbecue:
Fixes #205

### DIFF
--- a/trimesh/io/gltf.py
+++ b/trimesh/io/gltf.py
@@ -58,7 +58,7 @@ def export_gltf(scene):
 
         # create the buffer views
         current_pos = 0
-        for j in range(2):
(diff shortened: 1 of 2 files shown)


mikedh:
Looks great, thanks!

foobarbecue:
You're welcome -- trimesh has been very useful so I hope to contribute more in the future.

PR #229, opened 2018-10-09
add slicing mesh with plane that returns remaining mesh

mjd3:
Helpful for cases where you want to actually cut a mesh and get a mesh back, instead of just the intersection with the plane. Thanks to @mmatl for some help with this as well.


mikedh:
This looks awesome, thanks! I'll try to go through and test today or tomorrow and get it merged. Seems like someday someone could also potentially use `trimesh.creation.triangulate_polygon` to cap the ends and get a watertight result too which might be neat.

Best,
Mike

mikedh:
Merged into master and released in https://github.com/mikedh/trimesh/commit/b15a181e1ff26c994561f116bd8ca9f23c5f5661.

Thanks for the PR!

PR #231, opened 2018-10-13
Fix bug in mesh loader

mmatl:
There was a bug in the existing mesh loading code that was triggered whenever I used a string instead of a file as the source for a mesh loader. This PR fixes it. Basically, just an inconsistent number of return arguments in `parse_file_args()`.


mikedh:
Good catch! Looks like this only occurs when the file doesn't exist, and the file extension is passed?

mmatl:
Yep!

Issue #249, opened 2018-11-09
Big Endian failures


hroncok:
Another issue that came up in Fedora. We build packages on the first available builder and sometimes that's Big Endian (ppc64 on Fedora 28 or s390x anywhere). Apparently, the tests fail a lot on those.

Mos likely `load_stl_binary` is broken and `load_stl_ascii` is called for binary STLs on Big Endian.

```
+ /usr/bin/python3 -m pytest -v -k 'not test_export and not test_remote'
============================= test session starts ==============================
platform linux -- Python 3.6.6, pytest-3.4.2, py-1.5.4, pluggy-0.6.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /builddir/build/BUILD/trimesh-2.35.24, inifile:
collecting ... collected 193 items
tests/test_boolean.py::BooleanTest::test_boolean PASSED                  [  0%]
tests/test_boolean.py::BooleanTest::test_multiple PASSED                 [  1%]
tests/test_bounds.py::BoundsTest::test_2D FAILED                         [  1%]
tests/test_bounds.py::BoundsTest::test_cylinder FAILED                   [  2%]
tests/test_bounds.py::BoundsTest::test_obb_mesh FAILED                   [  2%]
tests/test_bounds.py::BoundsTest::test_obb_order FAILED                  [  3%]
tests/test_bounds.py::BoundsTest::test_obb_points FAILED                 [  3%]
tests/test_cache.py::CacheTest::test_contiguous PASSED                   [  4%]
tests/test_cache.py::CacheTest::test_hash FAILED                         [  4%]
tests/test_cache.py::CacheTest::test_track PASSED                        [  5%]
tests/test_collision.py::CollisionTest::test_collision PASSED            [  6%]
tests/test_collision.py::CollisionTest::test_distance PASSED             [  6%]
tests/test_collision.py::CollisionTest::test_scene PASSED                [  7%]
tests/test_convex.py::ConvexTest::test_convex FAILED                     [  7%]
tests/test_convex.py::ConvexTest::test_primitives PASSED                 [  8%]
tests/test_convex.py::ConvexTest::test_projections FAILED                [  8%]
tests/test_creation.py::CreationTest::test_annulus PASSED                [  9%]
tests/test_creation.py::CreationTest::test_path_sweep PASSED             [  9%]
tests/test_creation.py::CreationTest::test_soup PASSED                   [ 10%]
[...]

PR #250, opened 2018-11-10
WIP: Properly support Big Endian arches when loading binary meshes

hroncok:


hroncok:
Will fix https://github.com/mikedh/trimesh/issues/249 but still need more changes.

mikedh:
Looks great, thanks! I'm happy to merge this whenever, just let me know when you're ready.

mikedh:
I'm going to merge as it seems like a standalone improvement and we can address additional BE problems via another PR. Thanks for the fix!

hroncok:
OK. I'm not sure if/when I'll get to the rest of it :(

Issue #253, opened 2018-11-20
Are there some idioms to use camera intrinsic parameter to get rendered image?


wkentaro:
I'd like to use intrinsic parameter `K` to get rendered image. Does anyone have some idiom to do this?

mikedh:
Hey, I think this would require some wangling of ‘gl_init’ in viewer, this article looks pretty good: http://ksimek.github.io/2013/06/03/calibrated_cameras_in_opengl/

If you wanted to add it cleanly to ‘trimesh’ It would be pretty cool to have ‘Camera’ objects which pass their intrinsic matrix to the pyglet code in ‘scene.viewer’. 

Oh it’s also possible https://github.com/BerkeleyAutomation/meshrender does this or has something you could use as a starting point.

wkentaro:
Thank you for the nice feedback.
I sent PR here https://github.com/mikedh/trimesh/pull/254, which supports fx, fy to get correct fov but does not support cx, cy and skew.

wkentaro:
Closed via #254

PR #254, opened 2018-11-20
Add trimesh.scene.camera.Camera to set camera parameters

wkentaro:
Close https://github.com/mikedh/trimesh/issues/253

This only supports some parameters in camera intrinsic matrix.

### DIFF
--- a/tests/test_camera.py
+++ b/tests/test_camera.py
@@ -1,0 +1,39 @@
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+import numpy as np
+
+
+class CameraTests(g.unittest.TestCase):
+
+    def test_K(self):
+        resolution = (320, 240)
+        fovxy = (60, 40)
+        camera = g.trimesh.scene.Camera(
+            resolution=resolution,
+            fovxy=fovxy,
+        )
+        K_expected = np.array([
+            [277.128,       0, 160],
+            [      0, 329.697, 120],
+            [      0,       0,   1],
+        ], dtype=np.float64)
+        np.testing.assert_allclose(camera.K, K_expected, rtol=1e-3)
+
+    def test_consistency(self):
+        resolution = (320, 240)
+        fxfy = None
+        fovxy = (60, 40)
+        camera = g.trimesh.scene.Camera(
+            resolution=resolution,
+            fxfy=None,
+            fovxy=fovxy,
+        )
+        camera = g.trimesh.scene.Camera(
+            resolution=resolution,
+            fxfy=camera.fxfy,
+            fovxy=None
+        )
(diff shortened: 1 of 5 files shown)


wkentaro:
## Example

You need #255 

```python
#!/usr/bin/env python

import numpy as np
import trimesh
import trimesh.transformations as tf


trimesh.util.attach_to_log(level='WARNING')


def get_homography_rpy(roll, pitch, yaw, distance, center):
    translation = tf.translation_matrix(center)
    translation[2][3] += distance  # z from center
    rotation = (
        tf.rotation_matrix(roll, [1, 0, 0], point=center) @
        tf.rotation_matrix(pitch, [0, 1, 0], point=center) @
        tf.rotation_matrix(yaw, [0, 0, 1], point=center)
    )
    transform = translation @ rotation
    return transform


if __name__ == '__main__':
    extents = (0.5, 0.5, 0.5)
    box = trimesh.creation.box(extents)
    box.visual.face_colors = [255, 0, 0, 127]

    scene = trimesh.Scene()
    scene.add_geometry(box, node_name='box')

    scene.add_geometry(trimesh.creation.axis(np.eye(4)))

    roll, pitch, yaw = 0, 0, np.deg2rad(45)
    distance = 0.9
    center = (0, 0, 0)
    transform = get_homography_rpy(roll, pitch, yaw, distance, center)
    print('Using RPY (Red)')
    print('Distance:', distance)
    print('RPY:', (roll, pitch, yaw))
    print(transform)
    scene.add_geometry(
        trimesh.creation.axis(transform, origin_color=(255, 0, 0))
    )

    # Camera for SceneNet RGB-D
    camera = trimesh.scene.Camera(
        width=320,
        height=240,
        fovx=60,
        fovy=45,
        matrix=transform
    )
    scene.set_camera(camera=camera)

    # scene.graph.update(
    #     frame_from='camera',
    #     frame_to=scene.graph.base_frame,
    #     matrix=transform,
    #     geometry='camera',
    # )

    # scene.show()

    import io
    import PIL.Image
    with io.BytesIO() as f:
        f.write(scene.save_image((camera.width, camera.height)))
        PIL.Image.open(f).save('result.png')
```

mikedh:
Awesome looks great! Couple minor thoughts:

- Could you replace fx/fy with a (2,) float array and fovx/fovy with (2,) float array?  `width`/`height` could be combined into a `resolution` parameter
- How about a simple unit test which specifies parameters in varies ways then asserts K is close to ground truth?
- Could you add a docstring to `Camera.__init__` describing what the parameters do?

wkentaro:
I'm not sure what is the right name for 'fx/fy' and 'fovx/fovy'.
Currently, I'm using `fxfy` and `fovxy`.

wkentaro:
I think I addressed all comments.

PR #259, opened 2018-11-24
Fix camera resolution check in viewer.py

wkentaro:
To fix below:

```
  File "check_look_at.py", line 107, in <module>
    scene_camera.show(start_loop=False)
  File "/home/wkentaro/SceneNetRGBD/scenenetrgb-d/python/src/trimesh/trimesh/scene/scene.py", line 820, in show
    return SceneViewer(self, **kwargs)
  File "/home/wkentaro/SceneNetRGBD/scenenetrgb-d/python/src/trimesh/trimesh/scene/viewer.py", line 49, in __init__
    if resolution != scene.camera.resolution:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

### DIFF
--- a/trimesh/scene/viewer.py
+++ b/trimesh/scene/viewer.py
@@ -46,7 +46,8 @@ class SceneViewer(pyglet.window.Window):
 
         if scene.camera is not None:
             if resolution is not None:
-                if not all(resolution == scene.camera.resolution):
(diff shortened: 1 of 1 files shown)


wkentaro:
Fixed!

mikedh:
Awesome thanks for the fix!

PR #260, opened 2018-11-24
Fix indices_to_points to make it consistent to VoxelMesh.as_boxes

wkentaro:
I'm not so sure, but usually voxel point coordinate should be calculated as:

  (index - 0.5) * pitch + origin = index * pitch + origin - pitch / 2.0

In VoxelMesh.as_boxes this formulation is used, but the matrix_to_points function does not.

### DIFF
--- a/trimesh/voxel.py
+++ b/trimesh/voxel.py
@@ -273,7 +273,7 @@ class VoxelMesh(Voxel):
         else:
             filled = self.sparse_surface
         # center points of voxels
-        centers = indices_to_points(filled, pitch, self.origin)
+        centers = indices_to_points(indices=filled, pitch=self.pitch, origin=self.origin)
         mesh = multibox(centers=centers, pitch=self.pitch)
         return mesh
 
--- a/trimesh/voxel.py
+++ b/trimesh/voxel.py
@@ -112,8 +112,7 @@ class Voxel(object):
         point = np.asanyarray(point)
         if point.shape != (3,):
             raise ValueError('to_index requires a single point')
-        index = np.round((point - self.origin) /
-                         self.pitch).astype(int)
+        index = np.round((point - self.origin) / self.pitch + 0.5).astype(int)
         index = tuple(index)
         return index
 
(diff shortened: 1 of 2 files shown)


mikedh:
Yeah, you're right it should definitely be consistent. Maybe we should make two functions, `points_to_indices` and `indices_to_points`  to avoid screwing up the logic, then we can add a round trip or two into unit tests. 

Some quick thoughts:

- probably want to make sure indices is float before doing the operation, and check the other inputs
- in `as_boxes`, it needs to be `self.pitch`, not `pitch` 
- generally probably want to use kwargs rather than args when calling functions (`pitch=self.pitch` vs `pitch`), makes it more explicit and CI will start failing if someone decides to rearrange args.

trimesh/voxel.py
```diff
@@ -274,8 +273,7 @@ def as_boxes(self, solid=False):
         else:
             filled = self.sparse_surface
         # center points of voxels
-        centers = (filled * self.pitch).astype(np.float64)
-        centers += self.origin - (self.pitch / 2.0)
+        centers = indices_to_points(filled, pitch, self.origin)
```
mikedh:
```suggestion
        centers = indices_to_points(indices=filled, pitch=self.pitch, origin=self.origin)
```

trimesh/voxel.py
```diff
@@ -561,6 +559,24 @@ def fill_voxelization(occupied):
...
+    ----------
+    indices: (q, 3) int, index of voxel matrix (n,m,p)
+    pitch: float, what pitch was the voxel matrix computed with
+    origin: (3,) float, what is the origin of the voxel matrix
+
+    Returns
+    ----------
+    points: (q, 3) list of points
+    """
+    points = (indices - 0.5) * pitch + origin
```
mikedh:
```suggestion
    indices = np.asanyarray(indices, dtype=np.float64)
    origin = np.asanyarray(origin, dtype=np.float64)
    pitch = float(pitch)

    if origin.shape != (3,):
        raise ValueError('origin incorrect shape!')

    points = (indices - 0.5) * pitch + origin
```

trimesh/voxel.py
```diff
@@ -575,7 +591,8 @@ def matrix_to_points(matrix, pitch, origin):
     ----------
     points: (q, 3) list of points
     """
-    points = np.column_stack(np.nonzero(matrix)) * pitch + origin
+    indices = np.column_stack(np.nonzero(matrix))
+    points = indices_to_points(indices, pitch, origin)
```
mikedh:
```suggestion
    points = indices_to_points(indices=indices, pitch=pitch, origin=origin)
```

wkentaro:
I added `points_to_indices` and tests for both functions. Could you review again?

mikedh:
Looks great, thanks!

PR #264, opened 2018-11-29
Show filename when loading fails

wkentaro:


mikedh:
Merged into master, thanks! It doesn't make much difference but I changed it to new- style string formatting (i.e. `'hi {}'.format('world')`) just for the sake of consistency.

wkentaro:
Thanks!

PR #265, opened 2018-11-30
Slice with multiple planes and performance tweaks

mjd3:
Two main things:
1) I found that slicing with multiple planes (not necessarily parallel) gets a speed up when not creating a new Trimesh object every time, so I've expanded the slice_plane method to slice_planes, where you can instead pass an array of origins/normals. I left the old method for compatibility sake, but feel free to do what you want with it (e.g., the new method subsumes the old, so eventually might want to just switch over to the new one)!
2) Some minor performance tweaks. The biggest one was replacing np.unique with np.bincount. Since the faces arrays are all integer-valued, it seems like np.bincount/cumsum perform a lot better than np.unique (e.g., 2-3x faster for small meshes, 10x faster for large meshes). Here are some examples for a smaller mesh (10s of faces) and a larger mesh (~40,000) faces (using kernprof):
headings
np_unique
np_unique_big
unique_replacement
unique_replacement_big

Happy to answer questions and I hope it helps!

### DIFF
--- a/tests/test_section.py
+++ b/tests/test_section.py
@@ -176,6 +176,32 @@ class SliceTest(g.unittest.TestCase):
         # should be lots of stuff at the plane and nothing behind
         assert g.np.isclose(dot.min(), 0.0)
 
+        # Cut part of top off of box with multiple planes at once
+        # and make sure bounds and number of faces is correct
+        plane_origins = [mesh.bounds[1] - 0.05, mesh.bounds[1] - 0.05]
+        plane_normals = [g.np.array([0, 0, 1]), g.np.array([0, 1, 0])]
+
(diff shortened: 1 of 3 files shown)


mjd3:
Seems like the issue might be here (https://github.com/numpy/numpy/issues/4384) for the Windows build failing. Do you have any thoughts on this?

mikedh:
Looks like the wrap works in appveyor: https://ci.appveyor.com/project/mikedh/trimesh/build/job/9m2b921oekpd8455

I though about doing a check for 32 bit systems (`np.intp == np.int32`) but thought it might be a little brittle, here it just falls back to `np.unique`.

PR #267, opened 2018-12-05
Fix missing arg in call to unique_bincount

mmatl:
Missing kwarg `mincount` to `unique_bincount` was causing an error on master when removing unreferenced vertices.


mikedh:
Good catch, I'll also add a test for `remove_unreferenced_vertices`. Thanks for the PR!

PR #268, opened 2018-12-06
further speedup of nsphere

nbraem:
Eliminate loop in calculation of radii_2


mikedh:
Awesome, nice find! It looks like in a [quick benchmark](https://gist.github.com/mikedh/1ec968830eeed591989fd962aead2369#file-benchcdist-py) cdist is pretty much always faster except on very large arrays where looping is still the only way to get a result (at least in my crappy VM):
```
a.shape: (10, 3)   b.shape: (10, 3)
tile: 0.00016617897199466825
cdist: 0.00015958899166435003
loop: 0.0006842719740234315

a.shape: (10, 3)   b.shape: (100, 3)
tile: 0.00028696999652311206
cdist: 0.00014902098337188363
loop: 0.007028821972198784

a.shape: (100, 3)   b.shape: (10, 3)
tile: 0.00026292604161426425
cdist: 0.00013586296699941158
loop: 0.0007681719725951552

a.shape: (1000, 3)   b.shape: (1000, 3)
tile: 0.18622937001055107
cdist: 0.011909001041203737
loop: 0.16162009502295405

a.shape: (10000, 3)   b.shape: (1000, 3)
tile: 1.821985297021456
cdist: 0.22127155697671697
loop: 0.9185473470133729

a.shape: (10000, 3)   b.shape: (100000, 3)
tile: inf
cdist: inf
loop: 90.87495470099384
```

I might get rid of the tile method completely, always try `cdist` (no element count threshold) then only fall back to loops after an explicit MemoryError. Thanks for the PR!

PR #272, opened 2018-12-15
Support loading ShapeNet models: support Kd in mtl and some fixes in 'vt' and 'f'

wkentaro:
With this change, trimesh now supports loading textures and colors for models in ShapeNetSem.v0.
(This information was the key for uv_to_color with uv out of range from [0, 1]: https://shapenet.org/qaforum/index.php?qa=15&qa_1=why-is-the-texture-coordinate-in-the-obj-file-not-in-the-range)

```python
import pathlib
import numpy as np
import trimesh

# trimesh.util.attach_to_log()

scene = trimesh.Scene()

models_dir = pathlib.Path('ShapeNetSemModels')
for cad_file in models_dir.glob('*.obj'):
    print(cad_file)

    cad = trimesh.load(str(cad_file))

    # centerize
    xyz_min = np.min(cad.vertices, axis=0)
    xyz_max = np.max(cad.vertices, axis=0)
    xyz_center = (xyz_max + xyz_min) / 2.
    cad.vertices -= xyz_center

    max_size = (xyz_max - xyz_min).max()
    cad.vertices /= max_size  # normalize

    print(cad.bounding_box.primitive.extents)

    # shift for visualization
    x = len(scene.geometry) % 4
    y = len(scene.geometry) // 4
    transform = trimesh.transformations.translation_matrix(
        [1.5 * x, 1.5 * y, 0]
    )
    scene.add_geometry(cad, transform=transform)

    if len(scene.geometry) == 16:
        break
scene.show()
```

Of course, it works with YCB object models:

```python
import pathlib
import numpy as np
import trimesh

# trimesh.util.attach_to_log()

scene = trimesh.Scene()

models_dir = pathlib.Path('YCB_Video_Models')
for obj_dir in sorted(models_dir.iterdir()):
    cad_file = obj_dir / 'textured.obj'
    print(cad_file)

    cad = trimesh.load(str(cad_file))

    # centerize
    xyz_min = np.min(cad.vertices, axis=0)
    xyz_max = np.max(cad.vertices, axis=0)
    xyz_center = (xyz_max + xyz_min) / 2.
    cad.vertices -= xyz_center

    max_size = (xyz_max - xyz_min).max()
    cad.vertices /= max_size  # normalize

    print(cad.bounding_box.primitive.extents)

    # shift for visualization
    x = len(scene.geometry) % 4
    y = len(scene.geometry) // 4
    transform = trimesh.transformations.translation_matrix(
        [1.5 * x, 1.5 * y, 0]
    )
    scene.add_geometry(cad, transform=transform)

    if len(scene.geometry) == 16:
        break
scene.show()
```

### DIFF
--- a/trimesh/io/wavefront.py
+++ b/trimesh/io/wavefront.py
@@ -1,3 +1,4 @@
+import collections
 import os.path
 
 import numpy as np
@@ -12,7 +13,7 @@ from .. import util
 from .. import visual
 
 
-def _get_vertex_colors(uv, mtllibs, resolver):
+def _get_vertex_colors(uv, mtllib, resolver):
     """
     Turn loaded UV and materials into vertex colors.
 
@@ -20,7 +21,7 @@ def _get_vertex_colors(uv, mtllibs, resolver):
     ------------
     uv : (n, 2) float
       Texture coordinates
-    mttlibs : list of dict
+    mttlib : dict
       Material data including image names
     resolver : trimesh.visual.resolvers.Resolver
       Object to load data referenced by name
@@ -34,21 +35,21 @@ def _get_vertex_colors(uv, mtllibs, resolver):
         log.warning('pillow is required to load texture')
         return
 
-    success = False
-    vertex_colors = np.zeros((len(uv), 4), dtype=np.uint8)
+    if np.isnan(uv).all() and mtllib['Kd']:
+        color = np.array(mtllib['Kd'], dtype=np.float64)
+        if len(color) == 3:
+            color = np.hstack((color, [1]))
+        color = (color * 255).astype(np.uint8)
+        vertex_colors = np.full((len(uv), 4), color, dtype=np.uint8)
+        return vertex_colors
 
-    for mtl in mtllibs:
-        # get the data for the image
-        texture_data = resolver.get(mtl['map_Kd'])
-        # load the actual image
-        texture = Image.open(util.wrap_as_stream(texture_data))
-        # append the loaded colors
-        vertex_colors += visual.uv_to_color(uv, texture)
-        success = True
+    # get the data for the image
+    texture_data = resolver.get(mtllib['map_Kd'])
+    # load the actual image
+    texture = Image.open(util.wrap_as_stream(texture_data))
+    # append the loaded colors
+    vertex_colors = visual.uv_to_color(uv, texture)
 
-    # nothing actually loaded
-    if not success:
-        return
     return vertex_colors
 
 
@@ -63,16 +64,15 @@ def parse_mtl(mtl):
 
     Returns
     ------------
-    data : dict
-      Has keys map_Kd
+    mtllibs : list of dict
+      Each dict has keys: newmtl, map_Kd, Kd
     """
     # decode bytes if necessary
     if hasattr(mtl, 'decode'):
         mtl = mtl.decode('utf-8')
 
-    # initial data layout
-    data = {'newmtl': None,
-            'map_Kd': None}
+    mtllib = None
+    mtllibs = []
 
     # use universal newline splitting
     for line in str.splitlines(str(mtl).strip()):
@@ -82,10 +82,21 @@ def parse_mtl(mtl):
         if len(line_split) <= 1:
             continue
         # store the keys
-        if line_split[0] in data:
-            data[line_split[0]] = line_split[1]
+        key = line_split[0]
+        if key == 'newmtl':
+            if mtllib:
+                mtllibs.append(mtllib)
+            mtllib = {'newmtl': line_split[1],
+                      'map_Kd': None,
+                      'Kd': None}
+        elif key == 'map_Kd':
+            mtllib[key] = line_split[1]
+        elif key == 'Kd':
+            mtllib[key] = [float(x) for x in line_split[1:]]
+    if mtllib:
+        mtllibs.append(mtllib)
 
-    return data
+    return mtllibs
 
 
 def load_wavefront(file_obj, resolver=None, **kwargs):
@@ -175,13 +186,11 @@ def load_wavefront(file_obj, resolver=None, **kwargs):
                 texture = np.array(current['vt'], dtype=np.float64)
                 # make sure vertex texture is the right shape
                 # AKA (len(vertices), dimension)
-                try:
-                    texture = texture.reshape((len(vertices), -1))
-                    # save vertex texture with correct ordering
-                    loaded['metadata']['vertex_texture'] = texture[vert_order]
-                except ValueError:
-                    log.warning('texture information seems broken'.format(
-                                file_obj.name))
+                texture = texture.reshape((len(vertices), -1))
+                texture = texture[vert_order]
+                texture = texture[~np.any(np.isnan(texture), axis=1)]
+                # save vertex texture with correct ordering
+                loaded['metadata']['vertex_texture'] = texture
 
             # build face groups information
             # faces didn't move around so we don't have to reindex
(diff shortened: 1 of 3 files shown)


trimesh/io/wavefront.py
```diff
@@ -12,15 +13,15 @@
 from .. import visual
 
 
-def _get_vertex_colors(uv, mtllibs, resolver):
+def _get_vertex_colors(uv, mtllib, resolver):
```
wkentaro: `mtllib` corresponding to `uv` must be one, so it does not need to be a list.

trimesh/io/wavefront.py
```diff
@@ -34,21 +35,21 @@ def _get_vertex_colors(uv, mtllibs, resolver):
         log.warning('pillow is required to load texture')
         return
 
-    success = False
-    vertex_colors = np.zeros((len(uv), 4), dtype=np.uint8)
+    if np.isnan(uv).all() and mtllib['Kd']:
```
wkentaro: `uv` can be `nan` for re-ordering by `vertex_order`.

trimesh/io/wavefront.py
```diff
@@ -63,16 +64,15 @@ def parse_mtl(mtl):
...
     """
     # decode bytes if necessary
     if hasattr(mtl, 'decode'):
         mtl = mtl.decode('utf-8')
 
-    # initial data layout
-    data = {'newmtl': None,
-            'map_Kd': None}
+    mtllib = None
+    mtllibs = []
```
wkentaro: There can be multiple materials declared in .mtl file.

trimesh/io/wavefront.py
```diff
@@ -175,13 +186,11 @@ def append_mesh():
...
                 # AKA (len(vertices), dimension)
-                try:
-                    texture = texture.reshape((len(vertices), -1))
-                    # save vertex texture with correct ordering
-                    loaded['metadata']['vertex_texture'] = texture[vert_order]
-                except ValueError:
-                    log.warning('texture information seems broken'.format(
-                                file_obj.name))
+                texture = texture.reshape((len(vertices), -1))
+                texture = texture[vert_order]
```
wkentaro: For this re-ordering, we need fill `nan` to `current['vt']`.

trimesh/io/wavefront.py
```diff
@@ -193,23 +202,29 @@ def append_mesh():
...
+                    try:
+                        vertex_colors_i = _get_vertex_colors(
+                            uv=uv, mtllib=mtllibs[usemtl], resolver=resolver)
+                    except BaseException:
+                        log.error('failed to load texture: {}'.format(usemtl),
+                                  exc_info=True)
+                        vertex_colors_i = None
+                    if vertex_colors_i is None:
+                        vertex_colors_i = np.full(
+                            (len(uv), 4), (102, 102, 102, 255), dtype=np.uint8)
```
wkentaro: It seems this gray color is the default. Is this correct?

trimesh/io/wavefront.py
```diff
@@ -193,23 +202,29 @@ def append_mesh():
...
+                    vertex_colors = np.r_[vertex_colors, vertex_colors_i]
+                loaded['vertex_colors'] = vertex_colors[vert_order]
 
             # this mesh is done so append the loaded mesh kwarg dict
             meshes.append(loaded)
 
     attribs = {k: [] for k in ['v', 'vt', 'vn']}
     current = {k: [] for k in ['v', 'vt', 'vn', 'f', 'g', 'usemtl']}
+    usemtl_to_findices = collections.defaultdict(list)  # usemtl to 'f' indices
```
wkentaro: This is necessary to get correspondence of usemtl and 'f v/vt/vn' after that, for example:

```
# xxx.obj
usemtl m0
f 0/0/0
f 1/1/1
...
usemtl m1
f 10/10/10
f 11/11/11
```

trimesh/io/wavefront.py
```diff
@@ -246,9 +265,16 @@ def append_mesh():
                     if len(f_split) > 1 and f_split[1] != '':
                         current['vt'].append(
                             attribs['vt'][int(f_split[1]) - 1])
+                    else:
+                        current['vt'].append([np.nan, np.nan, np.nan])
                     if len(f_split) > 2:
                         current['vn'].append(
                             attribs['vn'][int(f_split[2]) - 1])
+                    else:
+                        current['vn'].append([np.nan, np.nan, np.nan])
```
wkentaro: `nan` is used for re-ordering afterwards.

trimesh/visual/texture.py
```diff
@@ -19,10 +19,18 @@ def uv_to_color(uv, image):
     colors : (n, 4) float
       RGBA color at each of the UV coordinates
     """
-    x = (uv[:, 0] * image.width).round().astype(int)
-    y = ((1 - uv[:, 1]) * image.height).round().astype(int)
+    # https://shapenet.org/qaforum/index.php?qa=15&qa_1=why-is-the-texture-coordinate-in-the-obj-file-not-in-the-range  # NOQA
```
wkentaro: it seems opengl supports uv values out of range from [0, 1] by truncating that int value, for example:
- 1.1 -> 0.1
- -0.3 -> 0.7

trimesh/visual/texture.py
```diff
@@ -19,10 +19,18 @@ def uv_to_color(uv, image):
...
+    if not (uv.min() >= 0 and uv.max() <= 1):
+        uv = uv - np.floor(uv)
+
+    x = (uv[:, 0] * (image.width - 1)).round().astype(int)
+    y = ((1 - uv[:, 1]) * (image.height - 1)).round().astype(int)
```
wkentaro: `-1` was missing in previous implementation. Now it supports uv value 0, 1. Previously, it causes out of index error at `image[y, x]`.

trimesh/visual/texture.py
mikedh: Ah that's a little weird, we may want to just clamp them to the range ourselves for the sake of explicitness, like `np.clip(uv, a_min=0.0, a_max=1.0)`

mikedh:
Awesome! Would you mind adding one of the models that failed before to unit tests? I'll go through a little bit and see why Travis is failing.

trimesh/io/wavefront.py
mikedh: Yeah, or since visual is already imported if you prefer you can use `visual.DEFAULT_COLOR`

trimesh/io/wavefront.py
mikedh: Adding NaN to the array makes me slightly nervous- is there any way we could do this with a boolean mask? Though I guess this way most of the lifting is done by the numpy constructor and is probably pretty fast.

trimesh/io/wavefront.py
```diff
@@ -193,23 +202,29 @@ def append_mesh():
...
-                        mtllibs=current_mtllibs,
-                        resolver=resolver)
-                    loaded['vertex_colors'] = vertex_colors
-                    loaded['metadata']['mtllibs'] = current_mtllibs
-                except BaseException:
-                    log.error('failed to load texture!', exc_info=True)
+                texture = np.array(current['vt'], dtype=np.float64)
+                vertex_colors = np.zeros((0, 4), dtype=np.uint8)
+                for usemtl, findices in usemtl_to_findices.items():
+                    uv = texture[findices]
```
mikedh: This is the line killing the builds with:
`IndexError: index 5280 is out of bounds for axis 0 with size 882`

trimesh/visual/texture.py
wkentaro: clip makes

- 1.1 -> 1
- -0.3 -> 0

right? This is not what we expect.

trimesh/io/wavefront.py
wkentaro: 8400f85 fixes this.

trimesh/io/wavefront.py
wkentaro: I'm not sure what is the best, but I updated in c71b473. Could you check that?

trimesh/visual/texture.py
mikedh: I guess it depends on what you expect haha, that behavior seems a bizzare to me but if it's what people do then we should probably replicate it, through something like:
```
In [12]: uv[uv < 0.0] += 1.0
In [13]: uv %= 1.0
```

trimesh/io/wavefront.py
mikedh: Yeah that looks great, thanks!

trimesh/visual/texture.py
wkentaro: Actually, uv can be under -1 like -21.3 for shapenet models.
Meshlab still can load that model, so it is still supposed to be an acceptable format.

mikedh: Ah good point. Looks like normalizing is not a good idea: https://stackoverflow.com/questions/49923442/threejs-normalize-uv-coordinates-to-0-and-1-issue

mikedh: Actually don't we just want to mod-1 the values right before evaluating? I.E: 
```
x = ((uv[:, 0] % 1.0) * (image.width - 1)).round().astype(np.int64)
y = (((1 - uv[:, 1]) % 1.0) * (image.height - 1)).round().astype(np.int64)
```

mikedh:
Awesome thanks for the changes! I'm going to merge and fix  tests.

PR #279, opened 2018-12-21
Filter invalid kwargs for Trimesh and ColorVisuals

### LINKED ISSUE #275: unused kwargs in trimesh.base.Trimesh
Why trimesh does not check unused kwargs passed to trimesh.Trimesh?
I had difficulty while debugging because of typo of kwargs: (x) preprocess (o) process.

wkentaro:
Close #275

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -15,7 +15,7 @@ from . import units
 from . import poses
 from . import graph
 from . import voxel
-from . import visual
+from . import visual as visual_module
 from . import sample
 from . import repair
 from . import convex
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! Maybe we should just remove `kwargs` entirely from the constructor and add a `visual=None`, rather than filtering if we're going to raise anyway if anything is passed.

wkentaro:
Updated.

mikedh:
LGTM- thanks for the PR!

wkentaro:
Ok, I see.

PR #291, opened 2019-01-12
Fixes two small bugs for glTF PBR material loading

mmatl:
Love the new texture edits, Mike! Great work!


mikedh:
Thanks! Good catch!

PR #292, opened 2019-01-13
Use abc.ABCMeta for Geometry class

wkentaro:
To avoid expected but not implemented methods like apply_translation of
PointCloud.

### DIFF
--- a/trimesh/parent.py
+++ b/trimesh/parent.py
@@ -4,12 +4,42 @@ parent.py
 
 The base class for Trimesh, PointCloud, and Scene objects
 """
+import abc
+
 import numpy as np
 
 from . import caching
 
 
-class Geometry(object):
+class Geometry(metaclass=abc.ABCMeta):
+    @abc.abstractproperty
+    def bounds(self):
+        pass
+
+    @abc.abstractproperty
+    def extents(self):
+        pass
+
+    @abc.abstractmethod
+    def apply_transform(self):
+        pass
+
+    @abc.abstractmethod
+    def apply_translation(self):
+        pass
+
+    @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


mikedh:
Great idea!

I might add a comment in the top of the `Geometry` object explaining why we're using a metaclass:

```
The `Geometry` object is the parent object of geometry classes, including:
Trimesh, PointCloud, and Scene objects.

By decorating a method with `abc.abstractmethod` it just means the objects that 
inherit from `Geometry` MUST implement those methods. 
```

One other thought: `apply_translation`, `apply_scale` can actually be implemented in the `Geometry` object as they just generate a transform and are the same everywhere, they just call `self.apply_transform` that needs the per- object implementation (it could be the `abc.abstractmethod`)


It looks like it needs some tweaks to pass in Python 2, there's a discussion [here](https://stackoverflow.com/questions/35673474/using-abc-abcmeta-in-a-way-it-is-compatible-both-with-python-2-7-and-python-3-5), it looks like the cleanest solution is (IMHO):

```
import sys
import abc

if sys.version_info >= (3, 4):
    ABC = abc.ABC
else:
    ABC = abc.ABCMeta('ABC', (), {})

class SomeAbstractClass(ABC):
    @abc.abstractmethod
    def do_something(self):
        pass
```

mmatl:
@mikedh Another option is using `six`, although it adds another (very light) dependency for Python 2/3 interop. Then, it would just look like:

```
import abc
import six

@six.add_metaclass(abc.ABCMeta)
class SomeAbstractClass(ABC):
    @abc.abstractmethod
    def do_something(self):
        pass
```

I think that should work.

wkentaro:
I updated.

PR #302, opened 2019-01-21
Use raw string for Windows paths

hroncok:
This avoids:

    DeprecationWarning: invalid escape sequence \P
        _search_path.append('C:\Program Files')


mikedh:
Ah good catch, thanks! I'm surprised this worked at all before 😁

hroncok:
it works because `\P` is not a valid escape sequence. that works but might not in the future, hence the DeprecationWarning I guess.

PR #310, opened 2019-02-03
closest_point() fix for (obtuse) triangles

dbukenberger:
Hai again,

I have noticed an issue with the signed_distance() and on_surface() methods and were able to trace it back to an incorrect computation of the closest_point() method of the triangle class.
The selection for case_vertex and case_edge based on the number of positive barycentric coordinates is only correct for certain points but not for all outside of a triangle.
For cases which belong in case_vertex but end up in case_edge, the clipping at the end (to the edge's extend) will fix this and return the correct vertex position.
However, for obtuse triangles there are scenarios that belong in case_edge but end up in case_vertex and will return a wrong point.
I extended the triangle class with a minimally invasive fix on the barycentric-coordinate's 'positive' mask to also handle these cases correctly.
Furthermore, I have added a simple test for one point with ground-truth result and a circle of points with a naive point-to-edge distance comparison to check the results.

The fix feels kind of hacky, but as far as I have checked, it seems to produce correct results now.
Do you think its worth to refactor the method? Let me know, I think I could come up with something.

Cheers,
Dennis

Also - here is a nice wysiwyg-tool which I used to play around with barycentric coordinates and to figure out when and why the current method failed:
https://www.geogebra.org/m/ZuvmPjmy

### DIFF
--- a/tests/test_triangles.py
+++ b/tests/test_triangles.py
@@ -34,19 +34,26 @@ class TrianglesTest(g.unittest.TestCase):
         self.assertTrue((comparison < 1e-8).all())
         g.log.info('finished closest check on %d triangles', len(closest))
 
-    def test_explicit_edge_case(self):
-        # taken from a real scenario where I encountered the issue
-        ABC = g.np.float32([[0.51729788, 0.9469626 , 0.76545976],
-                            [0.28239584, 0.22104536, 0.68622209],
-                            [0.1671392 , 0.39244247, 0.61805235]])
-        D = g.np.float32([0.02292462, 0.75669649, 0.20354384])
-        # point D projected on ABC plane is not in ABC triangle
+    def test_closest_obtuse(self):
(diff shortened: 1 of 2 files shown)


mikedh:
Once I added in tests against random triangles I was still seeing failures (compared to Shapely as ground truth), so I gave up and changed the function to the method in the [Real Time Collision Detection](http://www.r-5.org/files/books/computers/algo-list/realtime-3d/Christer_Ericson-Real-Time_Collision_Detection-EN.pdf) book on page 141. It's now passing on the obtuse case and against random data. Apparently there are 6 possible regions :grimacing:. The RTCD method is roughly 30% faster than the old method, which is nice. 

Let me know if you see any results that don't look right, I tested it a bunch but there is a lot of new and  hard to follow mathy code. Thanks for the PR/test/fix!

Issue #193, opened 2018-07-18
Set light in a scene


clemense:
Hi,
Is it possible to set the light in a scene?
(I would like to visualize colored point clouds whose colors are not affected by the perspective of the camera; right now colors change drastically when rotating the point cloud in the viewer)
Thanks!

mikedh:
Hey, right now the lighting is all [hardcoded in the viewer's gl initialization](https://github.com/mikedh/trimesh/blob/master/trimesh/scene/viewer.py#L129-L138), as the scene viewer was originally written more for "are there any triangles in this mesh" than general scene graph use. 

That being said, it would be nice if there were light and camera objects (maybe `trimesh.visual.Light`/ `trimesh.scene.lighting.Light` or something). I'm open to PR's, and may look in to doing that at some point.

ifermion:
Hi @clemense ! 
I've faced the same problem and was trying to find a work-around. I ended up modifying several lighting parameters in /trimesh/viewer/windowed.py and restoring them once visualisation is done. I'm not sure if this is of the sort that you've been after.

The idea was to remove the diffused and specular reflection effects from the light source initiated by the function init_gl (https://github.com/mikedh/trimesh/blob/master/trimesh/viewer/windowed.py) and increase the ambient effect. I admit that this is a crude way, but I managed to get done what I wanted. 

To automate the process, I created two functions that I share here. The first is to initiate and modify the init_gl to be called before the initiation and the second is to be called at the end of the visualisation, in order to restore the original parameters of the init_gl. Please let me know if you already managed to solve this issue. 

``` 
def pyglet_overwriter():

    # This function is used to overwrite some pyglet window initialising parameters
    # in order to remove the diffused and specular reflection effects initialised by default, in order
    # for the visualisation to be uniform and not biased on the vector of the light source.
    # I admit this is a crude way, but I could not find a more direct way of solving this problem as I could not find
    # a way to call these relevant initialising parameters from within the trimesh library.
    #
    # The function looks for the windowed.py file in the installation path and makes a copy of it as a backup.
    # Then it modifies the file and this modified file will be used for initialising the pyglet window.

    import sys
    import os

    paths = sys.path
    for path in paths:
        if '/lib/python3.' in path:
            #print(path)
            break


    dir = '/site-packages/trimesh/viewer/'
    full_path=path+dir

    wdir = os.getcwd()
    os.chdir(full_path)


[...]

PR #317, opened 2019-02-10
Add collada I/O

mmatl:
This does better COLLADA I/O w/ support for materials and such. Still need to do proper texture exporting, but can add that later. One quirk of this is that `TextureVisuals` should have its `uv` attribute be optional, as sometimes it's useful to specify a material without any texture (i.e. can still specify emission, metalness, roughness, etc...).

This relies on `pycollada`, which, as far as I can tell, only depends on `lxml` and optionally `pillow`.


mikedh:
Nice! `pycollada` looks a heck of a lot better than `pyassimp`, this looks good to me. I did notice that it looks like the latest version on pypi renamed `.img` to `.image`, which once I changed this worked great (I couldn't figure out how to push to the PR but feel free to take anything from https://github.com/mikedh/trimesh/commit/2f9ddd4edf3cc29f4819d8c36b2ef0a656fabbd9)

On some slightly janky test files I had lying around, I saw a lot of `DaeSomethingException` (I forget the exact one) when trying to load images, where it couldn't locate the files due to a resolver issue. I think the only thing for those is to wrap them. 

Also yeah I agree with making UV's optional.

mmatl:
Hi Mike! New code pushed just now:

- Integrated the changes you asked for in that link.
- Added code to handle resolving textures when loading from colladas.
- Also (EXCITING) -- figured out a relatively clean way to prevent textures from looking bad at the seams of objects. Basically, when merging vertices, you only merge vertices whose positions AND uv coordinates are identical. Then, you forcibly re-compute vertex normals and merge to the normal after grouping vertices by position. I included this change as well -- could probably go in its own PR, but can move it later if you want.

Still need to handle texture exports, although I do see a reasonable path forward for that I think (pass a resovler to the exporter, try to get a filename from PIL images, and export to a reasonable location).

Best,
Matt

mikedh:
Awesome! I’ve been working on a branch (`feature/scene`) to try to make the Scene objects slightly nicer.  I think it’d be cool if you could load a scene containing lights and cameras via collada/assimp/gltf/etc and send it straight to `pyrender` or any other endpoint (notebooks/yafaray/etc) and have it do something reasonable. 

My short list of things to make this happen (let me know if I missed anything):
- Add light objects to trimesh that can be used by any endpoints (also resolves #193). Your light model in pyrender looked great so I yoinked it, though I removed some of the rendering specific items (shadow), hopefully this is OK with you. 
- Every scene has a single camera object, `set_camera` creates it if it isn’t passed.
- Every scene has a list of lights and an “auto light” function if nothing is set that does something reasonable.

I’ll merge into the feature branch and try to get this released sometime this week. Thanks for the PR!

mmatl:
@mikedh Sounds awesome! And I'm totally up for that, since COLLADA and GLTF natually support scene graphs + cameras + lights and such. Happy to help with getting that feature branch done, let me know if there's anything I can do to help.

PR #321, opened 2019-02-15
Smoothing Algorithms

bbarroqueiro:
Hello, i added smoothing.py which as four def's. The Laplacian calculation which is the most expensive step. Then three filters: laplacian, humphrey, taubin, which can be applied multiple times in different orders in an i
inexpensive way...

### DIFF
--- a/trimesh/smoothing.py
+++ b/trimesh/smoothing.py
@@ -1,4 +1,4 @@
-import trimesh, itertools
+import itertools
 from scipy.sparse import coo_matrix
 import numpy as np
 
@@ -41,7 +41,6 @@ def filter_humphrey(mesh,alpha,beta,n_filter,laplacian_operator):
       
       # Number of passes
       vert_o=mesh.vertices.copy()
-      vert_b=mesh.vertices.copy()
       for _ in range(n_filter):
             vert_q=mesh.vertices.copy()
             mesh.vertices=coo_matrix.dot(laplacian_operator, mesh.vertices)
--- a/trimesh/smoothing.py
+++ b/trimesh/smoothing.py
@@ -91,9 +91,6 @@ def laplacian_calculation(mesh,weight_type):
             # Umbrella 
             elif  weight_type == 2:
                   umb_weights=1.0/np.sum((vert[i]-vert[vert_neigh[i]])**2, axis=1)**0.5; umb_weights/=np.sum(umb_weights)
-            # Vertex Normals 
-            elif  weight_type == 3:
-                  umb_weights=1.0/np.sum((vert_ori[i]-vert_ori[vert_neigh[i]])**2, axis=1)**0.5; umb_weights = 0.8*umb_weights/np.sum(umb_weights) + 0.2*np.ones(n_neigh)/n_neigh
             else:
                   raise IndexError("Select 1 for Equal Weights\n             2 for Umbrella weights")
             
--- a/trimesh/smoothing.py
+++ b/trimesh/smoothing.py
@@ -1,0 +1,109 @@
+import trimesh, itertools
+from scipy.sparse import coo_matrix
+import numpy as np
+
+def filter_laplacian(mesh,lamb,n_filter,laplacian_operator):
+      """
+      Mesh Smoothing using Laplacian algorithm with different weighting schemes
+
+      Articles:
+      J. Vollmer, R. Mencl, and H. Müller; Improved Laplacian Smoothing of Noisy Surface Meshes
+      
+      Input:
+            mesh to be modofied
+            lamb = (float)  the difusion speed constant: 0 no difusion,  1 full difusion (e.g. 0.5)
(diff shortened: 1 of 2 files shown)


trimesh/smoothing.py
```diff
@@ -0,0 +1,106 @@
...
+                  mesh.vertices = mesh.vertices - nu*( coo_matrix.dot(laplacian_operator, mesh.vertices) - mesh.vertices )              
+
+def laplacian_calculation(mesh,weight_type):
+      """
+      Input: Expects a trimesh mesh
+             weight_type = 1 - Equal Weights
+                         = 2 - Umbrella Weights
+      Output: Laplacian operator (scipy sparse matrix)
+      """
+      vert_neigh=mesh.vertex_neighbors; vert=mesh.vertices; vert_ori=mesh.vertex_normals.copy(); n_vert=len(vert)
```
Marviel: Could we split these out into separate lines without semicolons?

mikedh:
Looks great, huge improvement thanks! I'll probably change up the formatting when I merge (remove semicolons like @Marviel suggested, remove trimesh import), though one thing before I merge: could you add a unit test file, `tests/test_smoothing.py`? I'd copy an existing one as a template and then add a test something like:

```
m = g.trimesh.creation.icosahedron()
s = g.trimesh.smoothing.laplacian(m)

assert m.is_volume
assert s.is_volume

# I don't know what these ratios are supposed to be but this should be repeatable
assert g.np.isclose(m.area / s.area, AREA_RATIO)
assert g.np.isclose(m.volume / s.volume, VOLUME_RATIO)
```

trimesh/smoothing.py
bbarroqueiro: Done... Thanks for the suggestion...

bbarroqueiro:
I can try... But it is not easy to state a test... The Laplacian filter is known to reduce volume and Humphry and Taubin tries to address the issue... 

There always be some changes in volume and specially in area and it depends how severe the user sets the diffusion. If the diffusion is to severe large changes in shape can occur and the can be working just fine... I can think about it, see if i can come up with some test...

mikedh:
Hey, yeah no worries, the test mainly needs to be repeatable, it's not necessarily "exactly assess whether the solution is correct or not." The value is making sure that the smoothing routines are run in CI so we can make sure that given the same parameters changes elsewhere don't break the functionality. 

Thanks!

bbarroqueiro:
Sorry, but i'm new to github an programming.... 

How can i test my "test_smoothing.py" code? Before committing it?

mikedh:
No worries, you may be running in to an import paths issue: you can either put the directory with the trimesh repo into the `PYTHONPATH` env variable, here's mine:
```
mikedh@cue:tests$ echo $PYTHONPATH
/media/psf/Dropbox/robotics/pycollada:/media/psf/Dropbox/robotics/pypocketing:/media/psf/Dropbox/robotics/pyactp:/home/mikedh/trimesh:
```

You can also do an editable pip install from inside the trimesh directory, which will do basically the same thing:
`pip install -e .`


I'd put the following in `tests/test_smoothing.py` and then run `python test_smoothing.py`. 
```
try:
    from . import generic as g
except BaseException:
    import generic as g


class SmoothTest(g.unittest.TestCase):
    def test_smooth(self):
        """
        Load a collada scene with pycollada.
        """

        m = g.trimesh.creation.icosahedron()
	s = g.trimesh.smoothing.laplacian(m)

	assert m.is_volume
	assert s.is_volume

	# I don't know what these ratios are supposed to be but this should be repeatable
	assert g.np.isclose(m.area / s.area, AREA_RATIO)
	assert g.np.isclose(m.volume / s.volume, VOLUME_RATIO)

if __name__ == '__main__':
    g.trimesh.util.attach_to_log()
    g.unittest.main()
```


Or if you just have a code block that works as a script (with some `assert`'s) just send me a copy and I can put it in the right place.

trimesh/smoothing.py
Marviel: No prob man, thanks! I didn't say this before, but nice work! This will be very useful!!!

bbarroqueiro:
Hello, i added a test file hopefully it is properly done... But, i will have read a bit about how git works so far a just my script in vs code with anaconda and that's it...

About the algorithms, these are classic well know algorithms much more advanced algorithms are available. 

Now i'm working in a new weighting function... At a given vertices and its neighbour (that forms an edge i need to know the area of the two triangles that forms the edge...
My current solution quite slow, i'm checking all nearby_faces of all vertices and then i intersect the nearby faces of vertice and nearby_face of its neighbour.  Thus, i get the wanted two elements and then their area... Any better ideas?

Regards
Bruno

mikedh:
Looks terrific, thanks! I'll merge to a branch for testing and profiling and then get it into master ASAP. 

This is a huge improvement, thanks again for taking the time to PR!

Issue #301, opened 2019-01-21
Potential Integration with Pyrender PBR


mmatl:
Hi @mikedh!

I've created a new repo for my changes to `meshrender`. For now, I called it [pyrender](https://github.com/mmatl/pyrender) -- open to suggestions for a more catchy name :) 

Hoping to get a bit of feedback on what works and what you'd like to see change for it to integrate nicely with trimesh. Here are some points that I know will need work:

- For now, it seems to work well on Ubuntu! Performance can take a hit when shadowmapping for many lights -- I can fix this with a deferred rendering pipeline, but that would have to come at a later date when I have some time to write one.
- The repo uses GLSL 3+ shaders, so you need to be able to get a core profile in Pyglet to use the viewer. This works without difficulty on Linux, but Pyglet's code for MacOS is broken. I've submitted PRs to try to fix this, but they're quite slow to respond and release new versions. I forked Pyglet and can point the setup.py to use my fork, but it's not the prettiest solution.
- Shadowmapping still doesn't work nicely on MacOS (something to do with retina displays, will need to debug).
- I still need to test things on Windows.

Best,
Matt

mikedh:
Hey Matt, sounds great! I like the new name haha, and I see you registered it on pypi 👍.  How does this sound:
## Integration Plan
- `trimesh` will add `pyrender` to `pip install trimesh[all]` deps
- `trimesh.Scene.show` will add another viewer option, `pyrender`. Defaults to it if the scene only contains `Trimesh` objects (unless you want to support `PointCloud` and `Path` objects)
- `trimesh.save_image` will add a viewer option, `pyrender`


### Platform Notes
#### Windows
- all the deps install cleanly, except for `pyopengl_accelerate` which wants to compile stuff (and that's kind of a mud wrestling match in windows). 
  - 🙏 they have windows wheels in [pre- release](https://pypi.org/project/PyOpenGL-accelerate/3.1.3b1/#files) which I'll try later, and hopefully get released soon.
- I tried it on windows without `pyopengl_accelerate`, and wasn't able to get an OpenGL 3+ context due to something WGL_ARB. Possibly solved by the missing dep.

#### Mac
- re: pyglet, have you tried PR'ing into their main [bitbucket repo](https://bitbucket.org/pyglet/pyglet/overview)? I submitted a minor fix a few years ago and they were pretty responsive.

mikedh:
[...]

PR #281, opened 2018-12-21
[WIP] Texture Support

mikedh:
Opening a PR to track progress; this is slowly getting there. Fixes #218 

- [x] `TextureVisual` object
- [ ] Finalize `TextureVisuals` data structure (multiple materials? manage UV for multiple materials?)
- [x] `Material` Objects: PBR, Basic
- [ ] Finalize `Material` data structure
- [X] Assets resolver for file system, ZIP, and remote
- [x] OBJ load texture
- [x] UV/Texture to color
- [x] GLTF load texture
- [x] GLTF exports don't crash on texture
- [x] PLY load texture
- [x] Figure out why PLY textures are mapping wrong
- [x] Pyglet viewer show texture

### Post-PR Scope
- GLTF export texture
- PLY export texture
- OBJ export texture
- shaders for PBR in pyglet viewer `trimesh.viewer.windowed`

Other major changes: `trimesh.io` and `trimesh.path.io` renamed to `trimesh.exchange` and `trimesh.path.exchange`, as `io` is a builtin library that we were previously stomping on and relying on import precedence. `exchange` isn't so great but is better than anything else I could think of, and is used elsewhere to mean this (Hoops Exchange is basically a bunch of importers/exporters).

Issue #237, opened 2018-10-31
scene.set_camera() ignored in notebook


foobarbecue:
`.set_camera` successfully sets the position of the camera when displayed using pyglet, but has no impact on the camera position in notebook.

As a minimal case:

```
import trimesh
scene = trimesh.Scene(trimesh.primitives.Box())
scene.set_camera(angles=(0.5, 0, 0), distance=5, center=(0,0,0))
scene.show()
```

appears the same in jupyter notebook with and without the third line, but different if run from a command line.

mikedh:
Ah yeah, the three.js viewer uses an [auto- centering function](https://github.com/mikedh/trimesh/blob/master/trimesh/resources/javascript/load_base64.js#L5-L35) rather than the camera transform. Also GLTF doesn't currently export the camera transform at all (IIRC). This could be a nice improvement!

mikedh:
Should be fixed by https://github.com/mikedh/trimesh/commit/ef477f06fa88a5f6e5e923f9c1614bc587efaaab.

PR #322, opened 2019-02-15
Scenery: Lights and Cameras

### LINKED ISSUE #193: Set light in a scene

mikedh:
Previously, our support for cameras was limited and lights nonexistent. This is a followup to #281 which added texture support. The thought here is that it would be cool to load a scene file (DAE/GLTF/assimp) including lights and cameras, and then pass it to a render or export endpoint (pyrender, GLTF, notebooks, pyglet, yafaray, etc) and have light/camera information preserved. 

The basic change is that now scenes always have a SINGLE camera object `scene.camera` which can be specified manually or is autogenerated if not. Scenes have a LIST of light objects `scene.lights` which can be set, or is automatically generated to look "OK" otherwise.  

`scene.graph` is still pretty rough and probably the worst part about scene objects. I like having a single graph object based on networkx, but this could do with one (or several) API/cleanup passes that are out of scope of this PR. 

This is working towards #301, and includes #317.  I'm very open to suggestions haha- any puts @mmatl @wkentaro?

Changes:
- scene objects now have a single `scene.camera` object, which can be assigned or if unset will be autogenerated when requested
- scene objects now have a list of `scene.lights`, which `viewer.windowed` uses (fixes #193)
- `trimesh.scene.lighting` was pulled pretty much directly from `pyrender`. The only major change is renaming `range` to `radius` to avoid conflicting with the builtin function.
- `load_pyssimp` now returns scene objects. `pyassimp` is currently tough to install, I made some speedups and am working to get wheels including the correct version of the shared library building here https://github.com/mikedh/pyassimp 
- DAE now returns scenes. Tests are going to keep failing until https://github.com/pycollada/pycollada/pull/86 is merged or I remove `barrett-wam.zae` from models.


mmatl:
Looking amazing! Will provide input when I get a change.

mmatl:
@mikedh Anything in particular I can help with to get this finished up?

mikedh:
Hey Matt, this should be mostly ready to go. Just checking if the "single camera, multiple lights" model works for you before merging. 

There's still some miscellaneous stuff that can probably be addressed in master or later. Some quick notes:

- colors in the rest of trimesh are RGBA `uint8`. I kind of regret not making these 0.0-1.0 floats, but I changed the `scene/lighting` colors to match the rest of trimesh for consistencies sake. OpenGL takes RGBA colors for lights... what does alpha even mean for lights :grin: 
- `scene/cameras/look_at` function seems right-ish, but I'd like to check it more carefully and unit test. Meshes appear to be more than filling up the window, rather than exactly filling up one axis of the window as intended. 
- `exchange/dae` has TODO placeholders for converting lights + cameras to the new data types

mmatl:
Hi Mike,

Great work! Yeah, the single-camera, multiple-lights model works perfectly for me. I can't really think of cases where someone _really_ actually wants multiple cameras, and if they do, they can always just use pyrender directly to add them.

With regards to the other stuff:

- Yeah, `uint8` is more efficient, but all the OpenGL stuff uses `float32` for all colors :/ And yeah, I was SO confused about why lights have an alpha -- as far as I can tell, you _never_ use light alphas in shading...
- I'll look over `scene/cameras/look_at`.
- I can add those placeholders if you want, either here or in master.

mikedh:
Sounds good! Yeah I'm open to converting all colors to float RGBA someday, though I think it would be important to do it everywhere rather than mixing color data types. 

If you wouldn't mind taking a look at (hah) `look_at` and the DAE light/camera conversions that would be awesome thanks!

Brief punch list for visual stuff, in no order and with no particular timeline:
- DAE light+camera conversion
- `look_at` test and fixes
- export `camera.transform` to GLTF (#237) and update notebook viewer to use it
- export lights to GLTF
- export texture to GLTF

mmatl:
It seems to me like you should transform the points _into_ the camera's space (i.e. the inverse of the initial camera rotation), do your calculations there, and then transform the camera center back to the world space?

Here's my re-implementation:

```python
def look_at(points, fov, rotation=None):
    """
    Generate transform for a camera to keep a list
    of points in the camera's field of view.
    Parameters
    -------------
    points : (n, 3) float
      Points in space
    fov : (2,) float
      Field of view, in DEGREES
    rotation : None, or (4, 4) float
      Rotation matrix for initial rotation

    Returns
    --------------
    transform : (4, 4) float
      Transformation matrix with points in view
    """
    if rotation is None:
        rotation = np.eye(4)
    else:
        rotation = np.asanyarray(rotation, dtype=np.float64)
    points = np.asanyarray(points, dtype=np.float64)

    # Transform points to camera frame (just use the rotation part)
    rinv = rotation[:3, :3].T
    points_c = rinv.dot(points.T).T

    # Find the center of the points' AABB in camera frame
    center_c = points_c.min(axis=0) + 0.5 * points_c.ptp(axis=0)

    # Re-center the points around the camera-frame origin
    points_c -= center_c

    # Find the minimum distance for the camera from the origin
    # so that all points fit in the view frustrum
    tfov = np.tan(np.radians(fov) / 2.0)
    dist = np.max(np.abs(points_c[:, :2]) / tfov + points_c[:, 2][:, np.newaxis])

    # Set the pose translation by sending the 
    center_w = rotation[:3, :3].dot(center_c)
    cam_pose = rotation.copy()
    cam_pose[:3, 3] = center_w + dist * cam_pose[:3, 2]

    return cam_pose
```
The behavior is such that at least one edge of the screen is _juuuuust_ touching the object. The object won't necessarily touch both sides, as we're constrained to search along the view vector while keeping the object's AABB center in the middle of the scene. Getting the object to fill a full axis is quite a bit harder, I think.

Examples here:
fuze
ico
drill

PR #324, opened 2019-02-20
Fix collada I/O bugs

mmatl:
Hey man, sorry I got this to you late! There was a bug in your revision of the COLLADA exporter that crops up for geometry nodes with more than one primitive. Basically, each node could have several meshes assigned to it as primitives, and using the geometry's ID as a dict key was preventing these meshes from being added.

Also, added more types to the "list" check to make it easier to export many meshes into one DAE.

### DIFF
--- a/trimesh/exchange/dae.py
+++ b/trimesh/exchange/dae.py
@@ -85,7 +85,7 @@ def export_collada(mesh, **kwargs):
     export: str, string of COLLADA format output
     """
     meshes = mesh
-    if not isinstance(mesh, list) and not isinstance(mesh, tuple):
+    if not isinstance(mesh, (list, tuple, set, np.ndarray)):
         meshes = [mesh]
 
     c = collada.Collada()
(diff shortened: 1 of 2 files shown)


mmatl:
Example mesh for which this a problem: 
[shoulder.dae.zip](https://github.com/mikedh/trimesh/files/2883115/shoulder.dae.zip)

mikedh:
LGTM, thanks! I didn't know `isinstance` could check against multiple classes, that's handy!

Heh is that a model of UR10 shoulder I see? Could we include it in tests or is the model non-BSD/MIT?

mmatl:
Yeah, it's super handy! Just found out last week about it haha!

And yep, it's a UR shoulder ;) I pulled the model from here: https://github.com/ros-industrial/universal_robot. It doesn't say on Github explicitly, but on the ROS wiki it says that it's under BSD, so you can include it in tests.

PR #334, opened 2019-03-02
Added object_name to mesh metadata for wavefront

ThoBaro:
Hi,
I'm interested in the object names within my .obj files so i added the read out for .obj files.
Best,
ThoBaro


mikedh:
Looks great thanks! Would you mind adding a quick test which loads a model with this name defined, and runs something like:
 
` assert 'groupName' in mesh.metadata['object_name']`

ThoBaro:
Will do during the week!

PR #342, opened 2019-03-08
Make color conversion more permissive.

mmatl:
Extremely small PR here. I've found that sometimes you'll have DAE files that have colors written in the format [255.0 237.0 233.0 255.0], for example (floating point, but really should be int). Making the color converter a bit more permissive makes it possible to cleanly load these types of meshes. I've attached an example here.

[base.dae.zip](https://github.com/mikedh/trimesh/files/2944039/base.dae.zip)


mikedh:
LGTM, thanks!

PR #345, opened 2019-03-16
Format the %s string

hroncok:
This is not a log call, but an exception. The result was:

    ValueError: ('%s exporter not available!', 'dae')


mikedh:
LGTM, thanks!

PR #352, opened 2019-03-22
Fix small issues with DAE and GLB mats

mmatl:
Basically, this allows you to actually use DAE and GLB/GLTF materials when (A) vertex colors are present, or (B) when UV coordinates are not present.


mikedh:
LGTM, thanks for the PR!

PR #353, opened 2019-03-22
Update wavefront.py

tranduytrung:
fix "PIL has no attribute Image"

PIL itself does not have anything, need explicit import.


mikedh:
LGTM, thanks! I think different versions of PIL/pillow handle `__init__` differently, but your solution should work on all of them.

PR #354, opened 2019-03-22
Estimate memory usage before actually computing the radius using cdist

lejafar:
Hi Mike!

I came across a file that consumed 23 GB (94698 voronoi vertices and 31361 points), while computing the radius of the bounding sphere, no `MemoryError` was thrown but my machine became unusable and it took quite a while because it was using swap, so I thought that a memory estimate might by a good solution here?

Not sure if 1 GB is an appropriate limit though ...

### DIFF
--- a/trimesh/nsphere.py
+++ b/trimesh/nsphere.py
@@ -74,6 +74,10 @@ def minimum_nsphere(obj):
     try:
         # cdist is massivly faster than looping or tiling methods
         # although it does create a very large intermediate array
+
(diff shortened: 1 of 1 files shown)


lejafar:
Another possibility would be to query the available memory? https://stackoverflow.com/a/11615673/8890398

mikedh:
Makes sense, thanks! Yeah PSutil might be a good option too for setting that parameter.

PR #362, opened 2019-03-30
Support changing scene geometries in SceneViewer.callback

wkentaro:
Below script doesn't work without this change.
I guess there were some regressions because it had worked before.

```python

import numpy as np
import pyglet
import trimesh
import trimesh.viewer


def callback(scene):
    if scene.pause:
        return

    translation = np.random.uniform(0, 1, (3,))
    axis = trimesh.creation.axis()
    axis.apply_translation(translation)
    scene.add_geometry(axis)
    scene.set_camera()


def main():
    scene = trimesh.Scene()
    scene.add_geometry(trimesh.creation.axis())
    window = trimesh.viewer.SceneViewer(
        scene=scene,
        callback=callback,
        callback_period=0.1,
        start_loop=False,
    )
    scene.pause = True
    scene.window = window

    @window.event
    def on_key_press(symbol, modifiers):
        if symbol == pyglet.window.key.Q:
            window.close()
        if symbol == pyglet.window.key.S:
            scene.pause = not scene.pause

    pyglet.app.run()


if __name__ == '__main__':
    main()
```

### DIFF
--- a/trimesh/viewer/windowed.py
+++ b/trimesh/viewer/windowed.py
@@ -77,6 +77,7 @@ class SceneViewer(pyglet.window.Window):
         self.scene._redraw = self._redraw
         self.reset_view(flags=flags)
         self.batch = pyglet.graphics.Batch()
+        self._smooth = smooth
 
         # store kwargs
         self.kwargs = kwargs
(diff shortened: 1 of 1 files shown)


wkentaro:
It looks like this:

Screencast 2019-03-30 21:23:37

mikedh:
Oh, I must have accidentally removed that in a refactor. It's too bad CI would be such a pain to test this. Thanks for the PR!

wkentaro:
Thanks!

PR #366, opened 2019-04-08
Add color argument to trimesh.creation.icosphere for convenience

wkentaro:
It seems hard to pass `**kwargs` to `Trimesh()` in `icosphere` like `creation.box`.


mikedh:
Looks great, thanks!

Issue #206, opened 2018-08-18
Export mesh in .msh fileformat


Walshaikhli:
First, thank you for maintaining Trimesh, I have used it for a day so far and it's an amazing tool! I wish i knew it existed a year ago! 

I was wondering if I can export .msh files directly from Trimesh, this is the GMSH program format, and it's used in the state of the art EM solvers such as SCUFF-EM. I can use GMSH to convert to .msh but I'm hoping to use Trimesh for it. 

Here is what I'm doing:
`trimesh.io.export.export_mesh(mesh,'outputFile', file_type="msh")`

Here is the error I get:

> ValueError                                Traceback (most recent call last)
> <ipython-input-18-3d4dc592d077> in <module>()
> ----> 1 trimesh.io.export.export_mesh(mesh,'outputFile', file_type="msh")
> 
> ~\Anaconda3\lib\site-packages\trimesh\io\export.py in export_mesh(mesh, file_obj, file_type, **kwargs)
>      39 
>      40     if not (file_type in _mesh_exporters):
> ---> 41         raise ValueError('%s exporter not available!', file_type)
>      42 
>      43     log.debug('Exporting %d faces as %s', len(mesh.faces), file_type.upper())
> 
> ValueError: ('%s exporter not available!', 'msh')

mikedh:
Hey, glad you're liking it!

Currently `msh` doesn't have any importers/exporters in `trimesh`. On the upside, the format is [very well specified](http://gmsh.info/doc/texinfo/gmsh.html#MSH-binary-file-format) and looks pretty easy to implement. If I get a chance I'll take a look at doing it, but if you're feeling energetic and want to PR here's the punch list for supporting it:

- add a <200kB test model of ASCII MSH in /models/
- add a <200kB test model of Binary MSH in /models
- add Binary/ASCII importer and Binary exporter to /trimesh/io/msh.py (either `ply.py` or `stl.py` might be good templates) that uses `np.fromstring` / `np.frombuffer` for key calls to avoid being mega slow. It might have to reject input meshes with elements other than triangles or quads. 
- Add unit tests to `test_export.py`, and also possibly check the validity of our exports using a subprocess call to `gmsh`

mikedh:
Added to enhancements list. Thanks for the report!

selvakarna:
Hi  mikedh ,
      I have Error while Create STL from X Y Z co-ordinates using Trimesh.

[...]

PR #369, opened 2019-04-12
Volume Mesher

bbarroqueiro:
Hi Michael 

Its a small code that generates volume mesh from a surface mesh...

This was a issue i had and looking ate forums it quite frequent.... 

This code needs:
pip install gmsh-sdk
or since this sdk is so small it can included inside module!

Let me know what you think...

I hope his helps...

regards


mikedh:
Nice, that looks like it could be very useful! Thanks for the PR! A few suggestions:

- I'd rename the submodule from `trimesh.mesh3D` to  `trimesh.volumetric` or `trimesh.fea`, or something else
- `trimesh.mesh3D.mesh3D` -> `trimesh.fea.to_bdf`
- One goal of trimesh is to keep the mandatory dependencies to a minimum, so I'd like `gmsh` to be a "soft" dependency, i.e:
```
try:
    import gmsh
except ImportError:
    pass
```
- rather than using `id_name_in` and `id_name_out` I'd prefer to use `tempfile.NamedTemporaryFile` for both input and output. An example of doing this is [here](https://github.com/mikedh/trimesh/blob/f42f7a6c4838e9d18815028718acff070c275866/trimesh/exchange/ply.py#L616-L631)
- I'd rather the function returns the data from the `bdf`, and then the user can decide whether they want to write it to a file or not.

bbarroqueiro:
Hi,
I tried to follow your suggestions...
But, in the last comment i think you should keep it as is... But i leave the decision to you... I don't mind at all...

Just remember that a 200K elements surface mesh can easily generate a few millions elements of volumetric mesh (it an be memory pit even in Fortran codes and python can be memory hungry). I would leave it to file. 

If you return it and save it memory to do what? It should be export to be viewed, analysed and maybe optimized in order codes, right?

mikedh:
No worries, yeah I totally get not wanting to load a 20gb tet mesh into memory. Maybe a compromise would be:

```
generate(file_name=None, ... )
"""
Parameters
---------------------
file_name : None or str
  If specified, will save the output to a file at the specified location 
  If None, the data will be returned as bytes
"""

bbarroqueiro:
Hi, something like this code could be added:

    if id_name_out != None:
        gmsh.write(id_name_out)
    else:
        #Exports to disk for gmsh to read
        with tempfile.NamedTemporaryFile(suffix='.bdf') as id_out:
            pass
        gmsh.write(id_out.name)
        file = open(id_out.name, 'r') 
        return file.read() 

but it will be ascii since it is a input file for nastran or gmsh... I do know how to work around this...

mikedh:
Cool looks good! I'll try to merge this later this week, I want to play with it a bit more.

bbarroqueiro:
I am been digging a little more in gmsh api. It should be possible to save the mesh in binary format using *.msh file extension. But it not working maybe i have an old SDK... i do not know...
 
For binary form it should be just 

gmsh.option.setNumber("Mesh.Binary",1)

but doesn't work... 

The Enhancement #206 is coved by this one and many other formats... 
The gmsh script generates the 3D mesh based on the 2D mesh and exports everthing... So, if the 3D mesh is skiped the msh file will contain just the surface mesh...

Issue #371, opened 2019-04-13
Does trimesh expect vertex arrays to be writeable?


pauldmccarthy:
Howdy,

I've been trying to eliminate unnecessary copies of `numpy` arrays in my code. I ran into a problem when creating `Trimesh` objects with read-only vertex arrays - the  `trimesh.intersections.mesh_plane` function seems to induce an attempt to write back to the vertex array - this code:

```python
import trimesh
import trimesh.intersections as intersect
import numpy as np

verts = np.array([
    [-1, -1, -1],
    [-1, -1,  1],
    [-1,  1, -1],
    [-1,  1,  1],
    [ 1, -1, -1],
    [ 1, -1,  1],
    [ 1,  1, -1],
    [ 1,  1,  1],
], dtype=np.float64)

tris = np.array([
    [0, 4, 6], [0, 6, 2],
    [1, 3, 5], [3, 7, 5],
    [0, 1, 4], [1, 5, 4],
    [2, 6, 7], [2, 7, 3],
    [0, 2, 1], [1, 2, 3],
    [4, 5, 7], [4, 7, 6],
], dtype=np.float64)

# force the arrays to be read-only 
tris  = np.ndarray(tris .shape, tris .dtype, buffer=bytes(tris .tostring()))
verts = np.ndarray(verts.shape, verts.dtype, buffer=bytes(verts.tostring()))

tm = trimesh.Trimesh(verts, tris, process=False, validate=False)

normal = np.random.random(3)
origin = np.random.random(3)

intersect.mesh_plane(tm,
                     plane_normal=normal,
                     plane_origin=origin,
                     return_faces=True)
```

produces this error:
```
Traceback (most recent call last):
  File "test.py", line 38, in <module>
    return_faces=True)
  File "/Users/paulmc/Projects/Public/trimesh/trimesh/intersections.py", line 181, in mesh_plane
    for c, h in zip(cases, handlers)])
  File "/Users/paulmc/Projects/Public/trimesh/trimesh/intersections.py", line 181, in <listcomp>
    for c, h in zip(cases, handlers)])
  File "/Users/paulmc/Projects/Public/trimesh/trimesh/intersections.py", line 139, in handle_basic
    line_segments=False)
  File "/Users/paulmc/Projects/Public/trimesh/trimesh/intersections.py", line 335, in plane_lines
    intersection += np.reshape(d, (-1, 1)) * line_dir[valid]
[...]

PR #372, opened 2019-04-13
Make sure that plane_lines does not attempt to write back to the orig…

pauldmccarthy:
Addresses #371.

Please feel free to ignore this PR if I should be passing writeable arrays to `trimesh` :)


mikedh:
Good catch, that shouldn't have been silently mutating vertices! If you have some cycles I'd definitely love a `test_mutate.py` that sets `mesh.vertices.flags['WRITEABLE'] = False` and then runs through a bunch of operations that shouldn't be touching vertices. If not I may be able to get to it eventually :sweat_smile: . 

Thanks for the PR!

PR #373, opened 2019-04-13
Tests for mesh with immutable vertex/face arrays

pauldmccarthy:
Here you go!

I also made a change to the `Trimesh` constructor - I'm assuming that this was a mistake, although perhaps the cache is shared across instances? Let me know if you'd like that reverted.

### DIFF
--- a/tests/test_mutate.py
+++ b/tests/test_mutate.py
@@ -1,0 +1,138 @@
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+import numpy as np
+
+import trimesh
+
+
+def readonly_mesh(model):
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome thanks! I'll try to figure out why this fails on Windows.

pauldmccarthy:
Looks like a long-standing issue - `np.bincount` needs to be passed a `dtype` which matches the native pointer type?

 - https://github.com/numpy/numpy/issues/823
 - https://github.com/numpy/numpy/pull/4366
 - https://github.com/numpy/numpy/issues/4384
 - https://github.com/nipy/dipy/pull/789

PR #377, opened 2019-04-16
Fix transform (non identity) use in creation.camera_marker

wkentaro:
## Before

Screencast 2019-04-16 18:45:47

## After

Screencast 2019-04-16 18:45:18

## Code (includes internal)

```python
import numpy as np
import trimesh


cad_file = '/home/wkentaro/Documents/trimesh/models/fuze.obj'

eyes = uniform_points_on_sphere(5, radius=0.5)
targets = np.full(eyes.shape, (0, 0, 0), dtype=float)
T_world2cam = np.array([
    look_at(eye, target)
    for eye, target in zip(eyes, targets)
])

# -----------------------------------------------------------------------------

scene = trimesh.Scene()

geom = trimesh.load(str(cad_file))
geom.visual = geom.visual.to_color()
scene.add_geometry(geom)

scene.camera.resolution = (640, 480)
scene.camera.fov = (60, 45)

for transform in T_world2cam:
    scene.camera.transform = transform
    geom = trimesh.creation.camera_marker(scene.camera, marker_height=0.1)
    scene.add_geometry(geom)

    geom = trimesh.creation.axis(0.01)
    geom.apply_transform(transform)
    scene.add_geometry(geom)

scene.set_camera()

scene.show()
```

PR #378, opened 2019-04-17
Test for creation.camera_marker with non identity transform

wkentaro:
Follow up of https://github.com/mikedh/trimesh/pull/377.


mikedh:
Thanks for the test!

Issue #368, opened 2019-04-11
Scene viewer as a widget in a window


wkentaro:
To use SceneViewer as a widget in a window (for real application with buttons, labels, and additional image widgets), [glooey](https://github.com/kxgames/glooey) looks good.
(`trimesh.SceneViewer` is a `pyglet.Window` by itself, so we can create multiple windows but we cannot tile multiple `SceneViewer` in a single window.)

Does anyone know any other options?

---

In glooey, it creates `pyglet.Window` and add Widgets (e.g., `glooey.Image`) to that.
So, in order to render `trimesh.Scene` or `trimesh.Trimesh` in a widget, we [need creates a custom widget](https://glooey.readthedocs.io/en/latest/making_widgets_from_scratch.html) (e.g, `TrimeshSceneViewerWidget`), which is similar to `trimesh.viewer.SceneViewer`.

Here is a workaround of this (actually it still doesn't work): https://gist.github.com/wkentaro/d8cc30c6634389db02ba53adb7b934b2

mikedh:
I played with this briefly- `glooey` looks awesome! Heh I tried their example, but just replacing the window like:
```
import trimesh
from trimesh.viewer import windowed
mesh= trimesh.load('models/featuretype.STL')
window = windowed.SceneViewer(mesh.scene(),
                              start_loop=False)
```
And it didn't work immediately. The widget would be a nice abstraction for sure, seems like it would however require a lot of re- implementation.

wkentaro:
At least, below example does work, right?

```
#!/usr/bin/env python

import glooey
import pyglet
import trimesh
from trimesh.viewer import windowed

mesh= trimesh.load('models/featuretype.STL')
window = windowed.SceneViewer(mesh.scene(),
                              start_loop=False)

gui = glooey.Gui(window)
pyglet.app.run()
```

But it doesn't work with other widget like `glooey.HBox`. I'm working on this by asking the developers of `glooey`: https://github.com/kxgames/glooey/issues/24.

wkentaro:
I successfully implemented the widget that visualizes `trimesh`'s scene object.
https://github.com/wkentaro/trimesh-glooey/blob/master/scene_widget.py

If you run this script, the window looks like below:
![Screencast 2019-04-20 20:02:50](https://user-images.githubusercontent.com/4310419/56461285-f30e4e00-63a7-11e9-8df2-d00ba87a86a9.gif)

[...]

PR #381, opened 2019-04-22
Use pyrender.trackball.Trackball as alternative to Arcball

wkentaro:
For https://github.com/mikedh/trimesh/issues/368#issuecomment-485153083

### DIFF
--- a/tests/test_trackball.py
+++ b/tests/test_trackball.py
@@ -1,0 +1,168 @@
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+from trimesh.viewer.trackball import Trackball
+
+
+class TrackballTest(g.unittest.TestCase):
+
+    def test_resize(self):
+        trackball = Trackball(
+            pose=g.np.eye(4),
+            size=(640, 480),
+            scale=1.0,
+            target=g.np.array([0, 0, 0])
+        )
+        trackball.down((320, 240))
+        trackball.drag((321, 240))
+        pose1 = trackball.pose
+        trackball.drag((320, 240))
+        trackball.resize((1280, 960))
+        trackball.down((640, 480))
+        trackball.drag((642, 480))
+        pose2 = trackball.pose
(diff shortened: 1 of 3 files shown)


wkentaro:
## Before




## After

wkentaro:
## Example Script

```python
#!/usr/bin/env python

import numpy as np
import trimesh
import trimesh.transformations as tf


mesh = trimesh.load('../models/fuze.obj')
scene = trimesh.Scene()
scene.add_geometry(mesh)
scene.set_camera(angles=[np.deg2rad(90), 0, 0])
scene.show()

scene = trimesh.Scene()
for _ in range(10):
    axis = trimesh.creation.axis(0.01)
    axis.apply_translation(np.random.uniform(-0.3, 0.3, (3,)))
    scene.add_geometry(axis)
scene.show()

```

mikedh:
Looks great, thanks! And thanks @mmatl for the nice trackball implementation. 

Only one question about the PR: it now changes the transform of `scene.camera` when you drag the window around, vs before it didn't mutate the scene. Is this desirable? I guess this way you can drag it to where you want, close the window, then save the camera transform. If we are going to mutate the camera transform in the viewer, maybe we should make the keypress `z` call `scene.set_camera`?

wkentaro:
> it now changes the transform of scene.camera when you drag the window around, vs before it didn't mutate the scene. Is this desirable?

I think the desirable situation is that `self.scene.camera` always has the same information as the current view like `self.scene.camera.resolution` is set on `__init__()`.

>  If we are going to mutate the camera transform in the viewer, maybe we should make the keypress z call scene.set_camera?

This is because `self.scene` can change, isn't it? So `z` key sets default camera pose more dynamically instead of resetting to the fixed camera pose.

mikedh:
Makes sense! Thanks for the PR!

wkentaro:
> we should make the keypress z call scene.set_camera?

I sent PR for this.
https://github.com/mikedh/trimesh/pull/383/files

PR #383, opened 2019-04-23
Dynamically reset camera by SceneViewer.scene.set_camera

wkentaro:
Follow up is https://github.com/mikedh/trimesh/pull/381


wkentaro:
But with this, we can't set initial camera pose, since `reset_view()` is called in `__init__()`.
What do you think of this? @mikedh

wkentaro:
Maybe this is not useful, so I close this.

PR #386, opened 2019-04-23
Visualize edge in trimesh.creation.box

wkentaro:
See right side of https://github.com/mikedh/trimesh/pull/387#issuecomment-485711294


mikedh:
Nice! I think I may move this functionality to `trimesh.path.creation.box_outline` and remove the boolean flags from `trimesh.creation.box`, just to keep path stuff in one place.

wkentaro:
Sounds nicer. :+1:

Issue #370, opened 2019-04-12
Definition of origin: indices_to_points and points_to_indices for voxel


wkentaro:
If I run below code

```python
import numpy as np
import trimesh
import trimesh.transformations as tf


matrix = np.ones((2,)*3, dtype=bool)
centers = trimesh.voxel.matrix_to_points(matrix, pitch=0.5, origin=(0,)*3)
print(centers)
centers = trimesh.voxel.matrix_to_points(matrix, pitch=0.5, origin=(0.5/2,)*3)
print(centers)
centers = trimesh.voxel.matrix_to_points(matrix, pitch=0.5, origin=(0.5,)*3)
print(centers)
```

the output is

```
[[-0.25 -0.25 -0.25]
 [-0.25 -0.25  0.25]
 [-0.25  0.25 -0.25]
 [-0.25  0.25  0.25]
 [ 0.25 -0.25 -0.25]
 [ 0.25 -0.25  0.25]
 [ 0.25  0.25 -0.25]
 [ 0.25  0.25  0.25]]
[[0.  0.  0. ]
 [0.  0.  0.5]
 [0.  0.5 0. ]
 [0.  0.5 0.5]
 [0.5 0.  0. ]
 [0.5 0.  0.5]
 [0.5 0.5 0. ]
 [0.5 0.5 0.5]]
[[0.25 0.25 0.25]
 [0.25 0.25 0.75]
 [0.25 0.75 0.25]
 [0.25 0.75 0.75]
 [0.75 0.25 0.25]
 [0.75 0.25 0.75]
 [0.75 0.75 0.25]
 [0.75 0.75 0.75]]
```

This shows currently, the definition of origin is right top corner of the first element. (①in the figure)

<img src=https://user-images.githubusercontent.com/4310419/56053642-91693680-5d4c-11e9-9682-5d950a78d326.png width=30% />

Maybe this is too late (since -0.5 is added for a while ago), but for me ② or ③ looks more reasonable as origin. By flipping sign of +-0.5 or removing it.
What do you think about this?

https://github.com/mikedh/trimesh/blob/f785528788e89712a34074c552004a79adeb8264/trimesh/voxel.py#L657
https://github.com/mikedh/trimesh/blob/f785528788e89712a34074c552004a79adeb8264/trimesh/voxel.py#L629

mikedh:
Not sure what the right answer here is. I think generally we probably want "points" to refer to the center of the voxel box?

wkentaro:
[...]

PR #388, opened 2019-04-23
Remove index shift by +-0.5 in voxel.py

wkentaro:
Close https://github.com/mikedh/trimesh/issues/370

**NOTE**: This is not backward compatible.

### DIFF
--- a/tests/test_voxel.py
+++ b/tests/test_voxel.py
@@ -102,30 +102,38 @@ class VoxelTest(g.unittest.TestCase):
                                                 fill=True)
 
     def test_points_to_from_indices(self):
-        points = [[0.05, 0.55, 0.35]]
+        # indices = (points - origin) / pitch
+        points = [[0, 0, 0], [0.04, 0.55, 0.39]]
         origin = [0, 0, 0]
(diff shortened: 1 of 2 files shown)


mikedh:
Nice! This was sort of haphazard before, it's great to make it consistent. A few quick thoughts:

- could we unit test?
- are there other places do we do the conversion points->indices? We should probably be consistent
- Maybe a quick blurb in the `voxel.py` top docstring which says something like "all voxels are referenced by the **center** of their cell box"

wkentaro:
> could we unit test?

I fixed the unit test just now.

> are there other places do we do the conversion points->indices? We should probably be consistent

As far as I know, the conversion is only used in `voxel.py` file. So I think now is good. 
The only concern is someone can assume `Voxel.as_boxes` returns boxes in the previous coordinate system. (origin is right-top of the first box)

> Maybe a quick blurb in the voxel.py top docstring which says something like "all voxels are referenced by the center of their cell box"

I agree. I added the docstring for Voxel class.

mikedh:
Awesome, thanks!

PR #393, opened 2019-04-23
Make widget example to be more practical

wkentaro:
This PR adds a more practical example of usage of the widget (e.g., changing camera location, changing scene, integrate with Image widget (other widget than SceneWidget)).

Screencast 2019-04-23 23:12:14


mikedh:
LGTM- thanks!

PR #394, opened 2019-04-24
Use SceneViewer.on_close for has_exit flag to be changed

wkentaro:
With close(), has_exit flag is always False.


mikedh:
Makes sense- thanks!

Issue #390, opened 2019-04-23
Style check tool?


wkentaro:
What tool are you using to check the style of code? And options for that?

```
% pip list | grep flake8
flake8              3.7.7 

% flake8 trimesh
trimesh/intersections.py:90:80: E501 line too long (83 > 79 characters)
trimesh/intersections.py:395:80: E501 line too long (81 > 79 characters)
trimesh/intersections.py:518:80: E501 line too long (81 > 79 characters)
trimesh/intersections.py:545:80: E501 line too long (82 > 79 characters)
trimesh/base.py:5:80: E501 line too long (82 > 79 characters)
trimesh/base.py:22:1: F401 '.bounds' imported but unused
trimesh/base.py:25:1: F401 '.nsphere' imported but unused
trimesh/base.py:32:1: F401 '.collision' imported but unused
trimesh/base.py:34:1: F401 '.smoothing' imported but unused
trimesh/base.py:155:80: E501 line too long (80 > 79 characters)
trimesh/base.py:156:80: E501 line too long (80 > 79 characters)
trimesh/base.py:232:80: E501 line too long (81 > 79 characters)
trimesh/base.py:465:5: F811 redefinition of unused 'bounds' from line 22
trimesh/base.py:658:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:811:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:860:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:941:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:1240:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:1330:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:1336:80: E501 line too long (80 > 79 characters)
trimesh/base.py:1400:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:1585:9: F841 local variable 'populate' is assigned to but never used
trimesh/base.py:1705:80: E501 line too long (83 > 79 characters)
trimesh/base.py:1707:80: E501 line too long (82 > 79 characters)
trimesh/base.py:1708:80: E501 line too long (81 > 79 characters)
trimesh/base.py:2238:80: E501 line too long (80 > 79 characters)
trimesh/base.py:2507:80: E501 line too long (81 > 79 characters)
trimesh/integrate.py:104:80: E501 line too long (83 > 79 characters)
[...]

PR #395, opened 2019-04-24
Run style check on CI

wkentaro:
Close https://github.com/mikedh/trimesh/issues/390

- Now we run only for trimesh/ (to gradually introduce).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,17 @@ install:
 # coveralls started failing to install on Python 3.4 so make it optional
 - pip install coveralls || true
 
+# install style checkers
+- pip install -U autopep8 flake8
+
 script:
+# style check
+- flake8 --max-line-length=90 trimesh/
+# - flake8 --max-line-length=90 example/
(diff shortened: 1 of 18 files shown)


mikedh:
Nice, looks terrific thanks! Only though is maybe we should run just `flake8` in CI? `autopep8` is more of a formatter than a linter.

wkentaro:
If you use `autopep8 --in-place`, I think it's worth to add `autopep8` to CI, because it always makes sure that there is no diff after running `autopep8`.
If you use `autopep8` as an information teller and manually change ignoring some of the warnings, I think we should remove `autopep8` from CI.

wkentaro:
Actually, I've learned this strategy from Chainer (a deep learning library):
https://github.com/chainer/chainer/blob/0fd15073532496ed495b9318c4fece4e077ce15f/scripts/ci/steps.sh#L104

Also I guess `--exit-code` is designed for this usage.

mikedh:
Oh nice, `--exit-code` looks good. Thanks for the PR!

PR #400, opened 2019-04-25
File Outputs

bbarroqueiro:
It could be interesting to add a gmsh to import FEM models extract skin and return the stl (trimesh array)...

Do you think it is interesting?

### DIFF
--- a/trimesh/interfaces/gmsh.py
+++ b/trimesh/interfaces/gmsh.py
@@ -66,8 +66,9 @@ def to_volume(mesh,
 
     if file_name is not None:
         # check extensions to make sure it is supported format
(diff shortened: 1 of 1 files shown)


mikedh:
Yeah that could be cool! The other thing that I was thinking was to load IGS and STEP files with `gmsh`, and then producing a 2D mesh. I played with it a little bit, and the loading worked well but I couldn't figure out how to reliably generate a watertight surface mesh. It would be pretty sweet to get STEP/IGS support though, gmsh loads them quite reliably. 

Thanks for the PR!

bbarroqueiro:
I'm not a gmsh expert (far from it...), but i think i'll have to generate the 3d mesh and then extract its skin i believe... At least on the Python Gmsh API, but when i have the time i can do i little more digging...

Issue #396, opened 2019-04-25
Why we need to call util.attach_to_log for warning and error messages?


wkentaro:
Currently, most of the warning and error messages are hidden unless using `util.attach_to_log`, for example:
https://github.com/mikedh/trimesh/blob/08507c9b1b69c61f26a56a74d13187f3d2602293/trimesh/exchange/wavefront.py#L177-L178

Normally, I think this behavior is undesired because we sometimes make mistakes (e.g., forget installing Pillow), and that causes issues that are hard to fix (e.g., texture is not loaded and displayed). A solution is we call `util.attach_to_log` in every script, but we don't want to do that (at least I don't).

IMO, I think we should show warnings and error messages as default (without calling `util.attach_to_log` like some of the warnings like:

https://github.com/mikedh/trimesh/blob/323a6f0e503b05cefbe11fbb8eedb5705fb7d01f/trimesh/points.py#L362-L366

What do you think of this?

wkentaro:
Probably, one option is attaching to the `logger` of `trimesh` at the level of `warning` as default. By doing this, it shows `warning`, `error`, and `fatal` messages without calling `attach_to_log`.

mikedh:
Oh yeah, it seems like any `warnings.warn` should be printed. I wonder why the warnings from inside trimesh are hidden, when I call `warnings.warn` from an IPython terminal they show up:

```
In [1]: import warnings

In [2]: warnings.warn('hi')
/home/mikedh/miniconda/bin/ipython:1: UserWarning: hi
  #!/home/mikedh/miniconda/bin/python

In [3]: import trimesh

In [4]: warnings.warn('hi')
/home/mikedh/miniconda/bin/ipython:1: UserWarning: hi
  #!/home/mikedh/miniconda/bin/python

In [5]: trimesh.util.attach_to_log()

In [6]: warnings.warn('hi')
WARNING        warnings.py:99    /home/mikedh/miniconda/bin/ipython:1: UserWarning: hi
  #!/home/mikedh/miniconda/bin/python

```

wkentaro:
With below change, the `warnings` is shown.
```
diff --git a/trimesh/constants.py b/trimesh/constants.py
index 872c6ee..650040b 100644
--- a/trimesh/constants.py
+++ b/trimesh/constants.py
@@ -126,7 +126,7 @@ res_path = ResolutionPath()
 
 # logging
 log = _getLogger('trimesh')
-log.addHandler(_NullHandler())
[...]

PR #402, opened 2019-04-26
Remove NullHandler from logger since it filters messages

wkentaro:
Close https://github.com/mikedh/trimesh/issues/396

It filters every warnings and errors.

I'm not sure if how you decided adding NullHandler and if it is right answer removing it.

### DIFF
--- a/trimesh/collision.py
+++ b/trimesh/collision.py
@@ -8,7 +8,6 @@ _fcl_exists = True
 try:
     import fcl  # pip install python-fcl
 except BaseException:
-    log.warning('No FCL -- collision checking will not work')
     _fcl_exists = False
 
 
--- a/trimesh/constants.py
+++ b/trimesh/constants.py
@@ -1,8 +1,8 @@
 from time import time as time_function
-from logging import getLogger as _getLogger
-from logging import NullHandler as _NullHandler
 
 import numpy as np
+
+from .util import log
 
 
 class ToleranceMesh(object):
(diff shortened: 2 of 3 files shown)


wkentaro:
Now it shows like below:
```
$ python
>>> import trimesh
>>> trimesh.load('models/fuze.obj')
failed to load texture: material_0
  File "/home/wkentaro/Documents/trimesh/trimesh/exchange/wavefront.py", line 171, in append_mesh
    image = PIL.Image.open(
NameError: name 'PIL' is not defined
<trimesh.base.Trimesh object at 0x7fbeca8c9358>
```

trimesh/collision.py
```diff
@@ -8,7 +8,6 @@
 try:
     import fcl  # pip install python-fcl
 except BaseException:
-    log.warning('No FCL -- collision checking will not work')
```
wkentaro: This is because it's too noisy, and make it consistent to the import of `PIL.Image` or other libraries.

mikedh:
Oh nice, thanks for tracking this one down! I think the `NullHandler` was probably added as the [examples in the Python docs do it](https://docs.python.org/2/howto/logging.html). Makes sense to get rid of it though.  

Thanks for the PR!

mikedh:
Also, it's pretty sweet that CI caught that `flake8` warning now :smile:

wkentaro:
Oops :sweat_smile: 

```
% flake8 trimesh/
trimesh/collision.py:5:1: F401 '.constants.log' imported but unused
```

PR #403, opened 2019-04-27
Import Capabilities enlarged

bbarroqueiro:
new formats
.brep
.stp or .step
.igs or iges
.bdf
.msh
.inp
.diff
.mesh

And i was wrong a LOT... gmsh does generate watertight in 2D in fact it is one on the first steps of the 3Dvmesher...

### DIFF
--- a/trimesh/interfaces/gmsh.py
+++ b/trimesh/interfaces/gmsh.py
@@ -8,6 +8,74 @@ try:
 except ImportError:
     gmsh = None
 
+def from_volume(file_name,max_element=1.):
+    """
+    Returns a surface mesh from CAD model in Open Cascade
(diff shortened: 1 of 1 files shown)


mikedh:
Whoa this is awesome thanks! Seems like it works pretty well! Although it is fairly slow relative to a CAD package for some reason (IE 3-4s vs <1s, larger STEP files take basically infinity time), I was playing with the mesher settings to see if I could get it to be more competitive.

One thing I noticed was killing performance that setting `MaxElement` without knowing the scale of the surface can lead to unnecessarily gigantic meshes, so I changed it to `Mesh.CharacteristicLengthFromCurvature` which seems ok.

PR #406, opened 2019-04-29
Protection

bbarroqueiro:
During smoothing if an element flips it raises an error and the angle becomes NAN 
Quite annoying and messes all operations with angles... but it would be important to raise an warning?

### DIFF
--- a/trimesh/triangles.py
+++ b/trimesh/triangles.py
@@ -110,6 +110,8 @@ def angles(triangles):
     # run the cosine and an einsum that definitely does something
     a = np.arccos(np.clip(np.einsum('ij, ij->i', u, v), -1, 1))
     b = np.arccos(np.clip(np.einsum('ij, ij->i', -u, w), -1, 1))
+    a=np.nan_to_num(a)
+    b=np.nan_to_num(b)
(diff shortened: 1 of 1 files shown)


mikedh:
Oh interesting, this is probably happening on degenerate triangles?
```
tri = np.array([[[0,0,0],[1,0,0],[1,0,0]]], dtype=np.float64)
```

I guess I have the question, what would the "correct" thing to return here be?  Right now as you say it's returning NaN:
```
array([[ 0., nan, nan]])
```

With `nan_to_num` it replaces `nan` with `0.0`. If that's the behavior we want we might be better off checking the size of the vector norm, which would eliminate the divide by zero warning as well:
```
u_norm = np.linalg.norm(u, axis=1, keepdims=True)
# vectors with all components are all zero
u_norm[u_norm < tol.zero] = 1.0
u /= u_norm
```

bbarroqueiro:
yes... exactly... that solves two problems at once.. great

mikedh:
I played around a bit, but in the end it was a ton of bookkeeping so I went with your solution of `np.nan_to_num` Thanks for the PR!

PR #407, opened 2019-04-30
1 dimensional rgb for PointCloud colors

wkentaro:
In addition to the 1 dimensional rgba like
```python
pcd = PointCloud(vertices=points_xyz)
pcd.colors = [1.0, 0, 0, 1.0]
pcd.colors = [1.0, 0, 0]  # this doesn't work without this PR
```


mikedh:
LGTM, thanks! Other option here would be to use `trimesh.visual.to_rgba`

PR #408, opened 2019-04-30
Delete modified vertex_list to avoid re-rendering

wkentaro:
This is necessary because always `glooey` tries to render everything that has registered to the batch group before. (In `SceneViewer` it calls `vertex_list[geometry_name].draw()`)


mikedh:
LGTM, thanks!

PR #411, opened 2019-05-02
Make visualization faster with updating the scene

wkentaro:
If I update change scene frequently, `scene.centroid` computes bounding box of the scene every time, which makes the visualization very slow.
I suppose in most case setting fixed `far` is fine. If we need to simulate a depth camera, which has minimum and maximum range, we can add `far` as an argument to `__init__()`. However, currently this is not necessary since rendering depth is not supported in `SceneViewer` and `SceneWidget` with OpenGL.

### DIFF
--- a/trimesh/viewer/widget.py
+++ b/trimesh/viewer/widget.py
@@ -78,7 +78,7 @@ class SceneGroup(pyglet.graphics.Group):
         self._set_view()
 
         SceneViewer._gl_set_background(self._background)
-        SceneViewer._gl_enable_depth(self.scene)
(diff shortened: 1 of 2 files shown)


mikedh:
Looks good! Yeah, I think a lot of this is, when you profile, is the naive caching approach in the scene graph. When the camera transform is updated, it dumps every other cached transform and recomputes (rather than figuring out which things are effected and only dumping those). I tuned a little bit in https://github.com/mikedh/trimesh/commit/9bd8abb9be45bbdc1f60556439f60b749bb88298 but can still be slow.

PR #416, opened 2019-05-06
load ply files having only vertices

uflorin:
Loading of ply files without faces was failing


mikedh:
Thanks for the PR! Would you mind adding a small (<300KB) file that has these properties to `models`, and possibly unit test that the colors are surviving the import?

Thanks!
Mike

mikedh:
I'm closing for now but feel free to re-open with a sample file so we can test the logic.

PR #434, opened 2019-05-24
vectorized voxel is_filled and point_to_indices implementations

jackd:
Existing `is_filled` and `point_to_indices` implementations are based on single query points. The proposed implementations vectorize these, alowing for arbitrary leading dimensions. They may be -slightly- slower for queries of the a single point/index, but I can't imagine by much.

Apart from a returned `np.ndarray` rather than a tuple in `point_to_index` I don't think there are any breaking changes here. `point_to_index` is arguably poorly named for this implementation, though I've left it as-is in the interest of back compatibility.


mikedh:
Looks great, thanks!

PR #436, opened 2019-05-27
[WIP] New OBJ Loader

mikedh:
Replaces the old OBJ loader with a faster one (roughly 3x) that is also hopefully clearer. The biggest reason we wanted to do this was to add support for meshes with multiple materials, which this does. 

## Breaking Changes
- Before the OBJ loader returned a list of meshes, and now it returns a `Scene` object. 
- Some functions moved, most notably materials are moved out of texture and into their own file, `trimesh.visual.material`. Renamed `wavefront.py` to `obj.py`. 

## Remaining Before Merge
- Decide whether to have loader return `list` of meshes as before or stick with `Scene` object. The `Scene` object is more consistent with trimesh's output on other formats. But, OBJ doesn't support the concept of transforms or instances so a list isn't that bad. I could go either way. 
- Some error checking around material loading
- See if there is some way to support face groups without it being a huge pain
- Fix tests
- Test on more sample models (i.e. can it load everything from shapenet and thingi10k).


jackd:
I'll be rendering all examples from 13 classes of shapenet core over the next few days - will post here if there aren't any issues.

jackd:
watercraft/planes all work well. [This line](https://github.com/mikedh/trimesh/blob/feature/obj/trimesh/exchange/obj.py#L188) seems to forget about the space after `mtllib` (skipping 6 chars rather than 7) which can be dealt with in resolver implementations, but seems a bit unecessary.

jackd:
[Example](https://gist.github.com/jackd/ad565fa6a5a3f626d3c05161c219e8d4) for [fallback TODO](https://github.com/mikedh/trimesh/blob/feature/obj/trimesh/exchange/obj.py#L397)

mikedh:
Ooh, looks like that file has mixed triangles and quads, I was looking for one like that to test the looping load path:
```
f 1//1 10//8 18//14 9//7
f 1//1 64//92 2//2
```
Is that file MIT-license compatible? It would be great to add that file to unit tests.

jackd:
rabbit mode is from [pyglet](https://bitbucket.org/pyglet/pyglet/src/default/contrib/model/examples/) which is [BSD 3](https://bitbucket.org/pyglet/pyglet/src/default/LICENSE)

mikedh:
Thanks! I think that model covers pretty much all the edge cases (mixed quad/tri/16-gon, partially backwards winding, etc) haha.

trimesh/exchange/obj.py
```diff
@@ -0,0 +1,517 @@
...
+            faces, faces_tex, normal_idx = _parse_faces(face_lines)
+
+        # TODO: where we should have usable face normals
+        vertex_normals = None
+        # if v is not None and vn is not None and len(v) == len(vn):
+        #    vertex_normals =
+
+        # try to get usable texture
+        visual = None
+        if faces_tex is not None:
```
jackd: This check prevents materials that use `Kd` (rather than `map_Kd`) without a uv mapping from being registered properly.

I got significantly nicer renderings by replacing this section with the code below (I also had to creating single-pixel `image`s for `SimpleMaterial`s without one based on `diffuse` color, but that's not to do with loading).
```python
        visual = None
        mesh = kwargs.copy()
        if faces_tex is not None:
            # convert faces referencing vertices and
            # faces referencing vertex texture to new faces
            # where each face
            new_faces, mask_v, mask_vt = unmerge_faces(
                faces=faces, faces_tex=faces_tex)

            # we should NOT have messed up the faces
            # note: this is EXTREMELY slow due to the numerous
            # float comparisons so only use in unit tests
            if tol.strict:
                assert np.allclose(v[faces], v[mask_v][new_faces])
            uv = vt[mask_vt]
            mesh['vertices'] = v[mask_v]
            mesh['faces'] = new_faces
        else:
            uv = np.zeros((v.shape[0], 2))
            mesh['vertices'] = v
            mesh['faces'] = faces

        if material in materials:
            visual = TextureVisuals(uv=uv, material=materials[material])
        else:
            log.warning('visual creation failed for submesh!', exc_info=True)
        mesh['visual'] = visual
        mesh['vertex_normals'] = vertex_normals
        geometry[util.unique_id()] = mesh
```

PR #449, opened 2019-06-06
writeable/consistency changes

jackd:
Fixed a bug where implicitly calculated fov/focal camera values could become out of sync following an update to resolution. The implementation is a bit clunky, but at least I'm fairly confident it works.

Other smaller stuff:
* Fixed ignored `resolution` argument in `Scene.set_camera`
* Added `flags.writeable = False` to various camera properties. I'm assuming the performance penalty on this isn't too dear?
*  numpy-ized some small stuff (I'm guessing that's faster than non-numpy? Quite possibly not now that I think about it... happy to roll back)

### DIFF
--- a/tests/test_camera.py
+++ b/tests/test_camera.py
@@ -44,6 +44,42 @@ class CameraTests(g.unittest.TestCase):
             fov=None)
         assert np.allclose(camera.fov, fov)
 
+    def test_focal_updates_on_resolution_change(self):
+        """Test changing resolution with set fov updates focal."""
+        base_res = (320, 240)
+        updated_res = (640, 480)
+        fov = (60, 40)
+        base_cam = g.trimesh.scene.Camera(
+            resolution=base_res,
+            fov=fov)
+        base_focal = base_cam.focal
+        base_cam.resolution = updated_res
+        assert base_cam.focal is not base_focal
+        new_cam = g.trimesh.scene.Camera(
+            resolution=updated_res,
+            fov=fov,
+        )
+        np.testing.assert_allclose(base_cam.focal, new_cam.focal)
+
+    def test_fov_updates_on_resolution_change(self):
+        """Test changing resolution with set focal updates fov."""
+        base_res = (320, 240)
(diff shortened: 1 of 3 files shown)


mikedh:
Nice, thanks for the fix! Would you mind also adding a test which demonstrates the consistency issue?

jackd:
Added :)

mikedh:
Sweet thanks!

mikedh:
Man, bookkeeping is rough. One downside here is that if you re-order the calls to set resolution and focal length, the behavior is substantially different:

This fails:

```
        base_res = (320, 240)
        updated_res = (640, 480)
        fov = (60, 40)

        # start with initial data                                        
        base_cam = g.trimesh.scene.Camera(
            resolution=base_res,
            fov=fov)
        # update both focal length and resulution                        
        base_focal = base_cam.focal
        base_cam.resolution = updated_res
        print(base_cam.focal, base_focal)
        assert not g.np.allclose(base_cam.focal,
                                 base_focal)


        # try the same thing in reverse order                            
        base_cam = g.trimesh.scene.Camera(
            resolution=base_res,
            fov=fov)
        # update both focal length and resolution                        
        base_cam.resolution = updated_res
        base_focal = base_cam.focal
        print(base_cam.focal, base_focal)
        assert not g.np.allclose(base_cam.focal,
                                 base_focal)

```
With:
```
[554.25626 659.39458] [277.12813 329.69729]
[554.25626 659.39458] [554.25626 659.39458]
F...
======================================================================
FAIL: test_focal_updates_on_resolution_change (__main__.CameraTests)
----------------------------------------------------------------------
  File "test_camera.py", line 76, in test_focal_updates_on_resolution_change
    base_focal)
AssertionError
```

I don't think this is new behavior, but changing the order of setting things probably shouldn't wildly change the result. I'll merge into master as it's the same or better than before, but we should probably re-visit (I'll make a new issue: #451 ).

mikedh:
Thanks for the PR btw!

PR #453, opened 2019-06-11
scale needed for gltf loader

dapisani:
Noticed this issue when loading some glb files.


mikedh:
Looks awesome, thanks! When I get some time I still mean to run our loader on every one of the [GLTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models) to try to nip stuff like this in the bud.

PR #464, opened 2019-06-18
Safely copy a Camera instance

wkentaro:
I've found https://github.com/mikedh/trimesh/pull/463 is not enough.

### DIFF
--- a/trimesh/scene/cameras.py
+++ b/trimesh/scene/cameras.py
@@ -72,9 +72,18 @@ class Camera(object):
 
     def copy(self):
         """
-        Get a copy of the current camera.
+        Safely get a copy of the current camera.
         """
-        return copy.deepcopy(self)
+        scene = None if self.scene is None else self.scene.copy()
+        transform = None if self.transform is None else self.transform.copy()
+        return Camera(
+            name=copy.deepcopy(self.name),
+            resolution=copy.deepcopy(self.resolution),
+            focal=copy.deepcopy(self.focal),
+            fov=copy.deepcopy(self.fov),
+            scene=scene,
+            transform=transform,
+        )
 
     @property
     def resolution(self):
--- a/trimesh/scene/scene.py
+++ b/trimesh/scene/scene.py
@@ -856,7 +856,8 @@ class Scene(Geometry):
         geometry = {n: g.copy() for n, g in self.geometry.items()}
         # create a new scene with copied geometry and graph
         copied = Scene(geometry=geometry,
-                       graph=self.graph.copy())
+                       graph=self.graph.copy(),
+                       camera=self.camera.copy())
         return copied
 
     def show(self, viewer=None, **kwargs):


mikedh:
LGTM, thanks! If you have some cycles would you mind PR'ing a test that failed with the old `camera.copy` method but succeeds with this to demonstrate the edge case?

wkentaro:
Sorry, maybe I was misunderstanding something. Maybe this change is not so necessary since `camera.transform` is immutable. If I found something wrong, I will try to fix again.

wkentaro:
It seems f7ccfdc introduces something wrong. My script doesn't work with error `networkx.exception.NodeNotFound: Either source world or target camera_QS3VXO is not in G`.

wkentaro:
I realized we can reproduce it with
```python
In [5]: history
import trimesh
camera = trimesh.scene.Camera(resolution=(320, 240), fov=(60, 45))
scene = trimesh.Scene(camera=camera)
scene.camera.transform
history
```

wkentaro:
Maybe should we revert these? I didn't expect it causes this fatal errors..

wkentaro:
I found the critical part is here:

```
diff --git a/trimesh/scene/scene.py b/trimesh/scene/scene.py
index 2b2a063..e875e2f 100644
--- a/trimesh/scene/scene.py
+++ b/trimesh/scene/scene.py
@@ -499,8 +499,8 @@ class Scene(Geometry):
         """
         if camera is None:
             return
-        # assign the scene reference here
-        camera.scene = self
+        # # assign the scene reference here
+        # camera.scene = self
         self._camera = camera
 
     @property

```

wkentaro:
I still don't understand why we need to set `camera.scene` yet, since we can change `scene.camera.transform` and that is fine, no?

mikedh:
Hey yeah, I didn't realize `test_scene` was quite so sensitive, `camera` has a reference to `scene`, `scene` has a reference to `camera`.

Happy to accept fixes! Also if you could add your scenario that should have worked to `test_camera` or `test_scene` that would probably make it easier to figure out what the right move here is.

wkentaro:
Currently, I'm working on a visualization like this: https://gist.github.com/wkentaro/f0282e82ce234323ad01ffec71b9ac6d


Another situation, (actually this is where I decided to add `camera.copy`) is like this: https://gist.github.com/wkentaro/cc03b809b0be34153ec689f01eab0718
image
(OpenGL transform causes flipped camera marker)

Maybe we can discuss based on the second example.
In general, I use OpenCV coordinate system (+z facing to the scene) for scene understanding. However, for visualization with opengl, I need to change camera transform with `to_opengl_transform` function.
So `camera.transform` usually in OpenGL coordinate system (-z facing to the scene), but the `creation.camera_marker` function I added assumes it is OpenCV coordinate system (this is one confusing part).
However, I want to display the camera marker as if in OpenCV coordinate, so current behavior of `camera_marker` is fine at least for me. But before passing the camera instance to `creation.camera_marker`, I need to flip the transform to OpenCV coordinate ([here](https://gist.github.com/wkentaro/cc03b809b0be34153ec689f01eab0718#file-show_camera-py-L46-L49)). This is why I needed `camera.copy` to generate new camera instance.

Maybe `camera_marker` should be fixed, since `trimesh` should consistently use opengl coordinate, and I can create `camera_marker_opencv` function in my fork. But still, I will need `camera.copy` for this function.

PR #467, opened 2019-06-19
Major Release (3.0.0) 

mikedh:
A major release, which includes some API changes. Plan is to merge 7/15/2019.

- Includes voxel re-write from #437
- Includes OBJ re-write from #436 
  - Load performance is now ~3x better, and is more competitive with compiled options (a 37mb OBJ was loaded by `meshlab` in 1.4s, and `trimesh` in 2.3s)
- Reduces *hard* requirements from `networkx`/`scipy`/`numpy` to just `numpy`

#### OBJ
- OBJ files with a single object and material still return a single `trimesh.Trimesh` object
- OBJ files with multiple objects or material now return a `trimesh.Scene` object, rather than a list of `trimesh.Trimesh` objects
- Now loads most of shapenet and the assimp example corpus. 

#### Scene
- `scene.camera.transform` is now `scene.camera_transform` to avoid a circular reference

#### Docker
- images are now based off `debian:buster-slim` (upgraded from `debian:stretch-slim`)
- Images no longer come installed with `blender` and `openSCAD`, to avoid X11 in every image. 

#### GLTF
- Now cleanly loads every model in the official GLTF sample corpus

#### Voxel

- The `Voxel` class has undergone a major rewrite and been renamed `VoxelGrid`. It is now defined by separate encoding and transformation_matrix objects and can be thought of as a 3D occupancy grid with unit side lengths that is transformed lazily by an arbitrary transformation matrix. It now implements Geometry in its own right.
- `VoxelGrid.pitch` and `VoxelGrid.origin` are still provided for ease-of-transition, but for new code prefer to use `Voxel.transformation_matrix` directly. Note that `VoxelGrid.pitch` is ill-defined for transformation matrices which aren't diagonal apart from the translation part and will raise an error when computed if so.
- `trimesh.voxel.encoding.Encodings` classes have been added which represent various encodings of ND arrays - e.g. dense, sparse, run-length-encoding etc. Most methods mimic numpy equivalents, though not all numpy implementations are required so an Encoding is not a numpy array. As a fallback, they contain `dense` which gives the dense numpy array.
-  `VoxelMesh` class has been removed.
- `Trimesh.voxelized` has a new method - `binvox` - which uses the binvox CL tool.
- Filling in voxels can be achieved using `Voxel.fill` (operates in-place) and now supports a number of implementations.
- Added functionality:
      - hollow: remove internal voxels
      - strip: strip leading/trailing planes of zeros
      - revoxelized: create a new VoxelGrid based on a different shape with the same bounds.
- `util.grid_linspace` and `util.grid_arange` now using `numpy.meshgrid` with `indexing='ij'`. As a result, the returned arrays have the same rows, just in different orders.
- `caching.DataStore` now implements `collections.Mapping`, and raises a `KeyError` if one attempts to access attributes that are not present in the underlying data dict. Old behavior can be replicated by replacing `data[key]` with `data.get(key, np.empty((0,)))`.
- Various apply_transforms now return the object having the transformation applied to (previous versions returned None). This makes chaining operations easier in some instances. Note they still mutate the object.


trimesh/exchange/obj.py
```diff
@@ -0,0 +1,535 @@
...
+                faces=faces, faces_tex=faces_tex)
+
+            # we should NOT have messed up the faces
+            # note: this is EXTREMELY slow due to the numerous
+            # float comparisons so only use in unit tests
+            if tol.strict:
+                assert np.allclose(v[faces], v[mask_v][new_faces])
+
+            try:
+                visual = TextureVisuals(
```
jackd: Could this functionality be added outside the `if faces_tex is not None` block? The UVs might be rubbish, but the materials are still useful. (Demo)[https://gist.github.com/jackd/ca9d4ebf39f682dea43f9979d1470a9b] showing an example that gets much nicer with the below change. I know the demo code is a disgusting dirty hack and a proper implementation of non-image based renderings would likely be preferable but I really don't have the expertise for that, and the loader would still need to be modified to save the relevant material data.

```
        visual = None
        mesh = kwargs.copy()
        if faces_tex is not None:
            # convert faces referencing vertices and
            # faces referencing vertex texture to new faces
            # where each face
            new_faces, mask_v, mask_vt = unmerge_faces(
                faces=faces, faces_tex=faces_tex)

            # we should NOT have messed up the faces
            # note: this is EXTREMELY slow due to the numerous
            # float comparisons so only use in unit tests
            if tol.strict:
                assert np.allclose(v[faces], v[mask_v][new_faces])
            uv = vt[mask_vt]
            mesh['vertices'] = v[mask_v]
            mesh['faces'] = new_faces
        else:
            uv = np.zeros((v.shape[0], 2))
            mesh['vertices'] = v
            mesh['faces'] = faces

        if material in materials:
            visual = TextureVisuals(uv=uv, material=materials[material])
        else:
            log.warning('visual creation failed for submesh!', exc_info=True)
        mesh['visual'] = visual
        mesh['vertex_normals'] = vertex_normals
        geometry[util.unique_id()] = mesh

    if len(geometry) == 1:
        ....
```

trimesh/voxel/base.py
```diff
@@ -0,0 +1,357 @@
...
+        """int, number of occupied voxels in the grid."""
+        return self.encoding.sum.item()
+
+    def is_filled(self, point):
+        """
+        Query points to see if the voxel cells they lie in are filled or not.
+
+        Parameters
+        ----------
+        point: (n, 3) float, point(s) in space
```
jackd: In case it wasn't clear from my original documentation, the (..., 3) was meant to indicate any number of leading dimensions.

trimesh/exchange/obj.py
mikedh: Oh yeah, good point. There were a couple issues with people asking for that (materials with no UV's but uses the ambient/diffuse color properties).

trimesh/voxel/base.py
mikedh: Heh it takes arrays with `len(points.shape) == 2 and points.shape[1] == 3` right? Not arrays with shape `(a, b, c, d, e, f, g)` (etc)?

jackd: If I haven't stuffed up it should take arrays with shape `(a, b, c, d, e, f, g, 3)` and return arrays of shape `(a, b, c, d, e, f, g)`. This makes usage like in `revoxelized` slightly more convenient.

jackd:
I found the `camera_transform` change a bit counter-intuitive - particularly given `camera.transform` is still publicly accessible/mutable, and mutating it doesn't affect `camera_transform`. What's the circular reference issue it's trying to avoid?

mikedh:
Hey @jackd I replied in the PR that made the change: #469

JulienPalard:
Why did I received an email about this release? Are you just spamming every Python developers? Why?

PR #468, opened 2019-06-19
Reduce 3MF loading memory usage

luis-banuelos-chacon:
Reduce memory usage while loading big 3MF files by stream-parsing the model XML instead of completely loading into memory.

I ran into this problem while trying to load a 150MB 3MF file which contains around 200 objects. When decompressing, the file size jumps to around 1.5GB and loading the XML model with LXML balloons up to 9GB. My solution to this problem was to utilize etree.iterparse to stream each of the objects to be processed and clean the tree afterward, dramatically reducing memory usage to around 2GB.

I tried to making only minimal changes to the code.


mikedh:
Looks great, thanks!

PR #470, opened 2019-06-21
Support gltf scenes with multiple unnamed meshes

lordi:
When the glb file contains meshes that have no name, trimesh would choose "GLTF_geometry" for each one and thus overwrite prior ones. This PR will fix this. Test included.


mikedh:
Good catch, thanks!

PR #479, opened 2019-07-01
Small BRLE tweaks and VoxelGrid updates for breaking changes

jackd:
* Initial BRLE required even number of values. This constraint was removed, but functions still did unnecessary padding. I've removed this padding.
* A small bug fix with BRLE gathering
* Fixed runlength documentation
* changed mutability setter to set `mutable` property of values, rather than set `flags['WRITABLE']` and added `mutable` setter to `TrackedArray`. Could alternatively add `flags` implementation to `encodings` - just thought the former was simpler. Happy to be convinced.
* fixed writeability issues with loading from binvox. This involves a change from `np.frombuffer` to `np.fromstring` in which may be __slightly__ slower. I don't think it will be massive, but we could tweak mutable setter to make a data copy only when set to True.
* fixed VoxelGrid tests/examples (`transform_matrix` -> `transform`)

Didn't realize you were working on caching stuff just now - hope my changes aren't getting in the way.

### DIFF
--- a/examples/voxel.py
+++ b/examples/voxel.py
@@ -27,15 +27,15 @@ def show(chair_mesh, chair_voxels, colors=(1, 1, 1, 0.3)):
 
 base_name = 'chair_model'
 chair_mesh = trimesh.load(os.path.join(dir_models, '%s.obj' % base_name))
-if isinstance(chair_mesh, (list, tuple)):
+if isinstance(chair_mesh, trimesh.scene.Scene):
     chair_mesh = trimesh.util.concatenate([
-        trimesh.Trimesh(mesh.vertices, mesh.faces) for mesh in chair_mesh])
+        trimesh.Trimesh(mesh.vertices, mesh.faces)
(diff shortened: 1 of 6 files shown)


mikedh:
Thanks for the PR! Yeah I ran in to an issue with mutability this morning, so I added some tests and logic changes to `DataStore`, though it looks like you've branched off the latest so no horrible merge conflicts haha.

>> fixed writeability issues with loading from binvox. This involves a change from np.frombuffer to np.fromstring in which may be slightly slower. I don't think it will be massive, but we could tweak mutable setter to make a data copy only when set to True.

Isn't `np.fromstring` without a `sep` kwarg [depricated](https://github.com/nschloe/meshio/issues/252)?

Everything else looks great! Thanks for fixing the `tranform_matrix` -> `transform` references, I made that change to make things less verbose, as well as to match the API for `trimesh.primitives.Primitive` objects.

mikedh:
Thanks for the fix!

PR #482, opened 2019-07-04
allowed zero triangles in windings_aligned

jackd:
I was having issues rendering shapenet ships with the new obj loader. Not sure what the motivation is for not allowing `0` dimensions to match wildcards, but default behaviour is the same with the util method - just allows more flexibility.

### DIFF
--- a/trimesh/triangles.py
+++ b/trimesh/triangles.py
@@ -299,8 +299,9 @@ def windings_aligned(triangles, normals_compare):
       Are normals aligned with triangles
     """
     triangles = np.asanyarray(triangles, dtype=np.float64)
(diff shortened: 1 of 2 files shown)


mikedh:
Looks great, thanks!  I currently have a script to load every OBJ and BINVOX from shapenet (hopefully I can do thingi10k and few other corpuses before release) until one errors. It's pretty solid at the moment, though the texture coordinates from OBJ files are kind of screwy for reasons I haven't figured out yet.

By the way, could you take a look at `test_runlength` when you get a chance? It's failing, and I'm not entirely sure what it should be doing haha.

PR #483, opened 2019-07-05
texture loading when faces_tex is None, import tweaks

jackd:
Took me a little bit to work out which packages to install for tests - hence the comment/separated import.

Can provide an example where first commit fixes things if necessary, but it's a bit involved...


mikedh:
Looks much better, thanks!

Issue #486, opened 2019-07-05
`camera_to_rays` and renderer inconsistencies?


jackd:
I'm hunting a bug related to the consistency between rendered images and silhouettes of voxels but I'm at a loss. The following example using the current `release/major` branch:
* loads a mesh
* voxelizes via binvox (with `exact=True`)
* computes frustum coordinates via a slightly modified `camera_to_rays` (though the error persists using the trimesh implementation)
* uses `VoxelGrid.is_filled` to get occupancy values on the frustum coordinates
* reduces along the final dimension to get a silhouette.

Unfortunately, the silhouette does not match the image rendered via `scene.save_image`. I've either introduced a bug with the voxel stuff, failing to interpret something correctly, or there's something going on with cameras inconsistently between openGL rendering and `camera_to_rays`.

```python
import os
import trimesh
import numpy as np
from PIL import Image

from trimesh.geometry import align_vectors
from trimesh import util
from trimesh import transformations


def camera_to_rays(camera, transform):
    """
    Differences to trimesh.camera.camera_to_rays:

    * returned origin is not tiled
    * angles not returned
    * returned rays not normalized - have unit z-component before transform
    * rays point to center of pixels.
    """
    # radians of half the field of view
    half = np.radians(camera.fov / 2.0)
    # scale it down by two pixels to keep image under resolution
    # half *= (camera.resolution - 2) / camera.resolution

    bottom_right = np.sin(half)
    # move half a pixel in
    ## pixel_size = (bottom_right - top_left) / camera.resolution
    # pixel_size = bottom_right*2 / camera.resolution
    # bottom_right -= pixel_size / 2
    bottom_right *= 1 - 1. / camera.resolution
    
    top_left = -bottom_right

    xy = util.grid_linspace(
        bounds=[top_left, bottom_right], count=camera.resolution)
    # vectors = util.unitize(np.column_stack((xy, np.ones_like(xy[:, :1]))))
    vectors = np.column_stack((xy, np.ones_like(xy[:, :1])))
    # skip normalization
    # vectors = util.unitize(vectors)
    angles = np.arcsin(xy)

    # flip the camera transform to change sign of Z
    transform = np.dot(
        transform,
        align_vectors([1, 0, 0], [-1, 0, 0]))

    # apply the rotation to the direction vectors
    vectors = transformations.transform_points(
        vectors,
        transform,
        translate=False)

    # camera origin is single point, extract from transform
    origin = transformations.translation_from_matrix(transform)
[...]

PR #488, opened 2019-07-09
Fixed camera_to_rays implementation, slightly modified interface

jackd:
Resolved [this issue](https://github.com/mikedh/trimesh/issues/486). In addition, made the following interface changes to `camera_to_rays` (and `Camera.to_rays`):

* `angles` is no longer an intermediate value, so is not calculated/returned.
* `origin` is no longer tiled to match `vectors`' shape.
* `vectors` in not longer `unitize`d. This means rays point to the image plane rather than lying on a sphere. This makes it easier to make frustums with `z_near` and `z_far` values. They can always be unitized afterwards by the caller.

These interface changes can be accepted/rejected independently - just thought I'd throw them all in the same PR for convenience, but happy to change them back.

~~I've had to introduce a slight [hack](https://github.com/jackd/trimesh/blob/release/major/examples/raytrace.py#L36) to get the raytrace example working. I have no idea why this is necessary... if that makes sense/can be justified by changes elsewhere feel free to fix things up, otherwise I'll look into it when I get a chance.~~

There's a slight hack [here](https://github.com/jackd/trimesh/blob/release/major/examples/voxel_silhouette.py#L24) to make things work. Feels a bit dirty - I'm just reordering the rays so the output silhouette matches the image - but I'm not sure why it's necessary...

### DIFF
--- a/trimesh/scene/cameras.py
+++ b/trimesh/scene/cameras.py
@@ -256,7 +256,7 @@ class Camera(object):
         angles : (n, 2) float
           Ray spherical coordinate angles in radians.
         """
-        return np.arctan(_ray_pixel_coords(self))
+        return np.arctan(-_ray_pixel_coords(self))
         
 
 
@@ -319,12 +319,7 @@ def look_at(points, fov, rotation=None, distance=None, center=None):
 
 
 def _ray_pixel_coords(camera):
-    # radians of half the field of view
-    half = np.radians(camera.fov / 2.0)
-    # scale it down by two pixels to keep image under resolution
-    # half *= (camera.resolution - 2) / camera.resolution
-
-    bottom_right = np.tan(half)
+    bottom_right = np.tan(np.radians(camera.fov / 2.0))
     # move half a pixel width in
     ## pixel_size = (bottom_right - top_left) / camera.resolution
     # pixel_size = bottom_right*2 / camera.resolution
--- a/trimesh/scene/cameras.py
+++ b/trimesh/scene/cameras.py
@@ -245,19 +245,19 @@ class Camera(object):
           Ray direction unit vectors
         """
         return camera_to_rays(self, transform)
-    
+
     def angles(self):
         """
         Get ray spherical coordinates in radians.
 
-        
+
         Returns
         --------------
         angles : (n, 2) float
           Ray spherical coordinate angles in radians.
         """
         return np.arctan(-_ray_pixel_coords(self))
-        
+
 
 
 def look_at(points, fov, rotation=None, distance=None, center=None):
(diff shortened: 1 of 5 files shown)


trimesh/scene/cameras.py
```diff
@@ -307,6 +318,23 @@ def look_at(points, fov, rotation=None, distance=None, center=None):
     return cam_pose
 
 
+def _ray_pixel_coords(camera):
```
mikedh: What's the reason to break `_ray_pixel_coords` out from `camera_to_rays`? Is it used elsewhere?

mikedh:
Nice! I knew you'd get it in the end 😄. I had a quick API question which I left inline. 

Is there any simple check we could run in unit tests for camera-ray correspondence? Maybe check if everything is inside angle bounds, or something? Or if you were feeling fancy, create an optical plane of points and check that each vector comes within epsilon of its expected pixel location?

trimesh/scene/cameras.py
jackd: in `Camera.angles`

mikedh: Ah gotcha. Any reason to not make it "public-ish"?

jackd: Happy to do so, I just don't think it's currently very well designed/named. I'm unsure if it'd be used much elsewhere, and whether people will know to find it even if it is. The description is something along the lines of

> the `xy` coordinates of pixels in the `z = -1` plane and the camera is looking along the `-z` axis. Reshaping to `res_x, res_y, 2` will give you the image coordinates except it's flipped up-down for some reason I don't understand.`

See comment below about flipping issue.

jackd:
Not sure I quite follow what you mean in terms of tests - and before this gets accepted I'd really like to get to the bottom of the flipping issue. I'm not really a graphics person - and the reason I've tried to move a lot of my stuff towards this package is that I don't have to be :) . But I can't for the life of me work out why I need to keep flipping things up/down to make them consistent. Is it something to do with a front-back reflection, because the image plane is behind the camera or something? Differences in left-handed vs right-handed coordinate frames?

The folowing demonstrates the need to flip (and potentially a starting point for the tests you were mentioning above). I create a scene with default camera and put a pointcloud down at the ray endpoints (on the z=-1 plane) with dynamic coloring. We then compare renderings to the point cloud colors. Qualitatively the results match (so long as you include the mysterious flip). I could easily embed that flip into `_ray_pixel_coords`, but I'd like to understand why that's necessary so as not to introduce further bugs down the line...

```python

import numpy as np
import trimesh
from PIL import Image
import os

# bugged impl - cannot set_camera on empty scene
# scene = trimesh.Scene()
# scene.set_camera()

# create a sphere that allows us to get a camera that doesn't raise
sphere = trimesh.primitives.Sphere()
scene = sphere.scene()

origin, rays = scene.camera.to_rays(scene.camera_transform)
num_rays = len(rays)
endpoints = rays + origin

# create distinguishable colors
colors = trimesh.util.grid_linspace([[0, 0], [1, 1]], scene.camera.resolution)
colors = np.concatenate((colors, np.zeros_like(colors[..., :1])), axis=-1)
colors = (colors * 255).astype(np.uint8)

scene.add_geometry(trimesh.PointCloud(endpoints, colors=colors))
image = np.array(Image.open(trimesh.util.wrap_as_stream(
    scene.save_image(resolution=None))))
image = image[:, :, :3]

colors = np.reshape(colors, tuple(scene.camera.resolution) + (3,))
colors = colors[:, -1::-1]  # why is this necessary?
# make y-x ordering for comparison to image
colors = np.transpose(colors, (1, 0, 2))


def vis():
    _, ax = plt.subplots(1, 3)
    ax[0].imshow(colors)
    ax[0].set_title('colors flipped up-down')
    ax[1].imshow(image)
    ax[1].set_title('rendering')
    diff = np.abs(image.astype(int) - colors.astype(int))
    print('max difference: %s' % np.max(diff, axis=(0, 1)))
    diff = diff / np.max(diff)
    ax[2].imshow(diff)
    ax[2].set_title('rescaled difference')
    plt.show()


vis()
```

jackd:
Sorry for the spam. After drawing some pictures and waving my right hand in a pistol-grip around for a bit, I'm now satisfied why I had to flip things around and have effectively moved that flip inside the now public `ray_pixel_coords` function. Also implemented changes I discussed [here](https://github.com/mikedh/trimesh/pull/469) (happy to roll-back if you don't like them).

Let me know if there's anything else needed for this.

mikedh:
Looks great, thanks for all the work on voxels!! Just pressed merge :sweat_smile:

Issue #475, opened 2019-06-27
OBJ/MTL: Material information not used


mfeldt:
When reading a .obj + associated .mtl file, the material information is apparently ignored.
While the loading itself works (verified by trimesh.util.attach_to_log() and watching the error after removing the .mtl file), the rendered structure when using .show() is always grey.  

It works with the fuze example that has an associated jpg map, but apparently not with simple material definitions.

mikedh:
Yeah, currently this is true. The OBJ loader is being rewritten, which we'll hopefully release mid-July in https://github.com/mikedh/trimesh/pull/467

jackd:
@mfeldt Is this resolved in the latest `release/major` branch? If not, can you provide an example failure case?

mikedh:
Hey @jackd I think the new OBJ leader is currently transferring the properties correctly to `SimpleMaterial.ambient` etc, but `trimesh.viewer.windowed` and `trimesh.rendering` isn't using those values the moment and is still showing previews in the default grey.

jackd:
Oh yeah. I currently have super hacky stuff which creates single-pixel values after loading.

```python
def fix_visual(visual, nv):
    from PIL import Image
    if isinstance(visual, trimesh.visual.ColorVisuals):
        return
    material = visual.material
    assert(hasattr(material, 'image'))
    if material.image is None and material.diffuse is not None:
        image = np.array(material.diffuse)
        image = (image*255).astype(dtype=np.uint8).reshape((1, 1, 3))
        image = Image.fromarray(image)
        material.image = image
    if visual.uv is None:
        visual.uv = np.zeros((nv, 2))


def fix_geometry_visuals(geometry):
    visual = getattr(geometry, 'visual', [])
    nv = geometry.vertices.shape[0]
    if hasattr(visual, '__iter__'):
        for v in visual:
            fix_visual(v, nv)
    else:
        fix_visual(visual, nv)


def fix_scene_visuals(scene):
    geometry = scene.geometry
    if hasattr(geometry, 'values'):
        for v in geometry.values():
            fix_geometry_visuals(v)
    elif hasattr(geometry, '__iter__'):
        for v in geometry:
            fix_geometry_visuals(v)
    else:
        fix_geometry_visuals(v)
```

mfeldt:
> @mfeldt Is this resolved in the latest `release/major` branch? If not, can you provide an example failure case?

Apparently not @jackd . Loading my files leads to

```
a=trimesh.load('step.obj')
[...]

PR #490, opened 2019-07-09
added support for materials with names with spaces

jackd:
Partially resolves [this issue](https://github.com/mikedh/trimesh/issues/475). Not sure if spaces in newmtl names are officially part of the spec, but it is supported by meshlab and it doesn't add much.

### DIFF
--- a/trimesh/exchange/obj.py
+++ b/trimesh/exchange/obj.py
@@ -429,7 +429,7 @@ def parse_mtl(mtl, resolver=None):
                 materials[material.pop('newmtl')] = material
 
             # start a fresh new material
-            material = {'newmtl': ' '.joint(split[1:])}
+            material = {'newmtl': ' '.join(split[1:])}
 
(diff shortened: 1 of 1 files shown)


trimesh/exchange/obj.py
```diff
@@ -429,7 +429,7 @@ def parse_mtl(mtl, resolver=None):
                 materials[material.pop('newmtl')] = material
 
             # start a fresh new material
-            material = {'newmtl': split[1]}
+            material = {'newmtl': ' '.joint(split[1:])}
```
mikedh: Shouldn't `joint` be `join`?

mikedh:
Thanks for the PR! LGTM other than the typo. Who knew people would use spaces 🔥

jackd:
... one day I'll skip the testing phase and actually get away with it...

mikedh:
Haha they take so long now.

PR #493, opened 2019-07-14
Suppress pop-up consoles on Windows.

dbukenberger:
My current task makes frequent use of the Blender-engine for boolean mesh operations.
Although Blender is launched with the `--background` flag, on Windows, a console pops-up and closes immediately for **every** call.
Since these pop-up consoles receive the OS-window-focus, you actually are not able to do anything else while the script runs in the "_background_".
I added this little piece of code (https://stackoverflow.com/a/1016651) which resolves the issue and though you might want to include it. =)

Cheers,
Dennis


mikedh:
Looks great, thanks!

PR #505, opened 2019-07-22
Vox tweak

jackd:
Stripping empty encodings was causing issues, and `Encoding.is_empty` implementations - though unused - were incorrect (though not used elsewhere). Sorry these made it this far.

voxelize_binvox change is less important, but it's a more direct way of doing things if you know the output size rather than the pitch.

### DIFF
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -2,8 +2,8 @@ try:
     from . import generic as g
 except BaseException:
     import generic as g
-from trimesh.voxel import encoding as enc
-from trimesh.voxel import runlength as rl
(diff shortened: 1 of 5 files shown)


mikedh:
No worries, looks good! By the way, I've been pretty busy this week but on my todo list is to figure out the best place to wrap [some of the `astype` calls in `voxel.encoding`](https://ci.appveyor.com/project/mikedh/trimesh/branch/master/job/wk6ynmfc4rco0bvj) which are failing in Win32 CI. If you have any ideas/cycles just something to keep in mind.

Issue #506, opened 2019-07-22
vertex_normals as average of adjacent face_normals


dbukenberger:
The default way to compute vertex normals in trimesh is as the average of the faces' normals adjacent to a vertex.
I just noticed that while testing some other stuff and the normals of a simple cube just looked so wrong.
Due to the triangulation, normals on corner-vertices are not `(1,1,1)/sqrt(3)` as one would expect.

I thought this is kind of a crude way to determine the vertex normals.
Is this done on purpose - maybe for the sake of simplicity?
If not, I have some code laying around with a more suitable weighting for the normals which I could add + tests of course. =)

mikedh:
Haha I was going to protest that we area-weighted that, but looking at the [sparse matrix](https://github.com/mikedh/trimesh/blob/465a84a0f3e3068b71dad29b232f415656c47dd5/trimesh/geometry.py#L302-L311) it's just averaging every adjacent face normal as you say. Yeah definitely just done for simplicity and speed, the sparse matrix operations are shockingly quick. 

Here's a simple triangulation of a box, looks like the weight will be `[2, 2, 1]` in the corner rather than the more reasonable `[1, 1, 1]`: 
![box](https://user-images.githubusercontent.com/1403074/61962195-00366c80-af97-11e9-9606-d69787b10e99.png)

Yeah better weighting would be great! Looking at this mesh it's not clear to me how to get the right answer... do you only average *unique* normals? That seems like it would do the trick but possibly loose the rapid nature of a sparse matrix operation. PR's super welcome!

mikedh:
Fixed by PR, thanks!

PR #512, opened 2019-07-28
 Vertex normal computation extended from uniform to corner-angle weights.

dbukenberger:
Hai Mike,

this would be my solution to issue #506.

The method is based on
> Computing Vertex Normals from Polygonal Facets
> Grit Thuerrner & Charles A. Wuethrich (1998)
> Journal of Graphics Tools, 3:1, 43-46

and uses the corner angles around a vertex to weighten the normals contribution of each adjacent face.

Obviously it is a bit slower than the simple uniformly-weighted variant.
But nevertheless, it is still possible to use the spare-matrix dot-product computation method so it is not that terrible as the looped-version.

I also extended the corresponding tests and included a reference mesh.
Sorry, for the `obj` file - but I was somehow not able to load normals from any other format. o_O

Cheers,
Dennis


mikedh:
Oh nice, looks much better! The 5000 face sample mesh was well under 1ms for both methods (weighted vs mean) in my testing, so I think defaulting to the nice new one is the way to go. 

Thanks for the PR!

PR #547, opened 2019-09-05
Topology optimization 

bbarroqueiro:
Hi,

As we have spoken last time, this PR contains the code to perform the topology optimization... I ended up using an alternative to MMA, which is very straightforward to use. The implementation loses some efficiency, but easier to read, change, build on top...  I spent some time writing docs to each function hopefully it clear enough. 

Regarding the test, the provided example with max_iter=1 can be used, if no NAN or error appear in the convergence output, the code should be running fine

Give it try and let me know...

### DIFF
--- a/trimesh/optimize.py
+++ b/trimesh/optimize.py
@@ -1,0 +1,247 @@
+import numpy as np
+import trimesh, time
+from topology import fem
+from topology import alm
+from topology import to
+from topology import meshers
+from topology import picker
+from multiprocessing import Process
+
+def topology(mesh,grid_size,E,nu,vol_fraction,radius,forces,clamp=True,preview=True,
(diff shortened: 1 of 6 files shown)


mikedh:
Thanks for the PR! I'll try to go through this next week.

mikedh:
This looks awesome! I have a few organizational thoughts about what stuff is "library level," vs what is "application level," I'm going to merge to a branch and play with it a little bit. My initial read was that the part of this the most users would find the most useful is the "FEA on voxelized geometry."  I'll make a pass and maybe we can discuss in a week or two. 

Thanks for the PR, it looks great!

bbarroqueiro:
Hey Mike

If you can find the time take a look at this: https://doi.org/10.3390/jmmp3030068 , you can a idea of whats going on and understand the scope... The scope is not really really fea... it done in a quick and dirty way but good enough for Topology Optimization...

PR #557, opened 2019-09-26
Add support for importing and exporting xyz pointclouds

PaulDebus:
Hello, 

I was working with trimesh, while I realized, that it already supports pointclouds from `ply` files, but not for example `xyz` files, which I was using. Even though loading them with numpy is very easy, I implemented pointcloud import and export, with the idea, that more formats might follow, if desired. I took a lot of "inspiration" from the import and export functionality for meshes, to make it a little easier.

As there is no official standard for the xyz format, the implementation is highly based on my specific data and needs.

I would like to discuss these changes, if they are wanted, complete, in the shape and form required and so on. I would like to also offer to take care of further changes to this code.

Thanks for the awesome software and kind regards

Paul

### DIFF
--- a/trimesh/exchange/export.py
+++ b/trimesh/exchange/export.py
@@ -11,6 +11,7 @@ from .off import _off_exporters
 from .stl import export_stl, export_stl_ascii
 from .ply import _ply_exporters
 from .dae import _collada_exporters
+from .xyz import _xyz_exporters
 
 
 def export_mesh(mesh, file_obj, file_type=None, **kwargs):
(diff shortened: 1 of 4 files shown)


mikedh:
Apologies for the delay on this and thanks for the PR! I started researching the format then got distracted haha.

PR looks good, though it would be nice to identify the major software packages that output this format and get a test file or two from each of them. 

It looks like [CloudCompare](http://www.danielgm.net/cc/) is a major one? What other packages output these files?

PaulDebus:
Haha, no problem about the response time. 

I found, that Agisoft Metashape, which we use a lot here, also produces xyz files, though they use the file extension `.txt`, which I find even less useful then the format already is. I tested importing and exporting a file produced by Metashape and the result is the following:

```
paul@work:~/repos/trimesh$ head -1 export.xyz orig.xyz 
==> export.xyz <==
665187.522077 5648349.895035 241.767693 117.0 73.0 49.0 255.0

==> orig.xyz <==
665187.522077 5648349.895035 241.767693 117 73 49
```

So there are some differences, as I exported the color as `float` instead of `int` and also wrote the alpha value for the color.

For the export from CloudCompare, the result was even better:
```
paul@work:~/repos/trimesh$ head -1 export.xyz orig.xyz 
==> export.xyz <==
665187.522077 5648349.895035 241.767693 117.0 73.0 49.0 255.0

==> orig.xyz <==
665187.522077 5648349.895035 241.767693 117.0 73.0 49.0 255.0
```

I also attached both files, so you can have a look. [xyz_files.zip](https://github.com/mikedh/trimesh/files/3680053/xyz_files.zip)



I was thinking, whether exporting `xyz` files is a useful addition, as maybe other formats are better suited for the exchange of point cloud data. On the other hand, why limit functionality there. I am not sure, what my position is there. Further, I was thinking about letting the user define some kind of format string, to define which value per line represents which thing, so something like `"x y x r g b nx ny nz"` as an additional parameter. This would increase flexibility, but I do not know if it is worth it, since reading the file with numpy and feeding the required columns into the `trimesh.PointCloud` constructor is probably easier.

What is your general position on extending trimesh to more point cloud functionality? It primarily is a mesh library, so maybe stick with what you do best... 

Cheers

mikedh:
Oh nice thanks for the example files! If we can import/export from 2 packages that seems definitely worth merging. `xyz` might not be totally consistent as you point out, but it does seem like it might be better than using `ply` or (god help us) `obj` files to store point clouds. Want to add those XYZ files to tests or should I?

For PointClouds I use them pretty much exclusively for visualization alongside meshes, but am happy to add functionality if other people find them useful. There's also [pyntcloud](https://github.com/daavoo/pyntcloud) which looks pretty good and seems philosophically aligned with trimesh. There's also [python-pcl](https://github.com/strawlab/python-pcl) which seems quite rough.

PaulDebus:
Now its my turn for taking time to answer. We had some holidays here, so I did not have to go to work ;)

I am not really firm in writing unit tests, so this would be an exercise for me, that might take some time. If you could have a final look over them, I would like to give it a try. I would also look if I can find some smaller example files. No need to download millions of points, that all behave the same.

I also think it is worth adding some pointcloud functionality, it does come up from time to time for me and having it all together in one library and viewer is nice.

mikedh:
No worries! I'll merge and add the files to a test. 

Yeah I agree having some more PointCloud functionality would be nice, especially if it's self-contained vectorized stuff. It's probably never going to be ideal for the "real time kinect/lidar analysis" type things, but for static analysis it seems like it could be pretty handy and easy to use.

PR #566, opened 2019-10-03
Update dae.py

hauptmech:
Fix bug in export of specular color


mikedh:
Oh nice, I see [`pycollada`](https://github.com/pycollada/pycollada/blob/master/collada/material.py#L421) is expecting RGBA. Thanks for the PR!

Issue #575, opened 2019-10-22
issues freezing trimesh


pauldmccarthy:
Hi @mikedh,

Unfortunately, the recent removal of `pkg_resources` (32116bb283a1e0d6bb4064f7e4ac53dbf2d2f220) has caused issues with frozen instances of trimesh (applications created with e.g. [pyinstaller](https://pyinstaller.readthedocs.io/en/stable/)). The new approach of using `inspect.getfile(inspect.currentframe())` returns the path to `trimesh` at the time that it was frozen, rather than its actual path.

The `__file__` attribute should work in both source and frozen environments, however. I'm going to follow this issue up with a PR which replaces the use of `inspect` with `__file__`.

I'm not sure of your reasons for dropping `pkg_resources` and using `inspect`, so please feel free to ignore this issue and the associated PR - I'm happy to patch trimesh in my own application if you'd prefer to keep things they way they are :)

More info on accessing resources in frozen environments using pyinstaller can be found at https://pythonhosted.org/PyInstaller/runtime-information.html

Thanks!

mikedh:
Thanks for the PR! Is there a minimal way of creating a unit test for pyinstaller? I tried:

    pyinstaller basic.py --exclude-module="`python ../../setup.py --list-all\`"

Which I though should just install `numpy` and `trimesh`, but it was doing a lot of things before hitting a recursion limit.

pauldmccarthy:
This works for me, running from the `trimesh` project root. Using this entry point:

```python
#!/usr/bin/env python

import numpy as np
import trimesh.resources as r

print('Got it:', r.get_resource('units_to_inches.json'))
```

And invoking `pyinstaller` like so (I'm adding the entire `trimesh` package as additional "data" out of laziness, rather than adding just the resource files):

```sh
(venv) ➜  trimesh git:(master) ✗ pyinstaller entrypt.py --add-data=trimesh:trimesh
```

```sh
(venv) ➜  trimesh git:(master) ✗ ./dist/entrypt/entrypt
Got it: {
  "microinches": 0.001,
  "mils": 0.001,
  "inches": 1.0,
  ...
  "gigameter": 39370078740.15748,
  "light year": 3.72461748e+17,
  "parsec": 1.21483369e+18
}
(venv) ➜  trimesh git:(master) ✗

I'm happy to add this as a test, but I'm not sure how I would go about writing a non-python test with additional dependencies.

We could solve the first issue by calling pyinstaller programmatically, e.g.:

```python
[...]

PR #576, opened 2019-10-22
RF: Use __file__ instead of inspect to find resources dir

pauldmccarthy:
This PR addresses issue #575 - using `inspect` to identify the location of a python module will return the incorrect path when the code has been "frozen" (bundled into a standalone executable with e.g. [pyinstaller](https://pyinstaller.readthedocs.io/en/stable/)).

This PR proposes that the module-global `__file__` attribute is used instead - this does work in both frozen and conventional (i.e. source) execution environments.

Feel free to ignore if you have good reasons for using `inspect`!

### DIFF
--- a/trimesh/resources/__init__.py
+++ b/trimesh/resources/__init__.py
@@ -1,11 +1,8 @@
 import os
-import inspect
 
-# find the current absolute path using inspect
-_pwd = os.path.dirname(
-    os.path.abspath(
-        inspect.getfile(
-            inspect.currentframe())))
+
+# find the current absolute path to this directory
+_pwd = os.path.dirname(__file__)
 
 
 def get_resource(name, decode=True):


mikedh:
Thanks for the PR! `__file__` looks great and seems to be passing CI, I have no idea why it was using inspect if `__file__` is an option. Maybe we'll find out when we break someone else's workflow :smile:. I did google a bit and it looks like `__file__` is available in all of the supported versions of python and I couldn't find the downside.

I recently removed `pkg_resources`, as it is IIRC installed by setuptools, which is required by pip so pretty much everyone has it, except for the conda-forge Azure CI for some reason.

This also raises another point, I see a number of places where resources are retrieved on import, which should really be moved inside functions, perhaps with caching for subsequent calls. I'll try to remove those wherever possible.

PR #579, opened 2019-10-24
Clarify comment for slice_plane

panangam:
Copied over the comment from the original function in intersections.py which is more informative, since the slice_plane method is just a wrapper around that function.


mikedh:
Thanks for the PR!

PR #588, opened 2019-11-02
Preserves the alphaMode when reading and writing gltf

FreakTheMighty:
I found that in the course of reading and writing gltf files, transparent objects became opaque. This was caused by the alphaMode being unassigned in the PBRMaterial and not included in the gltf export.

I've also added a missing `matallicFactor` property.


mikedh:
Awesome thanks! Just merged to master.

PR #605, opened 2019-11-07
`collections.abc` compatibility for Python 3.8+

paulmelnikow:
Suppresses warnings like these:

```
trimesh/util.py:2061
  /Users/pnm/code/trimesh/trimesh/util.py:2061: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    class FunctionRegistry(collections.Mapping):

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

### DIFF
--- a/trimesh/caching.py
+++ b/trimesh/caching.py
@@ -15,6 +15,11 @@ import hashlib
 
 from functools import wraps
 
+try:
+    from collections.abc import Mapping
+except ImportError:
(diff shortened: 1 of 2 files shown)


paulmelnikow:
I think this build failure is unrelated to the change.

PR #606, opened 2019-11-07
Modest optimization in `nearby_faces` and `nearest`

paulmelnikow:
I added a flag to `nearby_faces()` to return conveniently stacked results.

The rest of this is pretty modest, removing loops and some unnecessary initialilzation and allocation.

### DIFF
--- a/trimesh/triangles.py
+++ b/trimesh/triangles.py
@@ -567,13 +567,13 @@ def closest_point(triangles, points):
         raise ValueError('need same number of triangles and points!')
 
     # store the location of the closest point
-    result = np.zeros_like(points)
+    result = np.empty(points.shape, dtype=np.float64)
     # which points still need to be handled
     remain = np.ones(len(points), dtype=np.bool)
 
     # if we dot product this against a (n, 3)
     # it is equivalent but faster than array.sum(axis=1)
-    ones = [1.0, 1.0, 1.0]
+    ones = np.ones(3)
 
     # get the three points of each triangle
     # use the same notation as RTCD to avoid confusion
--- a/trimesh/proximity.py
+++ b/trimesh/proximity.py
@@ -15,7 +15,7 @@ from .triangles import closest_point as closest_point_corresponding
 from collections import deque
 
 
-def nearby_faces(mesh, points):
+def nearby_faces(mesh, points, ret_stacked=False):
     """
     For each point find nearby faces relatively quickly.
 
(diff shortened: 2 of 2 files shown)


trimesh/triangles.py
```diff
@@ -567,13 +567,13 @@ def closest_point(triangles, points):
         raise ValueError('need same number of triangles and points!')
 
     # store the location of the closest point
-    result = np.zeros_like(points)
+    result = np.empty(points.shape, dtype=np.float64)
```
paulmelnikow: The assignment of `result[remain]` at the end ensures this gets filled.

mikedh:
Thanks for the PR! 

I haven't gotten a chance to test/profile the `closest_point_naive` improvements which look good, but re:`ret_stacked`:
Usually list comprehensions are slightly faster than explicit loops for what I assume are scoping or magic-related reasons. When I tested some timings the PR'd method was almost identical speed-wise to the current method (with 5000 triangles and 5000 points, 2.66s w/ list comprehension / 2.65s as-PR'd, not including the concatenate call for apples-apples). Is there a situation where this method is faster? If stacked results are needed I think composing the results outside the candidates function is probably the way to go, like:

```
In [6]: %timeit np.concatenate(candidates)
167 ms ± 795 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit [len(f) for f in candidates]
350 µs ± 4.66 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Both of which are pretty snappy, and it avoids a boolean kwarg / code path.

paulmelnikow:
Hmm, let me fiddle with that listcomp a bit. There are two things which are needed: the concatenation, and a second array containing the lengths. I'm not sure of a performant way to concatenate and produce lengths at the same time, which is what brought me to use a loop. Without it, ether there need to be two listcomps, or one listcomp and one concatenate. Does that make sense? It might still be faster.

The query I was testing was something like 20 query points and 25k triangles. It seemed to yield a consistent 2% speedup, though the code in my current use case is fairly hard to isolate. Maybe I can extract some things so I can check it more carefully.

It's unfortunate that rtree doesn't have a way to run many queries at once. I think that would speed this up the most, as the result could be stacked in C++.

mikedh:
> Without it, ether there need to be two listcomps, or one listcomp and one concatenate. Does that make sense? It might still be faster.

Yeah I think 2 listcomps are almost identical speed-wise, the `[len(f) for f in candidates]` listcomp executes in 350us, or .3ms

> The query I was testing was something like 20 query points and 25k triangles. It seemed to yield a consistent 2% speedup, though the code in my current use case is fairly hard to isolate. Maybe I can extract some things so I can check it more carefully.

Gotcha, my instinct is that single-digit-percentage gains are going to go away with varying workloads here.

> It's unfortunate that rtree doesn't have a way to run many queries at once. I think that would speed this up the most, as the result could be stacked in C++.

Yeah totally, I imagine a cythonized replacement module or more specific methods for an rtree could get something crazy, like 10-100x speedups for some of these queries.

PR #662, opened 2019-12-11
Apply `metallicFactor` property to `pbrMetallicRoughness` object on glTF export

thagberg:
When appending materials to the export tree for glTF/glb models, the `metallicFactor` property was being kept at the root of the material object instead of being moved down into the `pbrMetallicRoughness` object.

This is how the material was defined in the source glTF:
```
    {
      "pbrMetallicRoughness": {
        "baseColorTexture": {
          "index": 1
        },
        "metallicFactor": 0,
        "roughnessFactor": 0.5
      }
```
And this is how the material was defined when exporting from trimesh:
```
    {
      "metallicFactor": 0.110277727,
      "alphaMode": "OPAQUE",
      "pbrMetallicRoughness": {
        "baseColorTexture": {
          "index": 0
        }
      }
```
Finally, after this fix is applied and the model is re-exported:
```
    {
      "alphaMode": "OPAQUE",
      "pbrMetallicRoughness": {
        "baseColorTexture": {
          "index": 0
        },
        "metallicFactor": 0.110277727
      }
    }
```

This results in the material correctly being rendered in common glTF viewers.


mikedh:
Yikes, that concatenation when you forget the commas 😅 . Thanks for the fix!

PR #663, opened 2019-12-12
Fix trimesh.voxel.ops.matrix_to_marching_cubes with pitch != 1

wkentaro:
The generated mesh can be translated wrongly.


mikedh:
Looks great, thanks!

PR #676, opened 2019-12-24
Fix Scene.add_geometry with list of geometries

wkentaro:
Adding the same `node_name` overwrites graph.

### DIFF
--- a/tests/test_scene.py
+++ b/tests/test_scene.py
@@ -161,6 +161,16 @@ class SceneTests(g.unittest.TestCase):
         assert scene.md5() == md5
         assert converted.md5() != md5
 
+    def test_add_geometry(self):
+        # list-typed geometry should create multiple nodes,
+        # e.g., below code is equivalent to
+        #     scene.add_geometry(geometry[0], node_name='voxels')
+        #     scene.add_geometry(geometry[1], parent_node_name='voxels')
+        scene = g.trimesh.Scene()
+        geometry = [g.trimesh.creation.box(), g.trimesh.creation.box()]
+        scene.add_geometry(geometry, node_name='voxels')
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! Would you mind adding a unit test? I'm not totally sure which case this is handling, when I check a toy example it looks OK:
```
In [1]: import trimesh

In [2]: s = trimesh.Scene([trimesh.creation.icosphere(), trimesh.creation.box()])

In [3]: s
Out[3]: <trimesh.Scene(len(geometry)=2)>

In [4]: s.graph.nodes_geometry
Out[4]: array(['geometry_0_4AOEYQQQPETW', 'geometry_1_4QRLPHDQ9YXI'], dtype='<U23')

In [5]: s.graph.nodes
Out[5]: 
array(['world', 'geometry_0_4AOEYQQQPETW', 'geometry_1_4QRLPHDQ9YXI'],
      dtype='<U23')

```

tests/test_scene.py
```diff
@@ -161,6 +161,16 @@ def test_scaling(self):
...
 
+    def test_add_geometry(self):
+        # list-typed geometry should create multiple nodes,
+        # e.g., below code is equivalent to
+        #     scene.add_geometry(geometry[0], node_name='voxels')
+        #     scene.add_geometry(geometry[1], parent_node_name='voxels')
+        scene = g.trimesh.Scene()
+        geometry = [g.trimesh.creation.box(), g.trimesh.creation.box()]
+        scene.add_geometry(geometry, node_name='voxels')
+        assert len(scene.graph.nodes_geometry) == 2
```
wkentaro: This part was originally `1`, which causes unexpected behaviors.

mikedh:

PR #677, opened 2019-12-24
Do not merge vertices with different normals

gustavla:
I've been having issues with vertex normals not loaded correctly. I traced the issue to the automatic vertex merger, which does not seem to take vertex normals into account.

Here is an example before this fix:

Screen Shot 2019-12-24 at 12 13 33 PM

Here is an example after this fix (or with merging commented out):

Screen Shot 2019-12-24 at 12 13 59 PM

### DIFF
--- a/trimesh/grouping.py
+++ b/trimesh/grouping.py
@@ -56,6 +56,7 @@ def merge_vertices(mesh,
         # converted to integers at requested precision
         stacked = np.column_stack((
             mesh.vertices * (10 ** digits),
+            mesh.vertex_normals * (10 ** digits),
             mesh.visual.uv * (10 ** uv_digits))).round().astype(np.int64)
         # Merge vertices with identical positions and UVs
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! Yeah not merging vertices with different normals makes sense, and seems like it gets us closer to a more widely acceptable "merge vertices by default" solution.

The logic here looks good, though I think the `merge_vertex` logic is old code that needs a wider refactor, which I took a pass at here: https://github.com/mikedh/trimesh/commit/27f718dc0390ea789eaf589bb814d755546fb113

The reason I pulled from cache rather than use the property is because in the as-PR'd logic it would autogenerate vertex normals if they didn't exist rather than only using them if they were already calculated or set.

If that example model is small-ish would you mind PR'ing a test with that example model just to make sure I didn't screw the logic up? If not no worries, thanks for the PR!

PR #701, opened 2020-01-27
[dae] Fix unicode for Python 2

iory:
The current dae loader failed to load some `collada` file which contains `utf-8` geometry tag on Python2.
Example `collada` file is following.
https://drive.google.com/open?id=1BBFboFA-WSKoVlf1JihBZ85tFEQmU6IO

### DIFF
--- a/trimesh/exchange/dae.py
+++ b/trimesh/exchange/dae.py
@@ -219,7 +219,7 @@ def _parse_node(node,
                     vis = visual.texture.TextureVisuals(
                         uv=uv, material=material)
 
-                primid = '{}.{}'.format(geometry.id, i)
+                primid = '{}.{}'.format(geometry.id.encode('utf-8'), i)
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR and example file! I think it might be slightly easier to make the format strings unicode  https://github.com/mikedh/trimesh/commit/d96faf1c8d2eb967df28e786f6f5b337b0586419 which also appears to fix this for your example file.

iory:
That's great!
The commit(https://github.com/mikedh/trimesh/commit/d96faf1c8d2eb967df28e786f6f5b337b0586419) solves the problem.
Could you merge the commit to master?

mikedh:
Great! Just made an alternative PR, I'll merge as soon as CI finishes. Thanks for the PR!

iory:
Thank you!
I close this PR.

Issue #704, opened 2020-01-29
Delete registered geometry from SceneViewer


iory:
I'm visualizing objects by `SceneViewer`, this library.
The class has `add_geometry` and there is no `delete_geometry` function.
Is there are a way to delete objects from `Scene`?

mikedh:
Hey, not currently. You'd have to pop the geometry (`scene.geometry.pop(geometry_key)`) and then clean up the graph (maybe something like `scene.graph.transforms.remove_node(node)`). 

Happy to take PR's!

iory:
Thank you!

I created the PR related to this topic.

mikedh:
Thanks for the PR! Merged and released.

PR #707, opened 2020-02-05
Add delete_geometry and hide/show geometry to SceneViewer

iory:
Related to #704 

Add a function `delete_geometry` to `SceneViewer`.
By the way, `scene.graph.node_geometry` is the cached property.
What is the best way to update the cached property?

### DIFF
--- a/trimesh/scene/scene.py
+++ b/trimesh/scene/scene.py
@@ -198,6 +198,22 @@ class Scene(Geometry):
                           geometry_flags={'visible': True})
         return node_name
 
+    def delete_geometry(self, name):
+        """
+        Delete a geometry from the scene.
+
+        Parameters
+        --------------
+        name : hashable
+          Name that references geometry
+        """
+        self.graph.transforms.remove_node(name)
+        self.graph._cache.update(
+            {'nodes_geometry': np.array(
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome, thanks for the PR! I would actually in the viewer probably try to implement in a way that doesn't mutate the scene, and then (if still necessary) add a `Scene.delete_geometry` function that does mutate the scene. 

```
def hide_geometry(self, node):
    """"
    Don't display the geometry contained at a node.

    Parameters
    -------------
    node : str
       Node to not display
    """"
    self._nodes_hidden.add(node)
```

Then in the `on_draw` function you could do a quick check `if node in self._nodes_hidden: continue`.

iory:
Thanks for the review!
OK. I add `delete_geometry` function to  `Scene` class and removed `Scene`'s `delete_geometry`.
Also, I add a `sweap_geometries` function to delete key value of `vertex_list`, `vertex_list_hash`, `vertex_list_mode` and `textures`.

In addition to it, `hide_geometry` and `show_geometry` to `Scene` class.

mikedh:
Thanks for the PR! I renamed some of the methods, and moved some functionality into the `scene.graph`. I'll release shortly.

PR #708, opened 2020-02-05
Fix for initial camera aspect ration in Three.js view

hoosha:
Quick hack to fix the camera aspect ratio problem not being inited correctly which is fixed when resizing the browser window.


mikedh:
Awesome thanks for the fix! I'll rebuild the JS bundle with your change.

PR #712, opened 2020-02-07
add centroid property to Path3D class

panangam:
Copied the centroid cached method from Path2D to Path3D because why not?

### DIFF
--- a/trimesh/path/path.py
+++ b/trimesh/path/path.py
@@ -1017,6 +1017,19 @@ class Path3D(Path):
         if show:
             plt.show()
 
+    @caching.cache_decorator
+    def centroid(self):
+        """
+        Return the centroid of the path object.
+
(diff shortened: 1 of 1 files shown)


panangam:
I suppose the centroid method can live in Path base class, though I held off as I don't fully understand the classes hierarchy yet.

mikedh:
Looks good! I moved it up to the `Path` object, and also changed it to be the centroid of the AABB. I'll release shortly.

Thanks for the PR!

panangam:
@mikedh 
Oh, I think that might lead to unexpected behaviors for many people. Say you have a path of a triangle, the centroid is defined as the average of the 3 vertices. Centroid can also mean the center of mass, which again is the average of the points. 
If the center of AABB will be useful, I think that should have a different name.

PR #718, opened 2020-02-14
Ignore RuntimeWarning about invalid values

wkentaro:

PR #719, opened 2020-02-16
Fix triangles.angles divide by zero

mikedh:
In our angles function previously we were dividing by zero then using `np.nan_to_number` to convert back to zero, to avoid slow checks which I never really liked. This PR uses `util.unitize` which does zero-checks, and should no longer ever divide by zero and is mildly faster (350ms vs 380ms on random triangles).

Alternative PR to #718


mikedh:
Fixes #722 #718

PR #724, opened 2020-02-18
Support differing list-lengths in ASCII-PLYs

fegorsch:
PLY allows defining list-properties, where every instance of the
property can have a different length. This can for instance be useful
when storing submeshes in the metadata of a PLY-file.

trimesh only supported PLY-files with list-properties, if every
instance of the list had the same length. Properties with lists of
differing lengths let the parser crash at the `vstack`-call in
`ply.ply_ascii`, as arrays of differing size cannot be stacked.

The implementation at hand adds support for lists of differing length.

The drawback of the new implementation is an increased load-time for
elements containing lists. I tried to mitigate this effect by checking
whether an element contains a list and using the old, numpy-based
algorithm to load the data, if none is found.

### DIFF
--- a/models/points_ascii_with_lists.ply
+++ b/models/points_ascii_with_lists.ply
@@ -1,0 +1,18 @@
+ply
+format ascii 1.0
+element vertex 5
+property float x
+property float y
+property float z
+element point_list 2
+property list uchar uint point_indices1
+property list uchar uint point_indices2
+property float some_float
+end_header
+-0.06325 0.0359793 0.0420873 
+-0.06275 0.0360343 0.0425949 
+-0.0645 0.0365101 0.0404362 
+-0.064 0.0366195 0.0414512 
+-0.0635 0.0367289 0.0424662 
+3 10 11 12 2 13 14 1.1
+2 10 11 3 12 13 14 2.2
--- a/tests/test_ply.py
+++ b/tests/test_ply.py
@@ -57,6 +57,27 @@ class PlyTest(g.unittest.TestCase):
         assert m.vertices.shape == (1024, 3)
         assert isinstance(m, g.trimesh.PointCloud)
 
+    def test_list_properties(self):
+        """
+        Test reading point clouds with the following metadata:
+        - lists of differing length
+        - multiple list properties
+        - single-element properties that come after list properties
+        """
+        m = g.get_mesh('points_ascii_with_lists.ply')
+
+        point_list = m.metadata['ply_raw']['point_list']['data']
+        assert g.np.array_equal(
+            point_list['point_indices1'][0], g.np.array([10, 11, 12], dtype=g.np.uint32))
+        assert g.np.array_equal(
+            point_list['point_indices1'][1], g.np.array([10, 11], dtype=g.np.uint32))
+        assert g.np.array_equal(
+            point_list['point_indices2'][0], g.np.array([13, 14], dtype=g.np.uint32))
+        assert g.np.array_equal(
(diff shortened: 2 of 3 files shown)


mikedh:
Awesome thanks for the PR! Though couldn't we fall back to looping only after the `np.vstack` call fails? It seems like that would allow us in the case where every list is the same length we get the same speed we do now, but also handle the variable-length case?

fegorsch:
You're welcome :)

There's a corner-case where stacking with list-properties is problematic: there are multiple list properties containing lists with differing lengths, but the sum of elements is the same for each row. E.g.:

```
element point_list 2
property list uchar uint point_indices1
property list uchar uint point_indices2
[...]
3 10 11 12 2 13 14  # 3-element list, followed by a 2-element list
2 10 11 3 12 13 14  # 2-element list followed by a 3-element list
```

Stacking would work here, but the resulting data would be wrong. I think there's no way of handling this without looping through the rows.

I think a good compromise would be to only do stacking with lists, if there's just a single list-property in the element. This is probably the most common case (i.e. triangle-lists).

This gives us two cases:
- 0 or 1 list and stacking works => stack with knowing that the single list is always the same size
- more than one list or stacking did not work => loop and check every row for list lengths

I'll push a proposal shortly.

mikedh:
Great thanks for the update! When I was profiling the PR'd solution by loading every PLY in `/models` with this script:

```
import os
import trimesh
import numpy as np

if __name__ == '__main__':
    trimesh.util.attach_to_log()
    

    path = './models'
    # get every PLY file in our test set which 
    names = [os.path.join(path, f)
             for f in os.listdir(path)
             if f.lower().endswith('ply')]
    
    # a sampling profiler, get with `pip install pyinstrument`
    from pyinstrument import Profiler
    profiler = Profiler()
    profiler.start()

    # profile only the load
    m = [trimesh.load(f)
         for f in names]

    profiler.stop()
    print(profiler.output_text(unicode=True, color=True))
```

It looks like it was running into some issues (which is also why CI was failing), traceback:
```
/media/psf/Dropbox/robotics/trimesh/trimesh/exchange/ply.py in ply_ascii(elements, file_obj)
    589             # all rows have the same length and we only have at most one list property where all
    590             # entries have the same length. this means we can use the quick numpy-based loading.
--> 591             element_data = load_element_with_single_length_lists(values['properties'], data)
    592         else:
    593             # there are lists of differing lengths. we need to fall back to loading the data by

/media/psf/Dropbox/robotics/trimesh/trimesh/exchange/ply.py in load_element_with_single_length_lists(properties, data)
    538         start = end
    539 
--> 540     return {n: data[:, c[0]:c[1]].astype(dt) for c, (n, dt) in zip(col_ranges, properties.items())}
    541 
    542 

/media/psf/Dropbox/robotics/trimesh/trimesh/exchange/ply.py in <dictcomp>(.0)
    538         start = end
    539 
--> 540     return {n: data[:, c[0]:c[1]].astype(dt) for c, (n, dt) in zip(col_ranges, properties.items())}
    541 
    542 

~/conda/lib/python3.7/site-packages/numpy/core/_internal.py in _commastring(astr)
    177                     raise ValueError(
    178                         'format number %d of "%s" is not recognized' %
--> 179                         (len(result)+1, astr))
    180                 startindex = mo.end()
    181 

ValueError: format number 2 of "b'<u1, ($LIST,)<i4'" is not recognized

```

fegorsch:
Sorry for the delay. I finally got to fixing the bug and the formatting issues.

mikedh:
Looks great, thanks for fixing the tests! I like the refactor into 2 functions, seems like a nice improvement.

Thanks for the PR!

PR #725, opened 2020-02-19
add meshio loader

nschloe:
This PR adds a loader for many more mesh formats via [meshio](https://github.com/nschloe/meshio). trimesh can now
```python
import trimesh

# VTU, XDMF, Gmsh, Exodus,...
mesh = trimesh.load('out.vtk')
```
I have _not_ added an exporter because trimesh expects the function to return a string which, I suppose, is then written to a file. meshio does not support that since strings can take up huge amounts of memory. (meshio typically handles very large meshes.)

I have also not added meshio to the list of dependencies though this could be done. meshio supports Python 3.5+.


mikedh:
Looks great thanks! And thanks for pulling back my changes into the PR. My only real put is I moved the `trimesh.exchange.meshio` into `misc` to avoid the ambiguity of an imported module name matching one of our local imports. I think with that flake8 change this should be good to go.

nschloe:
As expected, tests fail for Python 2.7 and 3.4 which meshio doesn't support.

mikedh:
Gotcha, how about we move it `requirements_all`, vs `requirements_easy` then?

nschloe:
Done. Note that some of the `requirements_easy` packages don't support Python 2.7 or 3.4 anymore either, e.g., https://pypi.org/project/networkx/ and https://pypi.org/project/Pillow/. Others (https://github.com/sympy/sympy/issues/17943) will follow soon.

mikedh:
Awesome thanks for the PR! Yeah the older stuff is getting a little shaky for sure. I just checked stats, it looks like we still get a decent chunk of 2.7 downloads, but basically no Python 3.4 downloads.

nschloe:
>  I just checked stats, it looks like we still get a decent chunk of 2.7 downloads

[The same is true for numpy](https://pypistats.org/packages/numpy), but still they dropped support for 2.7. 1.16.* is their last release with with Python 2 support. (The current version is 1.18.1.) Py2 users simply get older versions.

In my experience, a good strategy is to support those versions which are supported by Python official (right now that's 3.5-3.8), but of course that's all up to you.

PR #745, opened 2020-03-19
Add cone mesh

iory:
I added a cone mesh creation.
The following gif is the result of `cone` creation.
cone


mikedh:

PR #748, opened 2020-03-21
Fix bug related to specifying the center of circle paths

dmontagu:
It seems to me the circle center is currently getting added twice.

Let me know if you'd like to keep the original test and just add a separate one (that is almost the same, just with center specified) for the offset circle.


mikedh:

PR #538, opened 2019-08-23
Support subdividing attributes.

keunhong:
Adds support for subdividing other attributes such as uvs.

e.g.,

```python
def subdivide_mesh(mesh):
    attributes = {}
    if hasattr(mesh.visual, 'uv'):
        attributes = {'uv': mesh.visual.uv}
    vertices, faces, attributes = trimesh.remesh.subdivide(
        mesh.vertices, mesh.faces, attributes=attributes)
    mesh.vertices = vertices
    mesh.faces = faces
    if 'uv' in attributes:
        mesh.visual.uv = attributes['uv']

    return mesh
```

PR #763, opened 2020-03-31
Add ignore_mtl and default_material flags for OBJ.

keunhong:
These flags were added in https://github.com/mikedh/trimesh/pull/538 but were removed in https://github.com/mikedh/trimesh/commit/b45b74dc3aaecd8833c789e41d77650b2ddbe0be#diff-24e7992eca78afee5fd5565cf0bf76e9

It was my mistake for not splitting the PR. This PR re-introduces these flags

 - `ignore_mtl`: ignore the existence of MTL files and do not throw an error if it doesn't exist
 - `default_material`: overrides the materials to one set in the `kwargs`. This is particularly useful if you don't want to load the materials but still need the UV mappings.

### DIFF
--- a/trimesh/exchange/obj.py
+++ b/trimesh/exchange/obj.py
@@ -61,7 +61,7 @@ def load_obj(file_obj,
     # Load Materials
     materials = {}
     mtl_position = text.find('mtllib')
-    if mtl_position >= 0:
+    if mtl_position >= 0 and not kwargs.get('ignore_mtl', False)::
         # take the line of the material file after `mtllib`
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR, and no worries! Would you mind adding a test which passes each flag and verifies the loader is doing the right thing?

Thanks!

mikedh:
Hey, I think this may have been superceded by #813 which stores an empty material if UV's were stored?

keunhong:
This does indeed seem to solve the issue (much more elegantly).

Thanks!

PR #776, opened 2020-04-10
Enable material transparency in collada importer

mmatl:
Hi Mike! Hope you've been doing well. This is just a quick, minor update that reads the "transparent" DAE attribute and uses it to set the base color's alpha channel if one is present. I tested this on meshes exported from Blender and it worked well. Let me know if you have any questions.

Best,
Matt


mikedh:
Hey Matt, likewise! Thanks for the PR, looks great!

PR #779, opened 2020-04-12
Removing conda from the Dockerfile.

psigen:
This PR includes a number of cleanups that reduce the complexity of the overall build process (smaller, lighter fetches) and strip all conda usage in favor of depending on python:3.8-slim-buster.


mikedh:
Looks great! I'm super on board with switching the base image to official Python, and the script cleanup is also appreciated. Most dependencies added wheels in the last year or two, which made miniconda in the image much less necessary. The only thing actually installed by conda at this point was `pyembree`, thanks for creating a build for that!

I ran through a few tests and everything looked good, I'll merge shortly.

@nick-parker you're the only other docker image user that I know of, are you using conda in your images for anything?

nick-parker:
Nope, all good to remove it! Thanks for checking

Issue #731, opened 2020-03-05
trimesh.intersections .slice_mesh_plane returns non watertight mesh


amitoved:
hello, 
under the  **slice_mesh_plane** documentation, exists a **cap** parameter which is supposed to seal the opening created by the mesh slicing but this parameter isn't used by the function or by **slice_faces_plane** which accepts **kwargs by  **slice_mesh_plane**.  
Has this capability been removed?
Is there another function which does that separately?
Thank you.

mikedh:
Hey, looks like I added the kwarg but never implemented it. Happy to take PR's to fill it in, or I can remove the errant kwarg on next release.

Thanks for the report!

mjd3:
@mikedh I think we got pretty close to this in #235, maybe that code was just never pushed? If there's still some work to be done, I might have time to submit a PR for this as well.

mikedh:
Ah yeah, looks like `cap` was implemented in the comments but never merged hahah. If you have cycles a PR would be awesome thanks!

mjd3:
@amitoved This should be addressed by #800 which got merged in about a week ago!

donno:
Using 3.8.18, I am experiencing `slice_mesh_plane()` failing to return a watertight mesh.

I was alerted to the problem as I was cutting by two planes. The capping algorithm checks after the first cap is made from the first slice that the resulting mesh is watertight before it attempts to apply the cap on the new side and that check failed.

Issue #235, opened 2018-10-20
Split mesh using a splitting plane


themmes:
I want to split a mesh using a plane, similar to `cross_section` but returning both halves of the mesh as closed meshes instead of the polyline boundary of the intersection surface. I wrote the following code to achieve this:

```
def split_mesh(mesh, plane_origin, plane_normal):
    """
    Splits mesh by intersection with a plane, 
    returns closed mesh for positive and negative side of plane 
    """
    plane_normal = unit_vector(plane_normal)
    
    polyline, triangles_edge = trimesh.intersections.mesh_plane(bbox_mesh, 
                                                           plane_origin= plane_origin, 
                                                           plane_normal= plane_normal, 
                                                           return_faces=True)
    
    vertices_negative = []
    vertices_positive = []
    for vertex_id, vertex in enumerate(bbox_mesh.vertices):
        # TODO: add threshold
        signed_distance = np.dot(plane_normal, np.subtract(vertex, plane_origin))
        if signed_distance < 0:
            vertices_negative.append(vertex_id)
        else:
            vertices_positive.append(vertex_id)
    
    triangles_negative = []
    triangles_positive = []
    # For each triangle check how many points on either side of the plane
    for triangle in bbox_mesh.faces:
        triangle_negative_check = [vertex_id in vertices_negative for vertex_id in triangle]
        # If all points on the negative side this triangle is not affected
        if all(triangle_negative_check):
            triangles_negative.append(np.array(bbox_mesh.vertices[triangle]))
        elif any(triangle_negative_check):
            # Get section line for this triangle on edge
            section = polyline[np.all(np.isin(bbox_mesh.faces[triangles_edge], np.array(triangle)), axis=1)][0]
            assert section.shape == (2,3)
            # If two points on negative side
            if sum(triangle_negative_check) == 2:
                vertex_1, vertex_2 = bbox_mesh.vertices[triangle[triangle_negative_check]]
                triangles_negative.extend(unordered_to_triangles([section[0], section[1], vertex_1, vertex_2]))
                vertex_3 = bbox_mesh.vertices[triangle[np.logical_not(triangle_negative_check)]]
                triangles_positive.append(np.vstack([section[0], section[1], vertex_3]))
            elif sum(triangle_negative_check) == 1:
                vertex_3 = bbox_mesh.vertices[triangle[triangle_negative_check]]
                triangles_negative.append(np.vstack([section[0], section[1], vertex_3]))
[...]

PR #788, opened 2020-04-13
trimesh.intersections .slice_mesh_plane returns non watertight mesh

MikeAS-Coder:
Hi, the fix found its way in the comments but never merged. Would be great if you could take a look.
[#235](https://github.com/mikedh/trimesh/issues/235) and [#731](https://github.com/mikedh/trimesh/issues/731). 
Thank you.


mikedh:
Ah gotcha. But what's up with the PR, is it related to that issue, or were you trying to update that branch?

PR #794, opened 2020-04-14
Add Box.as_box_outline()

wkentaro:


mikedh:
Thanks for the PR! I renamed it to `Box.as_outline`, and moved the path import into the function as `trimesh.path` may raise an `ImportError` on minimal installs. I'll release shortly.

PR #795, opened 2020-04-14
Fix VoxelGrid.bounds for half voxel shift

wkentaro:
```python
#!/usr/bin/env python

import trimesh
import trimesh.transformations as ttf

import morefusion


models = morefusion.datasets.YCBVideoModels()
voxel = models.get_solid_voxel_grid(class_id=2)

scene = trimesh.Scene()
scene.add_geometry(voxel.as_boxes())
if 0:
    scene.add_geometry(
        trimesh.path.creation.box_outline(voxel.extents),
        transform=ttf.translation_matrix(voxel.bounds.mean(axis=0)),
    )
else:
    scene.add_geometry(voxel.bounding_box.as_box_outline())
scene.show(resolution=(640, 480))
```

**Before**
Screenshot from 2020-04-15 00-45-54

**After**
Screenshot from 2020-04-15 00-46-57

### DIFF
--- a/trimesh/voxel/base.py
+++ b/trimesh/voxel/base.py
@@ -11,6 +11,7 @@ from . import transforms
 from . import morphology
 from . import encoding as enc
 
+from .. import bounds as bounds_module
 from .. import caching
 from .. import transformations as tr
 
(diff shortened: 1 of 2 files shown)


trimesh/voxel/base.py
```diff
@@ -141,8 +141,10 @@ def strip(self):
 
     @caching.cache_decorator
     def bounds(self):
-        points = self.points
-        bounds = np.array([points.min(axis=0), points.max(axis=0)])
+        indices = self.sparse_indices
+        bounds = np.array([indices.min(axis=0) - 0.5,
+                           indices.max(axis=0) + 0.5])
+        bounds = self._transform.transform_points(bounds)
```
mikedh, line 147: I like the sparse logic for getting bounds! I think however you need to transform all 8 corners of the AABB then take the min-max of that *after* transform as min-max may switch around. Suggestion:
```
from .. import bounds as bounds_module

def bounds(self):
    indices = self.sparse_indices
    # get all 8 corners of the AABB
    corners = bounds_module.corners([indices.min(axis=0) - 0.5, indices.max(axis=0) + 0.5])
    # transform these corners to a new frame
    corners = self._transform.transform_points(corners)
    # get the AABB of corners in-frame
    bounds = np.array([corners.min(axis=0), corners.max(axis=0)])
```

wkentaro: Fixed!

mikedh: Awesome thanks!

PR #796, opened 2020-04-15
Added option to merge buffers

FreakTheMighty:
I've added an option to `merge_buffers` when exporting to gltf. This option combines all buffers into one, similar to the glb export. The `test_gltf` export produces over a 100 binary files. Merging those buffers into one can be useful when you want to get the separate json header, but would like to reduce the number of requests required to download all data for the model.


mikedh:

PR #800, opened 2020-04-17
Add cap argument to slice_plane

mjd3:
Adds the cap argument to allow for watertight meshes after slicing. Some minor other updates for consistency (replaced try/except with unique_bincount, allow section to take cached dots). 

There is probably a faster way to implement this cap arg; open to suggestions! Right now, it takes the slice_plane function from  ~2 ms on my laptop to ~85 ms when adding a cap. From some basic profiling, it seems like this could be significantly faster if there was a good way to triangulate without the sectioning/polygon conversion. The sectioning can already be done in slice_faces_plane (we can easily get the edges that make up the 3D path), but I couldn't think of a good way to replace the polygon creation. 

```python
In [1]: import trimesh                                               
In [2]: m = trimesh.load('models/featuretype.STL')               
In [3]: origin = [m.center_mass, m.center_mass]
In [4]: normal = [trimesh.unitize([1,1,1]), trimesh.unitize([1,2,3])]
In [5]: %timeit sliced = m.slice_plane(plane_origin=origin, plane_normal=normal, cap=True)                                        
85.5 ms ± 477 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
In [6]: %timeit sliced = m.slice_plane(plane_origin=origin, plane_normal=normal)                                                  
2.35 ms ± 29.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```python
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
681         6         25.0      4.2      0.0              path = sliced_mesh.section(plane_normal=normal,
682         6         15.0      2.5      0.0                                         plane_origin=origin,
683         6      44398.0   7399.7     20.5                                         cached_dots=dots)
684                                                       
685                                                       # transform Path3D onto XY plane for triangulation
686         6      10778.0   1796.3      5.0              on_plane, to_3D = path.to_planar()
687                                           
688                                                       # triangulate each closed region of 2D cap
689                                                       # without adding any new vertices
690         6         29.0      4.8      0.0              v, f = [], []
691        12      61019.0   5084.9     28.2              for polygon in on_plane.polygons_full:
692         6         31.0      5.2      0.0                  t = triangulate_polygon(polygon, 
693         6         18.0      3.0      0.0                                          triangle_args='p', 
694         6      30608.0   5101.3     14.1                                          allow_boundary_steiner=False)
695         6         37.0      6.2      0.0                  v.append(t[0])
696         6         25.0      4.2      0.0                  f.append(t[1])
```

### DIFF
--- a/tests/test_section.py
+++ b/tests/test_section.py
@@ -140,7 +140,7 @@ class SliceTest(g.unittest.TestCase):
         mesh = g.trimesh.creation.box()
 
         # Cut corner off of box and make sure the bounds and number of faces is correct
-        # (Tests new triangles, but not new quads or triangles contained entirely)
+        # Tests new triangles, but not new quads or triangles contained entirely
         plane_origin = mesh.bounds[1] - 0.05
         plane_normal = mesh.bounds[1]
 
@@ -150,6 +150,16 @@ class SliceTest(g.unittest.TestCase):
         assert g.np.isclose(sliced.bounds[0], mesh.bounds[1] - 0.15).all()
         assert g.np.isclose(sliced.bounds[1], mesh.bounds[1]).all()
         assert len(sliced.faces) == 5
+
+        # Same test with capping (should only add three more triangles)
+        sliced_capped = mesh.slice_plane(plane_origin=plane_origin,
+                                         plane_normal=plane_normal,
+                                         cap=True)
+
+        assert len(sliced_capped.faces) == 8
+        assert g.np.isclose(sliced_capped.bounds[0], mesh.bounds[1] - 0.15).all()
+        assert g.np.isclose(sliced_capped.bounds[1], mesh.bounds[1]).all()
+        assert sliced_capped.is_watertight
 
         # Cut top off of box and make sure bounds and number of faces is correct
         # Tests new quads and entirely contained triangles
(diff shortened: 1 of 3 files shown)


mikedh:
Awesome thanks for the PR! Yeah polygonization of segments being slow was what discouraged me last time I looked at this. It's possible it could be special cased to be faster, as the `trimesh.path` polygon logic is doing a lot of stuff, most of which is probably not necessary for this. Triangle's input is some pretty simple [segments](https://github.com/mikedh/trimesh/blob/f0433cf6d886b1031d0d448daf5693df692d6f9f/trimesh/creation.py#L466-L501), so one thing we could do is special-case sections with a single connected component and skip all the interior-exterior loop checking stuff. Maybe there's a fancy way of using the (already calculated) face normals to figure out which loops are interiors? Though it might not be worth suffering hahah, 85ms isn't *that* slow. 

A while back I think I also benchmarked against `shapely.ops.polygonize_full` and the trimesh stuff was surprisingly not that much slower (IIRC, not certain this is still true). 

That being said, the performance of capping doesn't make the existing function slower, and people definitely want this. I'll see if anything comes to mind over the next few days then merge. Thanks for PR'ing the cap logic and cleanups!

mjd3:
No problem! Yeah skipping the polygons and feeding directly to triangle seems possible, but I'm not sure exactly how that might go. Probably not worth the headache for now :p.

mikedh:
Agreed, those methods sounds likely to be a finicky nightmare and not worth it haha. Thanks for the PR, I'll release shortly.

PR #809, opened 2020-04-22
adapting to_planar for generic normal cases

pmberkeley:
Maybe not the most elegant solution, but should fix problems with plane_transform behavior for this function, and allow for rotation during planar transforms.  May have eliminated functionality with removal of (N *= np.sign(np.dot(N, normal)).

### DIFF
--- a/trimesh/path/path.py
+++ b/trimesh/path/path.py
@@ -949,7 +949,7 @@ class Path3D(Path):
         normal_test = False
         for i in [[1, 0, 0], [0, 1, 0], [0, 0, 1],
                   [-1, 0, 0], [0, -1, 0], [0, 0, -1]]:
-            if not np.any(gm.align_vectors(dn, i) - np.eye(4)) :
+            if not np.any(align_vectors(dn, i) - np.eye(4)):
                 normal_test = True
         if normal_test is False:
             raise ValueError('normal does not align with major axis')
@@ -994,7 +994,7 @@ class Path3D(Path):
 
         # create the Path2D with the same entities
         # and planar values of vertices
-        planar_columns = [0,1,2]
+        planar_columns = [0, 1, 2]
         planar_columns.remove(axis)
         planar = Path2D(entities=copy.deepcopy(self.entities),
                         vertices=flat[:, planar_columns],
--- a/trimesh/path/path.py
+++ b/trimesh/path/path.py
@@ -935,7 +935,7 @@ class Path3D(Path):
                 dn = N
             # eliminate calculation errors
             C, N, dn = util.snap_to(C), util.snap_to(N),
-                                            util.snap_to(dn)
+                                    util.snap_to(dn)
             # create a transform from fit plane to destination plane
             to_2D = plane_transform(origin=C,
                                     normal=N,
--- a/trimesh/path/path.py
+++ b/trimesh/path/path.py
@@ -933,7 +933,7 @@ class Path3D(Path):
                 dn = N
             # create a transform from fit plane to destination plane
             to_2D = plane_transform(origin=C,
-                                    normal=N
+                                    normal=N,
                                     destination_normal=dn)
 
         # make sure we've extracted a transform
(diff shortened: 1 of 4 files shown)


mikedh:
Thanks for the PR! I think this is missing changes to `trimesh.geometry.plane_transform` at the moment?

pmberkeley:
Yeah, something happened with my computer while I was updating last night,
and I haven't gotten a chance to troubleshoot yet.

pmberkeley:
A few tests from pytest are still failing, and I couldn't figure out why. I also tested out the functions on my files, and the functionality is working if nothing is input into to_planar, but I'm getting errors associated with the logic of inputting to_2D and normal values, so I have to mess with that some more before it's ready to go. (Sorry for any strange git activity, this is my first time really using it and I am still learning the etiquette).

mikedh:
Thanks for the PR!

- This supersedes #811 right?
- I think the snippet here is going to just return an identity matrix, right?
```
    if destination_normal is None:
        destination_normal = normal
```
Which in order to duplicate the prior behavior needs to be:
```
    if destination_normal is None:
        destination_normal = [0, 0, 1]
```
- The CI failing is I believe related to the changed behavior of `plane_transform`
- My understanding of this issue is hazy at best hahah, the most helpful thing would be if you could add a unit tests which fails on current master, but passes with the fix from the PR.

pmberkeley:
> This supersedes #811 right?

Yep. I had started them as separate branches, not realizing that they would separately fail pytest because they weren't combined. This is the all-in-one fork.

> I think the snippet here is going to just return an identity matrix, right?
>    ```if destination_normal is None:
>         destination_normal = normal

That's the idea. Currently, the destination normal is assumed implicitly to be the Z axis, which is what's causing all the unintentional rotation problems. (If you're working with the X or Y axis as your normal, it's implicitly rotating it to the Z axis in several different spots in the code, which causes unexpected and inconsistent behavior). This part of the code is intended to only change the normal axis if a destination normal is provided. Otherwise it assumes the same axis. (I'm mostly self-educated on this type of coding logic in python, so I'm not sure this is the best way to handle it, but that's the intent).

> My understanding of this issue is hazy at best hahah, the most helpful thing would be if you could add a unit tests which fails on current master, but passes with the fix from the PR.

Where/how would I do this? (Where is this currently located - I could probably figure out how to do this if directed to the right example).

pmberkeley:
It just occurred to me that you mean that the plane_transform needs a unit normal test. I currently have that located in to_planar; I'll have to see how the logic would work if I moved it over to plane_transform. I had thought it would be best to keep plane_transform as a function that could do whatever it was needed for (eg to rotate one plane to any other; I tested this out on my own object, and it was quite useful and worked as-is).

trimesh/path/path.py
```diff
@@ -918,34 +921,54 @@ def to_planar(self,
...
+                normal.shape))
 
         # make sure we've extracted a transform
         to_2D = np.asanyarray(to_2D, dtype=np.float64)
         if to_2D.shape != (4, 4):
             raise ValueError('unable to create transform!')
 
+        # make sure normal aligns with X, Y, or Z axis
+        normal_test = False
+        for i in [[1, 0, 0], [0, 1, 0], [0, 0, 1],
```
mikedh, line 953: Probably an easier way to do this check is with `np.isclose`, which you can pass `atol` and `rtol` values for float comparisons:
```
In [1]: import numpy as np

In [2]: normal = [0,0,1.0]

In [3]: np.isclose(np.abs(normal), np.eye(3))
Out[3]: 
array([[False,  True, False],
       [ True, False, False],
       [ True,  True,  True]])

In [5]: check = np.isclose(np.abs(normal), np.eye(3)).all(axis=1)

In [6]: check
Out[6]: array([False, False,  True])
```

mikedh:
Thanks for the PR! Is this ready for another review pass?

mikedh:
Hey, I'm closing this for now as it's gotten stale. Looking back at it I think the main issue with merging was that I don't exactly understand what the desired behavior is in this case and so the fixes are a bit confusing to me. 

Maybe the best way to tackle this is if you could make a gist or PR with a bunch of `assert`'s that fail, and that way we can debate the correct behavior separately from the fixes? Thanks for the work on this, and sorry for not getting it merged!

PR #811, opened 2020-04-23
Plane_transform modified for passing explicit normal

pmberkeley:
Modified plane_transform so the transform defaults to the provided plane normal, but which allows for a destination normal to be included in the transform. I'm not entirely clear on why anybody would want to modify the normal of the plane this far downstream; I would personally think people would have modified the normal before this point, but decided to maintain the functionality of plane_transform in order to avoid functionality conflicts I'm not aware of.


mikedh:
Superceded by #809

PR #823, opened 2020-05-08
Loops in the closest_point() method vectorized with numpy arrays.

dbukenberger:
Hai Mike,

it's me again on the closest_point() methods. 😅 
I needed this to be a bit faster so I replaced the loop-parts with vectorized NumPy operations where possible.
Feel free to check it out and let me know what you think. 😊

Cheers,
Dennis

### DIFF
--- a/trimesh/proximity.py
+++ b/trimesh/proximity.py
@@ -12,8 +12,6 @@ from .grouping import group_min
 from .constants import tol, log_time
 from .triangles import closest_point as closest_point_corresponding
 
-from collections import deque
-
 
 def nearby_faces(mesh, points):
     """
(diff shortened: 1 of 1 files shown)


mikedh:
Looks awesome! I'm seeing a roughly a 2x speedup from this, thanks for the PR! 

So now the slowest part (~70% of runtime) is the rtree-based `nearby_faces` method. I was thinking a voxel-binning method could potentially perform better, so to investigate I added a `return_index` method to `subdivide_to_size`, so we could potentially do a super coarse voxelization of the mesh and then use each voxel as a "bin of original face indexes." Not sure if that actually winds up being faster though hahah.

dbukenberger:
Yeah, the 'seriality' of the rtree appears to be kind of a bottleneck right now. But that's a cool idea - I will play around with that one. 😅

Issue #824, opened 2020-05-08
GLTF loader missing meshes with same name


jengvi:
First of all - great library!

I am having a problem loading a GLTF scene that includes multiple meshes with the same name. The GLTF [here](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/www/TestScene/TestScene.gltf) has two meshes with the name `Cube`. Here is what this looks like in another GLTF viewer:
<img width="1227" alt="Screen Shot 2020-05-08 at 4 27 03 PM" src="https://user-images.githubusercontent.com/56892628/81447391-8a2c7c00-914a-11ea-85ff-8fdf82ffbebd.png">
This is what I see in Trimesh:
<img width="1912" alt="Screen Shot 2020-05-08 at 4 26 52 PM" src="https://user-images.githubusercontent.com/56892628/81447543-d24b9e80-914a-11ea-869c-25215582f0a9.png">
Note one missing cube in the right side of the scene. If I manually change the name of one of the cubes to another name like `Cube2`, I can get it to appear in Trimesh. I am using Trimesh 3.6.35 and loaded this scene with `merge_primitives=True`.  I'm looking through gltf.py now to see if I see anything obvious...

Thanks in advance!

loudonicus:
Hi you've managed to get further than me in loading a gltf, I wonder if you could give me step for a hint? I am trying to load one of the examples from Khronos, Avocado.gltf in jupyter notebook. 

```
currResolver = trimesh.visual.resolvers.FilePathResolver("C:/Users/brian/Dropbox/Python/")
gltfname = "Avocado.gltf"
gltf_import = trimesh.exchange.gltf.load_gltf(open(gltfname,"rt"),currResolver)
gltfScene = trimesh.Scene(gltf_import)  
```

Jupyter notebook then reports 

unknown type (str) added to scene!
unknown type (str) added to scene!
unknown type (float32) added to scene!
unknown type (float32) added to scene!
unknown type (float32) added to scene!
unknown type (float32) added to scene!
unknown type (float32) added to scene!
unknown type (float32) added to scene!
....

and then gltfScene.show() gives a whole host of errors.

Sorry for the noob question, but have I missed a vital step?

mikedh:
Hey, I don’t think all the instantiaton is necessary... have you tried just
‘trimesh.load(gltfname)’?

loudonicus:
ah! embarassing... yup that works a treat! Guess I was just making life too difficult for myself!

[...]

PR #827, opened 2020-05-11
Fix loading of gltf scenes with duplicate node names

jengvi:
Potential fix for #824


mikedh:
Looks great, thanks for the PR! Merged and should be released shortly.

PR #842, opened 2020-05-19
GLTF: Store multiple primitives as parent/child nodes 

fmannhardt:
The fix in https://github.com/mikedh/trimesh/pull/835 helps with loading nodes with multiple primitives, but it forgets the relationship between the GLTF primitives and their GLTF mesh as each primitive gets added with a random node suffix. This makes it difficult to reference the nodes by their GLTF identifier. 

This PR proposes to store each primitive as separate node with geometry under a common parent without geometry to retain the possibility to select nodes by the original GLTF identifier.


mikedh:
Looks like a great solution, thanks for the PR!

fmannhardt:
Thanks. I hope this works well with the rest of the library.

PR #849, opened 2020-05-25
Fix transparency bug in DAE exchange

mmatl:
Hi Mike,

Happy Memorial Day! I noticed that my prior PR introduced a bug in the DAE importer/exporter. I assumed that transparency was listed as 0-255, but it's actually 0-1. This bug resulted in loaded/saved meshes "disappearing" unexpectedly. This simple PR should fix the issue.

Best,
Matt


mikedh:

PR #873, opened 2020-06-11
Add without_world axis state to the viewer

kosuke55:
The current viewer has 'world' and 'all' in the display state of axis. If set it to'all', it can display all coordinates I added with the world axis. But I'm confused because I don't know which one I added. So in this PR I added a state of 'without_world' so that I can only see added coordinates.

world
state 'world'  

all
state 'all'  

without_world
state 'without_world'


mikedh:
Looks nice, thanks for the PR!

PR #894, opened 2020-07-01
Re-enable fix_inversion

clbarnes:
Some of the code makes the assumption that an inside-out trimesh will have a negative volume. However, this is not true as the `mass_properties` method returns the absolute volume if the mesh looks ok *except* for being inside-out ( https://github.com/mikedh/trimesh/commit/f0394c671a3703e08430ef7cf0c2def2652667a3 ). This means that, for single-body meshes, `fix_inversion` would never do anything.

Additionally, makes a mesh fix its winding when given `validate=True`.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -43,6 +43,21 @@ class MeshTests(g.unittest.TestCase):
         assert len(m.vertices) == pre_len
         assert g.np.isclose(m.volume, pre_vol)
 
+    def test_validate_inversion(self):
+        """Make sure inverted meshes are fixed by `validate=True`"""
+        orig_mesh = g.get_mesh("unit_cube.STL")
+        orig_verts = orig_mesh.vertices.copy()
+        orig_faces = orig_mesh.faces.copy()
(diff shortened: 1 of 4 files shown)


clbarnes:
Note that `is_volume` will still return `True` for incorrectly (but consistently) wound meshes. Is that desired behaviour?

mikedh:
Thanks for the PR and test! And yeah that's an oops. I am sort of wondering if we shouldn't also remove the "silent `abs`" logic inside `mass_properties`, seems like I planted a minefield and then immediately drove through it.

clbarnes:
Yes, this PR is a bit of a band-aid. Maybe best not to make allowances for nearly-valid meshes if there's no need to, I think trimesh can stand to be a little opinionated!

mikedh:
Sounds good, I stripped out the confusing `abs` logic in `trimesh.base` that was left over from when `validate` was still a "global-ish" property and your new test still passed. I'll release shortly, thanks for the PR!!

PR #901, opened 2020-07-03
Fix change detection for Path objects in the viewer

### LINKED ISSUE #897: trimesh.viewer.windowed: geometry_hash does not take color of path into account
The `geometry_hash` function of the `trimesh.viewer.windowed` module does not hash the colors of path entities, this means that changing colors of path entities does not get reflected in the viewer.

The `md5` method of `trimesh.path.path.Path` also does not include the colors in its hash.

(As an aside, why use MD5 for this? It was designed to be cryptographically strong. There are surely better ways to do change detection.)

I guess the proper way to fix this is to let path entities have something `visual`-like.

lericson:
Fixes #897 

See commit messages and #897 for discussion

### DIFF
--- a/trimesh/path/path.py
+++ b/trimesh/path/path.py
@@ -233,6 +233,20 @@ class Path(object):
 
         return target
 
+    @property
+    def visual_crc(self):
+        """A CRC of the current entity colors.
+
+        Returns
+        -------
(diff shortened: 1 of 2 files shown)


mikedh:

PR #905, opened 2020-07-05
Use tobytes instead of tostring

unnonouno:
`tostring` is alias of `tobytes`, and `tostring` is deprecated from numpy ver 1.19.0. I replaced `tostring` with `tobytes`.
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tostring.html

Note that `tobytes` is implemented in ver 1.9.0. In an old environment, this fix may causes an error.
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html#numpy.ndarray.tobytes

### DIFF
--- a/tests/test_mutate.py
+++ b/tests/test_mutate.py
@@ -35,8 +35,8 @@ def get_readonly(model_name):
     verts = original.vertices
     faces = original.faces
     # use the buffer interface to generate read-only arrays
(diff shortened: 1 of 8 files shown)


mikedh:
Looks good, thanks for the PR! Numpy 1.9 released in 2014 so it seems like that should be fine, even the oldest Ubuntu LTS still active (16.04) was on 1.11. I'll test a little and then merge.

unnonouno:
Thank you very much!

PR #907, opened 2020-07-06
proximity.closest_point returns correct point in ambiguous cases

brucedjones:
When computing the closest point on a mesh to a query point, we first get a list of nearby triangles and the closest points on those triangles. If the closest two points are both within some tolerance distance from the query point, then an additional test is made to select the point that lies in the direction closest to each respective triangle normal.

While the additional test is appropriate, the closest point returned may not correspond the the closest triangle, and instead to the other candidate triangle.

This change updates the closest point and distance to that point to correspond with the closest triangle chosen in teh ambiguous case. A unit test was added to ensure this is working, which fails before this change.


mikedh:

PR #913, opened 2020-07-12
Added a check for convex to avoid Qhull error

wwwslinger:
A zero-z (i.e. flat) mesh still has the `convex_hull` attribute, but is not convex.  This results in errors with, for example, getting the `bounding_box_oriented` attribute:
```
QhullError: QH6022 qhull input error: 2'th dimension's new bounds [-0.5, 0.5] too wide for
existing bounds [ 0,  0]
```
This PR just adds two checks to avoid some errors I've come across.  Not sure if there's enough need to look elsewhere since this was created for 3D meshes, but let me know and I might give it a whirl.

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -2109,8 +2109,10 @@ class Trimesh(Geometry):
         convex : trimesh.Trimesh
           Mesh of convex hull of current mesh
         """
-        hull = convex.convex_hull(self)
-        return hull
+        if self.is_convex:
+            hull = convex.convex_hull(self)
+            return hull
+        return None
 
     def sample(self, count, return_index=False):
         """
--- a/trimesh/bounds.py
+++ b/trimesh/bounds.py
@@ -135,7 +135,7 @@ def oriented_bounds(obj, angle_digits=1, ordered=True, normal=None):
     # extract a set of convex hull vertices and normals from the input
     # we bother to do this to avoid recomputing the full convex hull if
     # possible
-    if hasattr(obj, 'convex_hull'):
+    if hasattr(obj, 'convex_hull') and obj.is_convex:
         # if we have been passed a mesh, use its existing convex hull to pull from
         # cache rather than recomputing. This version of the cached convex hull has
         # normals pointing in arbitrary directions (straight from qhull)


wwwslinger:
Oops, forgot about the tests.  Testing convex when not convex should probably be avoided or asserted that it isn't convex.  Thoughts @mikedh ?

mikedh:
Thanks for the PR! Yeah a test would be great, I think the best illustration of this issue is probably with a 2 triangle mesh where all vertices are shared but the order is reversed:

```
In [1]: import trimesh

In [3]: import numpy as np

In [4]: m = trimesh.Trimesh(vertices=np.eye(3), faces=[[0,1,2],[2,1,0]])
 
In [5]: m
Out[5]: <trimesh.Trimesh(vertices.shape=(3, 3), faces.shape=(2, 3))>

In [6]: m.is_watertight
Out[6]: True

In [7]: m
Out[7]: <trimesh.Trimesh(vertices.shape=(3, 3), faces.shape=(2, 3))>

In [8]: m.show()
Out[8]: SceneViewer(width=1800, height=1350)

In [9]: m.bounding_box_oriented
---------------------------------------------------------------------------
QhullError                                Traceback (most recent call last)
<ipython-input-9-a67dc03fc9a8> in <module>()
----> 1 m.bounding_box_oriented

/home/mikedh/trimesh/trimesh/caching.py in get_cached(*args, **kwargs)
    101             return self._cache.cache[name]
    102         # value not in cache so execute the function
--> 103         value = function(*args, **kwargs)
    104         tic[2] = now()
    105         # store the value

/home/mikedh/trimesh/trimesh/parent.py in bounding_box_oriented(self)
    165         """
    166         from . import primitives, bounds
--> 167         to_origin, extents = bounds.oriented_bounds(self)
    168         obb = primitives.Box(transform=np.linalg.inv(to_origin),
    169                              extents=extents,

/home/mikedh/trimesh/trimesh/bounds.py in oriented_bounds(obj, angle_digits, ordered, normal)
    136     # we bother to do this to avoid recomputing the full convex hull if
    137     # possible
--> 138     if hasattr(obj, 'convex_hull'):
    139         # if we have been passed a mesh, use its existing convex hull to pull from
    140         # cache rather than recomputing. This version of the cached convex hull has

/home/mikedh/trimesh/trimesh/caching.py in get_cached(*args, **kwargs)
    101             return self._cache.cache[name]
    102         # value not in cache so execute the function
--> 103         value = function(*args, **kwargs)
    104         tic[2] = now()
    105         # store the value

/home/mikedh/trimesh/trimesh/base.py in convex_hull(self)
   2110           Mesh of convex hull of current mesh
   2111         """
-> 2112         hull = convex.convex_hull(self)
   2113         return hull
   2114 

/home/mikedh/trimesh/trimesh/convex.py in convex_hull(obj, qhull_options)
     57 
     58     hull = spatial.ConvexHull(points,
---> 59                               qhull_options=qhull_options)
     60 
     61     # hull object doesn't remove unreferenced vertices

qhull.pyx in scipy.spatial.qhull.ConvexHull.__init__()

qhull.pyx in scipy.spatial.qhull._Qhull.__init__()

QhullError: QH6214 qhull input error: not enough points(3) to construct initial simplex (need 4)

While executing:  | qhull i QbB Qt Pp
Options selected for Qhull 2015.2.r 2016/01/18:
  run-id 1446062571  incidence  QbBound-unit-box 0.5  Qtriangulate
  Pprecision-ignore  _pre-merge  _zero-centrum

```

mikedh:
There a bunch of edge cases here that I think we probably should be handling so this may be an ongoing effort. I made a pass at handling one of these cases (spherical) but I'd like to do it a bit more generally in the `fix/obb` branch. Here are the cases that come to mind:

```
	# a single point in space
	case_sphere = [[0, 102, 12.0]]
	transform, extents = trimesh.bounds.oriented_bounds(case_sphere)
	# obb size should be all zeros
	assert np.allclose(extents, np.zeros(3))
	# transform should move point back to the origin
	assert np.allclose(
            trimesh.transform_points(case_sphere, transform), 0.0)

	# check a line in space
	case_line = [[0, 0, 0], [0, 0, 1]]

	# check another colinear case with more than 2 points
	case_line = [[0, 0, 0], [0, 0, 1], [0, 0, 213]]

	case_plane = [[0, 0, 0], [1, 0, 0], [0, 1, 0]]

	# single and double coplanar meshes
	case_single_tri = trimesh.Trimesh(vertices=np.eye(3),
                                            faces=[[0, 1, 2]])
	case_double_tri = trimesh.Trimesh(vertices=np.eye(3),
                                            faces=[[0, 1, 2], [2, 1, 0]])

	# create a case with many random coplanar triangles
	vertices = np.random.random(30).reshape((-1, 3))
	vertices[:, 2] = 1
	case_many_coplanar = trimesh.Trimesh(
            vertices=vertices,
            faces=np.arange(9).reshape((-1, 3)))

```

PR #923, opened 2020-07-22
Update cameras.py

Alexibu:
fov exception pasta


mikedh:

PR #925, opened 2020-07-22
Export vertex colors if available.

jjmontesl:
I am using this change in order to export vertex colors when they are available. I needed this to export vertex colors when also having TextureVisuals.

Maybe this is not right, please let me know if I should use another approach.

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -577,7 +577,7 @@ def _append_mesh(mesh,
     assert len(buffer_items) >= tree['accessors'][-1]['bufferView']
 
     # check to see if we have vertex or face colors
-    if mesh.visual.kind in ['vertex', 'face']:
+    if mesh.visual.kind in ['vertex', 'face'] or hasattr(mesh.visual, 'vertex_colors'):
         # make sure colors are RGBA, this should always be true
         vertex_colors = mesh.visual.vertex_colors
         # add the reference for vertex color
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! I'm slightly hesitant to merge this because:
- For meshes with no visual properties defined it will *always* export the default vertex colors, because the default visual object is `ColorVisuals` and `hasattr(mesh.visual, 'vertex_colors')` will always be true.
- it relies on assigning properties to the `TextureVisuals` object that aren't always there. 

That being said I see why one might want that. There are some super involved fixes one could potentially do (i.e. supporting multiple visuals) but that seems like a rabbit hole for a very minor win, and would make the default use cases even more confusing.

What about just renaming that property to a unique private name, something like `mesh.visual._forced_vertex_colors`? Not super elegant but it seems like it would do the trick.

jjmontesl:
Thanks for the insight.

- I agree that the multiple visuals approach is overkill at least at the moment.
- I agree that reusing `mesh.visual.vertex_colors` with different semantics and availability is not desirable.

I guess that adding optional vertex color support to TextureVisuals would be a reasonable way to go. If not set, it wouldn't be exported. For that I would indeed use `vertex_colors`, set to None when not available. What do you think?

mikedh:
Sounds great! I talked with some graphics friends about this issue (i.e. should `TextureVisuals` always have a vertex color) and their general take was that they kind of just wanted a bin of arbitrary vertex attributes.

Since `TextureVisuals` already have a `DataStore` which is only used for UV's, I just renamed it to `TextureVisuals.vertex_attributes`. Then if you set `TextureVisuals.vertex_attributes['color'] = ...` we can check for that in the GTLF exporter, and it'll get the `update_vertices` mask applied and be hashed, etc.  

I'll update the check and release, thanks for the PR!

jjmontesl:
Sounds great. Yes, some shaders require multiple custom vertex attributes, so this is a much better approach. Thanks!

PR #926, opened 2020-07-23
blender interface to smart unwrap

michele-arrival:
As promised, here's an implementation of the interface with blender to create uv-maps for any kind of base objects.

The exchange format has to be obj because:
- stl does not carry texture coordinates
- blender ply format is not readable by trimesh (would have been preferred otherwise) - saves only only in ASCII (so no gain in precision there)
- glb-gltf is very slow in blender (and seems imprecise)
- there's no other "overlap" in the set of formats supported by trimesh and blender

result is far from optimal but very very usable if one wishes to "enrich" their bland, geometry-only, STL files. Here's an example using the Duck.glb model:

image


mikedh:
Awesome, thanks for the PR! One quick question:

> blender ply format is not readable by trimesh (would have been preferred otherwise) - saves only only in ASCII (so no gain in precision there)

So the issue I had with supporting the blender UV format was that I couldn't figure out where/how/if blender exported the corresponding texture image. Any ideas, happy to support this if they are actually outputting the image.

michele-arrival:
> Awesome, thanks for the PR! One quick question:
> 
> > blender ply format is not readable by trimesh (would have been preferred otherwise) - saves only only in ASCII (so no gain in precision there)
> 
> So the issue I had with supporting the blender UV format was that I couldn't figure out where/how/if blender exported the corresponding texture image. Any ideas, happy to support this if they are actually outputting the image.

So I checked and you are right, the ply exported by blender does not have a texture. Originally I thought it was because the input file didn't have one.
But i tried to do it manually using an already textured model, and it just doesn't include it in the ply file.

I assumed that the mesh that you want "unwrapped" is not going to have any TextureVisuals (therefore no image). 
Anything modelled manually is going to be better than the maps that blender can fabricate with no or little information.

But maybe it's more common than i think to have an asset with messed up uv, and one would want to recreate them anyway. In that case, would the trimesh ply exporter write the image in the ply file? 
If the ply exporter doesn't write the image, then it doesn't matter that blender cannot write it back after unwrap.

Issue #917, opened 2020-07-13
The default smoothing of the OpenGL widget


LorenzoDAsie:
Hi,

While it's easy to remove smoothing on the trimesh.show() function, I am having a hard time figuring out how to remove smooting from trimesh.viewer.SceneWidget. There is no option to disable it and I cannot find the option when digging into the code.

It would be useful to expose the option on the widget, or perhaps implement the possibility at all. I am happy to help implementing the solution and contribute to the project, but I would need some help in where to look.

mikedh:
Happy to take PR's! The place to figure out how to pass `smooth=False` would be the call to `rendering.convert_to_vertexlist`:

https://github.com/mikedh/trimesh/blob/master/trimesh/viewer/widget.py#L277

LorenzoDAsie:
Perfect, I will work on it and let you know soon. Thanks!

PR #932, opened 2020-07-27
Adding option smooth=False Issue#917

LorenzoDAsie:
Adding the smooth option to the widget. It was smooth=True as default, now it is possible to set the option to False when initializing the widget. This should solve Issue#917


mikedh:
Awesome thanks for the PR! Looks good, and don't worry about the white space check, I'll autoformat before releasing.

LorenzoDAsie:
Ok! Thanks! And thanks for this amazing package!

mikedh:
Fixes #917

PR #934, opened 2020-07-29
Fixed bug in path.remove_entities()

acbergan:
Super simple fix. This was not working and was not used anywhere in the library.


mikedh:
Looks great, thanks for the fix and test!

PR #939, opened 2020-08-02
rotate matrix bug fix

bigeyesung:
I tested it if we want to rotate 45 deg in Y axis in the mesh.center as origin, the transform matrix should be changed.


mikedh:
Ah yeah that order is confusing. Thanks for the PR!

PR #948, opened 2020-08-10
Add color attribute to path gltf export

Gmadges:
The Path Entity contains vertex color data.
I've added this data as the COLOR_0 attribute for gltf export.
This allows uses to export Paths and maintain the color data in other software


mikedh:
Awesome, thanks for the PR!

yingzhang1109:
I am wondering if you can share an example on how to export path with color to a gltf file? Thanks in advance

PR #950, opened 2020-08-11
Adding __radd__ method for multiple concatenation

hoodwinker:
if you override only __add__ then you can only concatenate two meshes at once.

when you try to do sum([mesh1, mesh2, mesh3]) sum starts with 0 + mesh1 and fails.
this will provide concatenations with sum:
now sum starts with "0 + mesh1" then fails and run mesh1 + 0 = mesh1.

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -2849,3 +2849,25 @@ class Trimesh(Geometry):
         """
         concat = util.concatenate(self, other)
         return concat
+
+    def __radd__(self, other):
+        """
+        Concatenate the mesh with 0.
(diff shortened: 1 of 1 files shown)


mikedh:
Ah that explains the difference between `np.sum` (works) and `sum` (doesn't). Thanks for the PR!

One other related issue is that `util.concatentate([a, b, c, d, ...])` is WAY faster than `a + b + c + d...` or `sum(a, b, c, d, ...)` because it can avoid a bunch of instantiation work. I guess there's no `__sum__` magic method unfortunately.

PR #952, opened 2020-08-12
added PointCloud support to gltf export

Gmadges:


mikedh:

mikedh:
If you're feeling energetic it would be awesome to get the loader side working too, so we could round trip point clouds as GLB/GLTF. But if not no worries!

PR #971, opened 2020-08-27
Speed up proximity.signed_distance and add vertex/face attributes to ply export

brucedjones:
Please see individual commits for details.

### DIFF
--- a/tests/test_ply.py
+++ b/tests/test_ply.py
@@ -78,6 +78,52 @@ class PlyTest(g.unittest.TestCase):
         assert g.np.array_equal(
             point_list['some_float'], g.np.array([1.1, 2.2], dtype=g.np.float32))
 
+    def test_vertex_attributes(self):
+        """
+        Test writing vertex attributes to a ply, by reading them back and asserting the
+        written attributes array matches
+        """
+
+        m = g.get_mesh('box.STL')
+        test_1d_attribute = g.np.copy(m.vertices[:, 0])
+        test_nd_attribute = g.np.copy(m.vertices)
+        m.vertex_attributes['test_1d_attribute'] = test_1d_attribute
+        m.vertex_attributes['test_nd_attribute'] = test_nd_attribute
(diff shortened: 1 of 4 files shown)


mikedh:
Awesome thanks for the PR! When profiling I was seeing a solid 20% speedup on the signed distance change, looks great! 


For the PLY vertex attribute export I had two quick thoughts: first, rather than being passed to the exporter it would be nice if it automatically got the attributes from `mesh.face_attributes` and `mesh.vertex_attributes` if they were populated:
```
In [5]: import trimesh

In [6]: m = trimesh.creation.box()

In [7]: m.vertex_attributes
Out[7]: {}

In [8]: m.face_attributes
Out[8]: {}
```
Although I could see there could be some challenges with data types doing that and if it's not possible we can skip it :smile:. Second, rather than `kwargs.get(___, [])` I'd prefer to add an explicit kwarg like `vertex_attributes=None`, or `include_vertex_attributes=True` so the docstring is easier for people to understand.

brucedjones:
Anythings possible when we write the code :)

Agree that using the existing vertex/face attributes on the mesh makes more sense, I wasn't aware that these were a thing 😳 . Opted to read from those only and not add the keyword args at all.

I made that change, and while I was there I generalized to accept multidimensional attributes. I also wrote some tests this time around.

mikedh:
That's awesome thanks for the PR and test, the attributes change is great! I've heard from multiple people that they like the "bag of vertex and face attributes" paradigm and adding support throughout the library is really nice. 

I might make the additional change in a few weeks moving re-imported attributes back into `mesh.face_attributes` instead of `mesh.metadata['ply_raw']['face']['data']` or if you're feeling ambitious always happy to take additional PR's :smile:.

Merged to master and CI should do the release shortly.

brucedjones:
heh, I seriously considered doing so but unfortunately couldn't justify the time for my current needs.

PR #994, opened 2020-09-15
Fix export of ASCII ply with attributes

brucedjones:
**Summary**
Exporting a mesh with attributes with ASCII encoding would result in a malformed ply file which specified the attributes in the header but did not include them in the data section. This fix was facilitated by a new `unstructured_array_to_string` method. I originally looked at generalizing the existing method but decided upon a bespoke method instead due to the implications of mixed data type rows.

Also fixed a couple bugs in the existing `array_to_string` method

**Test Plan**
- Added tests for `array_to_string`
- Added tests for `unstructured_array_to_string`

### DIFF
--- a/trimesh/util.py
+++ b/trimesh/util.py
@@ -1131,10 +1131,15 @@ def array_to_string(array,
         raise ValueError('conversion only works on 1D/2D arrays not %s!',
                          str(array.shape))
 
-    # allow a value to be repeated in a value format
-    repeats = value_format.count('{}')
+    # abort for structured arrays
+    if array.dtype.names is not None:
+        raise ValueError(
+            'array is  structured, use structured_array_to_string instead')
 
-    if array.dtype.kind == 'i':
+    # allow a value to be repeated in a value format
+    repeats = value_format.count('{')
+
+    if array.dtype.kind in ['i', 'u']:
         # integer types don't need a specified precision
         format_str = value_format + col_delim
     elif array.dtype.kind == 'f':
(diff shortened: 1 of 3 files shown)


trimesh/util.py
```diff
@@ -1167,6 +1172,102 @@ def array_to_string(array,
...
+    """
+    # convert inputs to correct types
+    array = np.asanyarray(array)
+    digits = int(digits)
+    row_delim = str(row_delim)
+    col_delim = str(col_delim)
+    value_format = str(value_format)
+
+    # abort for non-flat arrays
+    if len(array.shape) > 1:
```
brucedjones, line 1215: Note that it is assumed here that the second dimension is represented by the structured array elements

trimesh/util.py
```diff
@@ -1167,6 +1172,102 @@ def array_to_string(array,
...
+        format_str += element_row_length * element_format_str
+
+    # length of extra delimiters at the end
+    format_str = format_str[:-len(col_delim)] + row_delim
+
+    # expand format string to whole array
+    format_str *= len(array)
+
+    # Flatten the array to a list as numpy array column type cannot be mixed
+    partially_flattened = [
```
brucedjones, line 1254: I'm not super happy with this part and would have preferred to keep this as a numpy reshape but that doesn't work because numpy doesn't support mixed types in an array column.

Maybe a custom generator would be more efficient than list comprehension + concatenation, but I figured this works for now and that is an optimization that can be done if required.

mikedh:
Looks great, thanks for the PR! I'll make a pass through tomorrow and then release.

trimesh/util.py
mikedh: Yeah the flattening operation is tough. I was playing with this a bit, if you stay in numpy-land and only loop through field names it's quite a bit faster (~45x):
`flattened = np.hstack([array[k].reshape((count, -1)) for k in array.dtype.names]).reshape(-1)`

As you say it upgrades everything to a float. On the other hand if we wangled `value_format= '{:0.0f}'` or possibly `value_format.replace('{}', '{:0.0f}')` for those fields we might be able to cut off the decimal using the string format we're doing anyway.

PR #1011, opened 2020-09-26
Fix without_world axis state

kosuke55:
I'm sorry to make a bug.
In #873, I added `without_world` axis state. 
I thought the first element of `node_names` was the world, but in reality it wasn't always the case.
Therefore, it was modified to judge whether it is world from the value of transform.


mikedh:
No worries thanks for the fix! I updated the check to use `util.allclose` (very similar to `np.allclose`) as `transform` is a float and will release shortly:
```
-                if not np.all(transform == np.eye(4)):
+                if not util.allclose(transform, np.eye(4), atol=1e-5):

```

Issue #1018, opened 2020-09-30
Missing penetration depth


jonnymaserati:
Great code! I'm using it on a project I'm working on and it's indeed easier to use than the python-fcl library.

However... for my application I need to know the depth of the penetration of collided objects. The base FCL code calculates this but for some reason the trimesh code is not relaying this information.

Can you add this line or something similar to the ContactData class?

![image](https://user-images.githubusercontent.com/41046859/94735671-68d36300-036b-11eb-8303-5c9c166915d8.png)

Regards, Jonny

jonnymaserati:
Maybe `self._depth` is more consistent with your other property naming?

mikedh:
Yeah that would make sense! Happy to take PR's!

jonnymaserati:
> Yeah that would make sense! Happy to take PR's!

Just tried but I got an error trying to push my branch.

`remote: Permission to mikedh/trimesh.git denied to jonnymaserati.
fatal: unable to access 'https://github.com/mikedh/trimesh.git/': The requested URL returned error: 403`

PR #1035, opened 2020-10-22
Added penetration depth

jonnymaserati:
Reading `penetration_depth` from `ContactData` and created a `depth` property. It follows the same format as the point data.

Added a line in `test_collision.py` to assert that the depth data is being read.


jonnymaserati:
#1018

mikedh:

PR #1059, opened 2020-11-17
Gltf import fix

CiuffoliniFLx:
Hi!

1. 
I needed to load .glb scenes with no meshes, only empty nodes. 
I know this is a corner case, but according to Khronos group docs, this looks to be within the specifications.
These files produced a reading error, as "bufferViews" and "meshes" keys were not found while reading the header. 

I've added 2 small ifs to check if "bufferViews" and "meshes" actually exists in the header; if not, the code skips directly to the nodes.

2. 
I needed to read extra properties written on the gltf scenes. 
If I'm not mistaken the importer doesn't read this info, only the header extras.
I've added a a _parse_scene_extras method that reads this info, only for the loaded scene.
This info is then added to the metadata dict as scene_extras

3. i've added .idea to .gitignore

it looks like tests run without any error.
thank you!


mikedh:
Thanks for the PR! Would you mind also adding a small test model and a unit test?

mikedh:
Thanks for the test! One other quick question: is `parse_scene_extras` necessary? It already puts extras into metadata:
```
In [1]: import trimesh

In [2]: m = trimesh.creation.icosphere()

In [3]: s = trimesh.Scene(m)

In [4]: s
Out[4]: <trimesh.Scene(len(geometry)=1)>

In [5]: s.metadata
Out[5]: {}

In [6]: s.metadata['extras'] = {'hi': 20}

In [7]: r = trimesh.load(trimesh.util.wrap_as_stream(s.export(file_type='glb')), file_type='glb')

In [8]: r.metadata
Out[8]: {'extras': {'hi': 20}}

In [9]: s.metadata
Out[9]: {'extras': {'hi': 20}}

```

CiuffoliniFLx:
Hi mikedh,
thank you for the quick feedback.
I'm committing a test file that should pass the check formatting, too.

Sorry if i'm being slow, but i'm not sure i'm following.

I've added a monkey.glb model to the repo, that has some extras written on the Scene entity.
Can you give me a minimal example on how to load that .glb and print the extras?
I'll be happy to delete the code if it's not needed 

Ty!

mikedh:
Ah I see, before there was the root-level `extras` whereas this is `scene/extras`. Thanks for the fix and test!

CiuffoliniFLx:
Thank you Michael :)

PR #1065, opened 2020-11-23
Allow gltf tree postprocessing during export

noamgat:
This allows the user of the library to add custom properties (for example, GLTF extensions to materials) during export. I am using this to add the https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit extension to materials containing the string "unlit" in their name, but this is only one example.

### DIFF
--- a/tests/test_gltf.py
+++ b/tests/test_gltf.py
@@ -467,7 +467,7 @@ class GLTFTest(g.unittest.TestCase):
         gltf_2 = g.trimesh.exchange.gltf.export_gltf(scene, tree_postprocessor=add_unlit)
 
         def extract_materials(gltf_files):
-            return g.json.loads(gltf_files['model.gltf'])['materials']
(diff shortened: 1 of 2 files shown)


noamgat:
Example:

`
def example_tree_postprocessor(gltf_tree):
    for material_dict in gltf_tree['materials']:
        if 'unlit' in material_dict.get('name', '').lower():
            material_dict["extensions"] = {
                "KHR_materials_unlit": {}
            }
    gltf_tree["extensionsUsed"] = ["KHR_materials_unlit"]
`

mikedh:
Looks great, thanks for the fix! Would you mind adding that example to the unit tests, just to make sure the code path stays tested?

noamgat:
Done

mikedh:
Awesome, thanks for the PR! I'll release shortly.

PR #1066, opened 2020-11-24
min_distance_single function parameter nameing issue

bigeyesung:
Hi,
I think it is a minor thing but just to bring it up here.
I noticed that "min_distance_single" function has a "return_name" parameter. But in the same collision class other functions are using "return_names" as their parameters. Not sure if it is due to the return value of "min_distance_single" has only a singular "name". In my suggestion, it is better to make all functions the same so they are easier to use. Let me know what you think.
Also, I don't know how to change API webpage, could you tell me how to do that? thx

https://trimsh.org/trimesh.collision.html#trimesh.collision.CollisionManager.min_distance_single

```
def min_distance_single(self,
                            mesh,
                            transform=None,
                            return_name=False,
                            return_data=False):
```

### DIFF
--- a/tests/test_collision.py
+++ b/tests/test_collision.py
@@ -124,17 +124,17 @@ class CollisionTest(g.unittest.TestCase):
         dist = m.min_distance_single(cube)
         assert g.np.isclose(dist, 4.0)
 
-        dist, name = m.min_distance_single(cube, return_name=True)
+        dist, name = m.min_distance_single(cube, return_names=True)
         assert g.np.isclose(dist, 4.0)
         assert name == 'cube1'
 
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! I agree the kwarg name change makes sense, but I'm a little hesitant to merge changes that break people's kwarg function calls.

It's kind of arbitrary, but my general rule of thumb is that breaking kwargs is fine as part of a larger logic refactor or fixes, but I'd like to avoid them if they're just a little awkward (like in this case). There are a whole bunch of `plane_origin` kwargs I would really like to rename to `origin` but oh well, I'd rather not inflict it on users unless there are other benefits coming along.

PR #1067, opened 2020-11-24
rename to vedo

marcomusy:
project `vtkplotter` has been renamed to `vedo` + link to example gallery.


mikedh:
That's probably a better name! Gallery looks awesome, thanks for updating the readme!

PR #1068, opened 2020-11-24
Only add the default material once to the GLTF tree

noamgat:
Currently, the default material is re-added to the materials queue for every path in the scene. This is wasteful (and also causes unneeded state switches when rendering).
This fixes it to add the default material at most once, no matter how many paths are added.


mikedh:
Looks great, thanks for the PR! I'll confirm this still works with `GL_LINES`/`Path` objects which is what I think it was there for then release.

PR #1074, opened 2020-12-01
Fix in the inertia moment computation.

dbukenberger:
Hai Michael,

I have tracked down some strange behavior to the point in trimesh where the inertia moment is computed in the `mass_properties()` method of the `triangles.py`.

Compared to the literature reference, given in the comments of the method, there were actually three minus signs missing.

I have fixed that point and added two tests in the `test_inertia.py`.

With this fix the `negate_nondiagonal` matrix in `inertia.py` is also no longer required.
Or did this actually serve any purpose? I have not seen this anywhere in the literature.

Cheers,
Dennis

### DIFF
--- a/tests/test_inertia.py
+++ b/tests/test_inertia.py
@@ -13,6 +13,7 @@ except BaseException:
 #    3
 
 triIdxs = g.np.ravel([[2, 1, 0], [3, 0, 1], [3, 2, 0], [3, 1, 2]])
+
 def tetToTris(tet):
     return g.np.reshape(tet[triIdxs], (-1, 3))
 
@@ -172,7 +173,7 @@ class InertiaTest(g.unittest.TestCase):
                                  [-1, -1, -1],
                                  [-1, 1, -1],
                                  [1, 1, -1],
-                                 [1,-1, -1]]) * 0.5
+                                 [1, -1, -1]]) * 0.5
 
         # 6 quad faces for the cube
         quads = g.np.int32([[3, 2, 1, 0],
@@ -238,6 +239,6 @@ class InertiaTest(g.unittest.TestCase):
             assert g.np.allclose(MI_gt, MI)
 
 
-if __name__ == '__main__':   
+if __name__ == '__main__':
     g.trimesh.util.attach_to_log()
     g.unittest.main()
--- a/tests/test_inertia.py
+++ b/tests/test_inertia.py
@@ -3,11 +3,9 @@ try:
 except BaseException:
     import generic as g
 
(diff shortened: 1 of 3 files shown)


mikedh:
Awesome, thanks for the PR! 

So the previous logic was sort of "studying for the test": i.e. the principal axis of inertia of a transformed cylinder should be aligned with the cylinder axis. The `negate_nondiagonal` was copying what they were doing in some of the examples in this [MIT rigid body lecture note](https://ocw.mit.edu/courses/aeronautics-and-astronautics/16-07-dynamics-fall-2009/lecture-notes/MIT16_07F09_Lec26.pdf) which did the "right thing" although I didn't/don't know why :smile:.

We're also checking a few "truth" examples in `test_util.py` (I should move that to `test_inertia.py`) generated in Solidworks. The incorrect signs must have been on non-diagonal terms we're not checking:

```
In [9]: [np.array(i['inertia']) for i in g.data['mass_properties']]
Out[9]: 
[array([[0.008312, 0.      , 0.      ],
        [0.      , 0.012823, 0.000807],
        [0.      , 0.000807, 0.009005]]),
 array([[1.2108243e-01, 0.0000000e+00, 0.0000000e+00],
        [0.0000000e+00, 9.1319810e-02, 1.3893000e-04],
        [0.0000000e+00, 1.3893000e-04, 2.0576510e-01]])]

```

mikedh:
Hey, so it actually looked like the mass tests (which run on semi-complex meshes and validate against results pulled from Solidworks) weren't being run against this change, and when I added them back they disagreed with the signs on those non-diagonal terms.  

I'm not totally sure what's going on (i.e. is Solidworks computing a slightly different quantity or something? does Solidworks just `abs` every non-diagonal?). I checked the [polyhedral mass properties notes](https://www.geometrictools.com/Documentation/PolyhedralMassProperties.pdf) and they definitely agree with your sign change :smile:. I updated `test_inertia.py` to include those truth tests in my release branch `fix/multbody` which is now nicely failing haha, here's the error:
```
ValueError: angle_block.STL(inertia):
[[ 0.00832  0.       0.     ]
 [ 0.       0.01283 -0.00081]
 [ 0.      -0.00081  0.00901]]
!=
[[0.00831 0.      0.     ]
 [0.      0.01282 0.00081]
 [0.      0.00081 0.00901]]
```

Any idea what's going on here?

mikedh:
[Solidworks Forum: Positive or Negative Inertia Tensors?](https://forum.solidworks.com/thread/84904)

My read: we can just negate the non-diagonal values in the truth data (which passes and I'll release soon). I had no idea SW was doing that in their inertia results. Thanks for tracking down and fixing/testing this!

PR #1082, opened 2020-12-07
3mf fixes

carolina-mesquita-hp:
Hello there!

I found some more test cases the 3MF loader wasn't covering.

First, the kwargs passed to `load_3mf` weren't saved, and so if I tried to load a file with the `process=False` flag, vertices were still merged. 

Second, if objects had a `name` attribute, they were used as dict keys. However, there is nothing in the [3MF specification](https://github.com/3MFConsortium/spec_core/blob/master/3MF%20Core%20Specification.md#chapter-4-object-resources) specifying that this attribute has to be unique, and so I found a case where there were two different objects with the same name, and one object "overwrote" the other. So I just removed this part of the code to let all objects be indexed by id.

I added two unit tests for these cases.

Thanks for your work in the project!


mikedh:
Awesome, thanks for the fix and test!

PR #1088, opened 2020-12-14
camera.look_at() example

damienallen:
Took me a while to find the [one example](https://github.com/mikedh/trimesh/blob/ee58b616cfebf2d1e9066a57388ef028ceb86cd0/examples/outlined.py#L29) where `camera.look_at()` is used. Thought it might be useful to have a small example in the docs.


mikedh:

PR #1092, opened 2020-12-16
collision.py: Make an error more explicit

pgunn:
Was helping someone use this package today and had to read the code to figure out what wasn't working. This diff improves the error message enough that scientists might have a shot at fixing the problem themselves without reaching out to their local software support people


mikedh:
Look good, thanks for the PR!

PR #1100, opened 2020-12-24
added alphaCutoff material mapping

gleb-an:


mikedh:

PR #1101, opened 2020-12-28
Fix for import/export of empty ply and obj files.

dbukenberger:
Hai Mike,

recently I ran into some trouble while processing a dataset with objects and files from various sources.
While empty OBJ files could be loaded, empty PLY files would crash.
Therefore, I tried to fix this, so that empty PLY files would also just create an empty Scene instance, without vertices or faces.

Do you think this is a useful fix?
I was unsure because some tools, i.e. Blender 2.79b, didn't support empty PLY exports either, but then the recent versions of Blender do ... and I think by the official PLY [file-specification](http://paulbourke.net/dataformats/ply/), geometry is actually not a requirement for a valid PLY file. 

Cheers,
Dennis


dbukenberger:
Now I have further extended the support for importing and exporting empty or point-cloud files (where possible) in OBJ, PLY and STL format.
I know trimesh is primarily a mesh-library, but I thought it would be quite useful if I could internally handle points-only content analogously to my mesh objects - at least for compatible file formats.

Cheers,
Dennis

mikedh:
Thanks for the PR, and happy new year! Looks good. Quick question, didn't PLY importer already support empty meshes as `PointCloud` objects?

mikedh:
Thanks for the PR and tests! On a re-read I realized this was for fully empty (not just no faces like I was thinking) meshes and this made more sense 😄. Yeah I agree loading an empty scene/mesh is a lot more graceful than straight failure!

dbukenberger:
Happy new year. 😄
Oh okay, I missed what happens with really _empty_ files.
PLY fails with a `ValueError: Not a ply file!`
Do you feel, we should include that case as well and initialize an empty cloud or should this remain as valid error?

PR #1106, opened 2021-01-01
Add metadata for primitive meshes

iory:
# What is this.

This PR adds metadata of `shape` for primitive objects such as `box`, `icosphere`, `uv_sphere`, `capsule`, `cone`, `cylinder`, and `annulus`.
This will be helpful for converting primitive `mesh` objects to other formats like signed distance function. https://www.iquilezles.org/www/articles/distfunctions/distfunctions.htm


mikedh:

iory:
Thank you!

PR #1120, opened 2021-01-17
added support for Close commands in svg path import

ljmatkins:
this adds SVG import support for the SVG close path command (https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand), which functions as a Line command to the initial point in the current subpath.

this prevents SVG paths ending in a Close command from missing their final line segment.


mikedh:

PR #1130, opened 2021-01-26
fix interpolation of attributes in subdivide()

cal-armstrong:
current implementation with vstack() compiles the list of attributes such that they are grouped by g[0] (from all triangles) then g[1] (from all triangles) then g[2] (from all triangles). This is different to how they are grouped in faces_to_edges(). New implementation with hstack() and reshape() lists the attributes such they are grouped first by each triangle, then within each triangle by g[0] then g[1] then g[2].

### DIFF
--- a/trimesh/remesh.py
+++ b/trimesh/remesh.py
@@ -81,11 +81,11 @@ def subdivide(vertices,
         new_attributes = {}
         for key, values in vertex_attributes.items():
             attr_tris = values[faces_subset]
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! Looks like tests are failing, would you mind adding a minimal example or a unit test?

mikedh:
Closing for now, but feel free to re-open with tests!

Issue #692, opened 2020-01-16
Constraining/Pinning points when smoothing mesh?


KelSolaar:
Hi,

I wanted to know if it is possible to constrain or pin some points when smoothing a mesh?

Cheers,

Thomas

mikedh:
Hey, it's not currently an option. You could probably get that behavior by smoothing, then between each iteration for each of your pinned points find the nearest vertex on the smoothed result and snap its value to the pinned point.

KelSolaar:
Thanks! I had that in mind actually however I was thinking there would be a smarter way, maybe with custom Laplacian matrix?

bbarroqueiro:
Hi Thomas... 

You can modified the laplacian easily but the problem would be to known the index of pinned point! In fea branch  of trimesh, you can find the code for a node picker. Thus, just pick the nodes that you want to pin, and construct the laplacian with the neighboors weights to zero for corresponding pinned points.... and there you go...

bbarroqueiro:
It appears the code was moved to trimesh/examples/pick_vertex.py

regards

mikedh:
Yeah totally, you can also create a `cKDTree` between iterations to find the closest vertex to a pinned point, its constructor is quite fast:

```
In [17]: mesh
Out[17]: <trimesh.Trimesh(vertices.shape=(1425, 3), faces.shape=(2990, 3))>

In [18]: mesh.bounds
Out[18]: 
array([[-0.5, -0.5, -0.5],
       [ 0.5,  0.5,  0.5]])

In [19]: pinned = mesh.bounds.copy()

In [20]: pinned
Out[20]: 
array([[-0.5, -0.5, -0.5],
       [ 0.5,  0.5,  0.5]])

In [22]: from scipy.spatial import cKDTree

In [24]: distance, index = cKDTree(mesh.vertices).query(pinned)

In [25]: index
Out[25]: array([757, 815])
```

The custom laplacian matrix also sounds like a interesting solution, happy to take PR's if you get it working!

bbarroqueiro:
Just did a PR for custom Laplacian matrix... 

Likely to late... 

example:

pinned = mesh.bounds
distance, index = cKDTree(mesh.vertices).query(pinned)
lap=smoothing.laplacian_calculation(mesh, pinned_vertices=index)
smoothing.filter_mut_dif_laplacian(mesh,lamb=0.1,iterations=10,volume_constraint=True,laplacian_operator=lap)

PR #1136, opened 2021-01-30
new smoothing stuff

bbarroqueiro:
Hi Mike

As promised, the PR for recently published new laplacian filter (a mutable diffusion algorithm).

I also a custom Laplacian calculation  in order to constraint some vertices as per issue "Constraining/Pinning points when smoothing mesh? #692"

Let me know what you think

Regards
BB


bbarroqueiro:
Nonetheless if volume constraint is True, the pinned vertices can drift a bit using the Laplacian smoothing and mutable diffusion algorithm (less but still present)...

I forgot to add this remark. The volume constraint can be extended to exclude the pinned nodes.
But it not worth it for practical use... the drift is small

mikedh:
Thanks for the PR, sounds good! I'll try to test and benchmark in the next week or so and get this merged.

Issue #199, opened 2018-07-31
Enhancement Ideas


mikedh:
I was pinged recently looking for some projects to help out with, here are some ideas. PR's are always appreciated!

*Updated 7/27/2020*

- (docs) Figure out how to get the examples embedded in the actual docs webpage. Right now when you build docs, example ipynb files are rendered to HTML which is  just [linked to from the examples page](https://trimsh.org/examples.html). It would be really nice if each example was embedded in sphinx somehow, and kept the sidebar on the left so one could navigate around between examples and back to the docs.
- (feature) Mesh simplification, either through creating a `meshlabserver` interface or through implementing a vectorized numpy version of quadratic edge collapse: https://github.com/mikedh/trimesh/issues/41
- (feature) MSH support, as requested in https://github.com/mikedh/trimesh/issues/206. Probably via `meshio` 
- (feature) Color and line width in Path objects: https://github.com/mikedh/trimesh/issues/356
- (feature) More 3MF tests. Trimesh has a 3MF importer, it would be great to put together a test corpus of various small- ish models (specifically interesting would be multibody assemblies with duplication) from various exporters. 
- (feature) Do whatever it takes to get `pyinstaller` to work with `trimesh`, as per #412 
- (infrastructure) Put all of Shapenet, Thingiverse 10k, GLTF samples, and all available meshes into an AWS S3 bucket, and continuously benchmark trimesh, pymesh, and assimp for percentage of files loaded and time.  #510 
- (community) See if upstream `numpy` has interest in including `TrackedArray` or `ndarray.crc` #283 
- (speedup) Add broad-phase culling for `mesh.slice_plane` to speed up on large meshes: #630 
- (feature) Exact rather than approximate medial axis calculation as per #924

Marviel:
@mikedh awesome list, we'd all love to see some of these changes, stoked to know you're considering things like built-in mesh simplification, etc.

A convention I've seen in some other repos is to have each task of this nature listed as their own github issue, and give them an `easy` `medium` `hard` label applied to them, based on the perceived difficulty.

Also allows us to start threaded conversations on each of the tasks in your list.

Anyhow just a thought! Not sure the extent to which we want to use Issues here :)

mmatl:
@mikedh I'm going to try to tackle textures. Let me know if you have any thoughts -- I was also thinking a new `TextureVisuals` object. Loading the texture file itself will probably require`imageio` -- are you okay with introducing that dep? It's really lightweight.

[...]

PR #1162, opened 2021-02-23
Allow use of a two-dimensional pitch in Path2D.rasterize

tpank:
Enhancement proposed in https://github.com/mikedh/trimesh/issues/199#issuecomment-776796485


mikedh:

PR #1171, opened 2021-03-05
Added support for 'extras' metadata to nodes and glTF/glb import/export.

jjmontesl:
The current version allows for glTF `extras` metadata at scene level only.

This PR adds support for 'extras' metadata at node level through the scene graph. These are exported and imported to/from glTF/glb.

image

What do you think?


mikedh:
Looks great, thanks!

PR #1178, opened 2021-03-14
Remove debug print statement.

drknotter:
Sorry for the thrash!


mikedh:
No worries thanks for the followup!

PR #1185, opened 2021-03-16
Adding metadata field primitive with value True to geometries loaded …

ibozhilov:
…from GLTF primitives.

### DIFF
--- a/tests/test_gltf.py
+++ b/tests/test_gltf.py
@@ -517,10 +517,10 @@ class GLTFTest(g.unittest.TestCase):
         s = g.get_mesh('CesiumMilkTruck.glb')
 
         # Assert that primitive geometries are marked as such
-        assert s.geometry['Cesium_Milk_Truck_0'].metadata.get('from_gltf_primitive', False)
+        assert s.geometry['Cesium_Milk_Truck_0'].metadata['from_gltf_primitive']
 
         # Assert that geometries that are not primitives are not marked as such
-        assert not s.geometry['Wheels'].metadata.get('from_gltf_primitive', False)
+        assert not s.geometry['Wheels'].metadata['from_gltf_primitive']
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! 

- it would be nice if the flag was *always* populated rather than sometimes true and sometimes nonexistent
- could we add the full explanation/backstory for the flag somewhere, something like: "Some GLTF meshes include more than one primitive, which translates to multiple Trimesh objects. This flag being True indicates that this Trimesh object was one of multiple GLTF primitives under a single GLTF mesh." Maybe a rename to something like `from_gltf_multiprimitive_mesh`, or `from_gltf_primitive` (not sure if either of those are clearer).

ibozhilov:
Usually I go with no default value for such kind of meta, but having default value will be more consistent with the trimesh philosophy, so I have set the meta to be False by default. I have also change the key to `from_gltf_primitive` as you have suggested. As per back story I can suggest something like that:
"GLTF meshes consist of one or more primitives (You can read more here - https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes). If the mesh is represented as multiple primitives, this will result in multiple trimesh objects - one per each primitive. If the trimesh object originated from multi-primitive GLTF mesh you can use, then `trimesh.metadata['from_gltf_primitive']` will be True, otherwise False."
However, I'm not sure where in the docs this is supposed to go. Also the automated tests fail and I'm not entirely sure why.

mikedh:
Awesome thanks for the changes! Don't worry about builds, they were broken upstream and I'm hoping they will be fixed by https://github.com/fredrik-johansson/mpmath/issues/580 although if they're not fixed in the next few days I'll remove the sympy install in python 2.

ibozhilov:
Great! I'm not sure what is your release cycle, but would it be possible to make a release with the new field, once you merge the PR. Thanks in advance!

PR #1189, opened 2021-03-19
Make section_multiplane doc match parameters

PaulDebus:
The order of the doc string did not match the order of the parameters for the `section_multiplane` method


mikedh:
Thanks for the fix!

PR #1190, opened 2021-03-19
Make mesh_multiplane doc match parameters

PaulDebus:
The order of the parameters in the function definition of ` mesh_multiplane` did not match the order of the docstring. Changed the docstring to match.


mikedh:
Dupe of #1189

mikedh:
Oops different file. Thanks for the fix!

PR #1206, opened 2021-04-06
Fix range for baseColorFactor's transparent

iory:
In case of importing collada file, 
the alpha value set is 0-255, but it should be in the range 0-1.
https://github.com/mikedh/trimesh/blob/master/trimesh/exchange/dae.py#L313-L316


mikedh:
Thanks for the PR! This makes me think the `PBRMaterial` objects should have a validation method or something.

Issue #1265, opened 2021-06-06
[Bug?] trimesh.trimesh.subdivide with textured mesh does not produce expected result


nikwl:
On Ubuntu 18.04, python3.6, trimesh version 3.9.20.

### Bug?

[Drive link to the shapenet file](https://drive.google.com/drive/folders/1w5trdSU8FZ_N8ZJcYYx6XJBmAsLBuB3O?usp=sharing)

I was playing around with the library trying to supersample a ShapeNet model using trimesh.trimesh.subdivide and trimesh.remesh.subdivide_to_size. Looking at the output, it seems like the uv is not being subdivided correctly.
 
Here's the ShapeNet model before subdivide:
![purse](https://user-images.githubusercontent.com/38300882/120925747-ea705300-c6a7-11eb-8116-776f98425500.png)

And after subdivide
![purse_wonky](https://user-images.githubusercontent.com/38300882/120925752-f0663400-c6a7-11eb-9891-c6011101cfc1.png)

Using the following code:
```
import trimesh
mesh = trimesh.load('test/purse.obj')
mesh.show()
mesh = mesh.subdivide()
mesh.show()
```

I don't have much experience with uv so I'm not sure if this is the expected result - it doesn't seem right though. 

### Potential Fix for subdivide 
I was able to get the seemingly correct output by appending the uv coords to the vertices:
![purse](https://user-images.githubusercontent.com/38300882/120925747-ea705300-c6a7-11eb-8116-776f98425500.png)
```
mesh = trimesh.load('test/purse.obj')

# Concat the uv with the vertices and subdivide
vertices, faces = trimesh.remesh.subdivide(
    vertices=np.hstack((mesh.vertices, mesh.visual.uv.copy())),
    faces=mesh.faces
)

mesh.vertices = vertices[:, :3]
mesh.faces = faces
mesh.visual.uv = vertices[:, 3:]

mesh.show()
```

Is this a generalizable solution? It works because trimesh.remesh.subdivide doesn't double check that the vertices are actually nx3.

### Potential Fix for subdivide_to_size
This is also a possible workaround with trimesh.remesh.subdivide_to_size (which is what I was actually interested in) with one tweak to the function:
```
def subdivide_to_size(vertices, faces, max_edge, max_iter=10, return_index=False):
    """
    Subdivide a mesh until every edge is shorter than a
    specified length.
    Will return a triangle soup, not a nicely structured mesh.
    Parameters
    ------------
    vertices : (n, 3) float
[...]

PR #1266, opened 2021-06-11
Fix: subdivide crocodile patterns

nikwl:
In reference to https://github.com/mikedh/trimesh/issues/1265.

Rewrote base.subdivide. 
Seems like the visual was only kept if: 
1) the visual was an instance of trimesh.visual.texture.TextureVisuals and
2) the uv coordinates were the same shape as the vertex coordinates. 

The rewrite keeps this behavior. 

Changes pass the test_remesh.py unittset. 

Wasn't really sure how to write a unittest for pattern correctness, but uncommenting the mesh.show() in test_remesh.py will demonstrate that the patterns are gone.  

Followup question - is there any reason that subdivide_to_size isn't implemented in trimesh.base? As I mentioned in my original issue subdivide_to_size is really what I'm after, and I'd be happy to add that as well if it's not missing for a specific reason.

### DIFF
--- a/tests/test_remesh.py
+++ b/tests/test_remesh.py
@@ -68,6 +68,7 @@ class SubDivideTest(g.unittest.TestCase):
     def test_uv(self):
         # get a mesh with texture
         m = g.get_mesh('fuze.obj')
+        #m.show()
         # get the shape of the initial mesh
         shape = m.vertices.shape
         # subdivide the mesh
@@ -79,6 +80,7 @@ class SubDivideTest(g.unittest.TestCase):
         assert s.vertices.shape[0] > shape[0]
         # should have UV coordinates matching vertices
         assert s.vertices.shape[0] == s.visual.uv.shape[0]
+        #s.show()
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


mikedh:
Looks great, thanks for the PR! Also not sure how we could unit test for this well, but as a basic sanity check it looks like the following would have caught the current thing (and passes with your PR):
```
    mesh = trimesh.load_mesh('models/fuze.obj')

    o =	mesh.visual.uv[mesh.faces]
    ms = mesh.subdivide()
    s = ms.visual.uv[ms.faces]

    # if each face has wildly different UV coordinates it probably means
    # something was done incorrectly
    assert o.ptp(axis=1).mean(axis=0).max() < 0.1
    assert s.ptp(axis=1).mean(axis=0).max() < 0.1

```

> Followup question - is there any reason that subdivide_to_size isn't implemented in trimesh.base? As I mentioned in my original issue subdivide_to_size is really what I'm after, and I'd be happy to add that as well if it's not missing for a specific reason.

No reason, happy to take PR's adding it!

PR #1270, opened 2021-06-14
Add support for per-vertex texture coordinates for ply files

georgegu1997:
Problem: to load the png texture for ply file, the code assumes there is `texcoord` in the ply file for each face. But for some ply files [generated by MeshLab](https://www.agisoft.com/forum/index.php?topic=16.0) (one example mesh model is attached), the texture coordinates are stored for each vertex as 'texture_u' and 'texture_v'. Therefore, the code failed to create the texture without any complaints. 

Fix: Check if related keys are in the vertex data, the code converts the texture coordinates from vertices to faces. 

[example mesh.zip](https://github.com/mikedh/trimesh/files/6650881/example.mesh.zip)

### DIFF
--- a/trimesh/exchange/ply.py
+++ b/trimesh/exchange/ply.py
@@ -524,6 +524,13 @@ def elements_to_kwargs(elements,
             # accessing numpy arrays with named fields
             # incorrectly is a ValueError
             pass
+    
+    # If texture coordinates are defined with vertices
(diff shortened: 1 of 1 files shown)


mikedh:
Awesome thanks for the PR! Would you be able to add a unit test adding a small mesh (maybe a cube from meshlab) which has those parameters?

PR #1271, opened 2021-06-15
Release: Materials And Fixes

mikedh:
- release #1266 fixing subdivide with UV coordinates and add simple test
- release #1246 adding multi-material support
- release #1270 adding per-vertex UV exported how meshlab exports


ibozhilov:
Do you have a timeline on when the release will go live?

mikedh:
Sorry for the delay, just released! Yeah in testing another thing I was thinking would be nice for `MultiMaterial.collapse` or something that ran concatenation (which also has packing bugs at the moment 😢) so then `trimesh.rendering` and other exporters and things that need a single material could work seamlessly.

Thanks for the work on the PR!

PR #1274, opened 2021-06-20
Feature: added subdivide_to_size to base

nikwl:
In reference to https://github.com/mikedh/trimesh/issues/1265.

Added subdivide_to_size to the trimesh base class. 
Added 3 unit tests that mimic the checks for the original subdivide_to_size fn.
Needed to update remesh to prevent it from requiring vertices are of size nx3 (to allow passing hstacked vertices and uv coords). 

Changes pass the test_remesh.py unittest.

Thanks!


mikedh:

Issue #1299, opened 2021-07-26
Where is scene.graph.transforms.remove_node(node_name)?


clemense:
Hi,
There used to be a method called `scene.graph.transforms.remove_node(node_name)`. This is gone now (I assume this happened during the scene graph refactoring). What's the equivalent in the current trimesh version?

Thanks!

PR #1300, opened 2021-07-26
add remove_node method to EnforcedForest class

clemense:
Adding a method to remove a node.

This is related to issue #1299


mikedh:
Awesome thanks! Would you be able to add a quick validation test to `tests/test_scenegraph.py` that confirms the tree structure still generally functions after the operation?

clemense:
I added a test, probably not the most creative one. I'm open to any further suggestions.

mikedh:
Looks great, thanks for the PR! I'll release later today.

PR #1305, opened 2021-07-30
Typo fix in exchange.obj docstrings

mz8i:
Lack of colon was causing incorrect docs generation
(`bool` swapped with `include_texture` param name in RTD HTML)
Added colon - this should fix generated docs.


mikedh:

PR #1309, opened 2021-08-05
Fix material merging

zian-w:
The offsets returned by packing.images is XY coordinate, which is different from UV coordinate (flip Y). When stacking texture map in the Y direction, this function will give errors.


mikedh:
Awesome thanks! This issue has been on the back of my mind for a while, this solution looks great! Would you be able to add a quick unit test that hits this code path and does a simple check, maybe to `tests/test_texture.py`?

mikedh:
Thanks for the PR!

PR #1317, opened 2021-08-09
better support loading uv coordinates from ply

smerzbach:
Ply has no clear definition of how texture coordinates are stored, before this fix, many ply files are loaded without texture coordinates, as trimesh currently only parses texture_u and texture_v fields. This fix additionally enables parsing of texture coordinates defined as pairs (u,v) or (s,t) in the ply header.


mikedh:
Awesome thanks for the PR! Would you be able to add a small test model that hits this code path?

smerzbach:
Hey, I added the test `test_blender_uv()` in `test_ply.py` that should cover the changed code, together with some small ply files.

I also made some further changes regarding the order of handling non-triangular faces (triangles mixed with quads, as can be found, e.g., in `suzanne.ply`). There was a bug when indexing the `texcoord` array with `faces` that was never triggered when the texture coordinates were not loaded before my first changes. Now, the faces are first ensured to be triangular, then the texture coordinates are handled.
Another small change is an omitted check for `image` to be different from `None`. This check prevented `TextureVisuals` to be instantiated and hence the texture coordinates not being stored in the object, even after successful parsing. As far as I understand the constructor of TextureVisuals already handles `image` being `None`.

mikedh:
Looks perfect, thanks for adding the test!

> Another small change is an omitted check for image to be different from None. This check prevented TextureVisuals to be instantiated and hence the texture coordinates not being stored in the object, even after successful parsing. As far as I understand the constructor of TextureVisuals already handles image being None.

Makes sense! Yeah UV's without an image is more common than I would have thought 😄

PR #1324, opened 2021-08-15
calculate integral mean curvature

keisuke-ishihara:
This function calculates the integral mean curvature of a mesh. With this addition, all four Minkowski functionals for a surface mesh embedded in 3d can be calculated by Trimesh. (Volume, Area, Integral Mean Curvature, Euler number)

In the most general case, the lengths of the two numpy arrays,
self.face_adjacency_angles
and
self.edges_unique_length,

may not be equal. This could be checked with an 'assert', if necessary.

### DIFF
--- a/tests/test_integralmeancurvature.py
+++ b/tests/test_integralmeancurvature.py
@@ -3,42 +3,43 @@ try:
 except BaseException:
     import generic as g
 
+
 class IntegralMeanCurvatureTest(g.unittest.TestCase):
-	
+
     def test_IMCsphere(self):
-    	# how close do we need to be - relative tolerance
-   		tol = 1e-3
-   		for radius in [0.1, 1.0, 3.1459, 29.20]:
-   			m = g.trimesh.creation.icosphere(subdivisions=4, radius=radius)
-   			IMC = m.integral_mean_curvature
-   			ref = 4*g.np.pi*radius
-   			assert g.np.isclose(IMC, ref, rtol=tol)
+        # how close do we need to be - relative tolerance
+        tol = 1e-3
+        for radius in [0.1, 1.0, 3.1459, 29.20]:
+            m = g.trimesh.creation.icosphere(subdivisions=4, radius=radius)
+            IMC = m.integral_mean_curvature
+            ref = 4*g.np.pi*radius
+            assert g.np.isclose(IMC, ref, rtol=tol)
 
     def test_IMCcapsule(self):
-    	# how close do we need to be - relative tolerance
-   		tol = 1e-3
-   		radius = 1.2
-   		for aspect_ratio in [0.0, 0.5, 1.0, 4.0, 100]:
-   			L = aspect_ratio*radius
-   			m = g.trimesh.creation.capsule(height=L,
-   										   radius=radius, count=[64,64])
-   			IMC = m.integral_mean_curvature
-   			ref = g.np.pi * (L + 4 * radius)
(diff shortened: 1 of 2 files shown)


mikedh:
Nice idea thanks for the PR! Is there a paper somewhere that has reference values for some common primitives (sphere, box, capsule) we could check for a unit test?

So `edges_unique` isn't necessarily going to correspond to the `face_adjacency` indexing scheme, I think the lengths need to be recalculated from `face_adjacency_edges`. This can probably be done more gracefully but for now I think this  replicates your logic with the realigned indexing:

```
In [28]: edges_length = np.linalg.norm(np.subtract(*m.vertices[m.face_adjacency_edges.T]), axis=1)

In [30]: edges_length
Out[30]: 
array([1.        , 1.        , 1.41421356, 1.        , 1.        ,
       1.        , 1.41421356, 1.41421356, 1.        , 1.        ,
       1.        , 1.        , 1.41421356, 1.41421356, 1.41421356,
       1.        , 1.        , 1.        ])

In [31]: m.face_adjacency_angles
Out[31]: 
array([1.57079633, 1.57079633, 0.        , 1.57079633, 1.57079633,
       1.57079633, 0.        , 0.        , 1.57079633, 1.57079633,
       1.57079633, 1.57079633, 0.        , 0.        , 0.        ,
       1.57079633, 1.57079633, 1.57079633])

In [32]: m.face_adjacency_angles * edges_length
Out[32]: 
array([1.57079633, 1.57079633, 0.        , 1.57079633, 1.57079633,
       1.57079633, 0.        , 0.        , 1.57079633, 1.57079633,
       1.57079633, 1.57079633, 0.        , 0.        , 0.        ,
       1.57079633, 1.57079633, 1.57079633])

In [33]: (m.face_adjacency_angles * edges_length).sum() * 0.5
Out[33]: 9.42477796076938

```

keisuke-ishihara:
Thanks for the suggestion of using face_adjacency to obtain the edge lengths!

Some background of integral mean curvature can be found in slide 44-56 in [this lecture note (pdf)](https://brickisland.net/DDGFall2017/wp-content/uploads/2017/11/CMU_DDG_Fall2017_10_DiscreteCurvature.pdf).

I've derived the integral mean curvature (IMC) for the common primitives:

For a sphere of radius R:
IMC = (mean curvature) * (area) = 1/R * 4* pi *R^2 = 4* pi *R

For a capsule defined as the cylinder of radius R and length L with its two ends decorated by a hemispherical cap of radius R:
IMC = (IMC of cylinder) + (IMC of two hemispheres)
= 1/2R * 2* pi *R*L + 4* pi *R = pi*L + 4 * pi * R
note: L should be equivalent to the height parameter for the 'capsule' function in  trimesh/creation.py

For a box defined by length L, width W, height H, we use the discrete definition:
IMC = sum(edge lengths * angles) * 0.5 = 4*(L+W+H) * pi/4 * 0.5 = (L+W+H)*pi/2 

Let me know where I could implement the unit test. If you could point to a good example (e.g. volume and area of common primitives) that I can follow, that would be nice.

mikedh:
Awesome unit test looks great, thanks for deriving the analytical equivalents!

keisuke-ishihara:
Thanks for the feedback! This was my first time contributing to a software package with a PR. I learned about unit tests and the tools for PEP8 formatting, and it was fun.

mikedh:
Thanks for the contribution, it's much appreciated! Glad contributing wasn't too painful!

Issue #1315, opened 2021-08-07
RuntimeError: generator raised StopIteration when calling `is_filled` method


Co1lin:
`is_filled` method works fine for some points, but it crashes for some other points.

The error below indicates that for some inputs, argument `ordered_indices` for function `sorted_rle_gather_1d` in `runlength.py` is empty.
```python
Traceback (most recent call last):
  File "/home/aidrive1/miniconda3/envs/dr/lib/python3.8/site-packages/trimesh/voxel/runlength.py", line 378, in sorted_rle_gather_1d
    index = next(index_iter)
StopIteration
```

Details:

```python
In [5]: voxel2_surface
Out[5]: <trimesh.VoxelGrid(18, 18, 18)>

In [6]: v1_points
Out[6]: 
array([[ 1.15973   , -0.25133982, -0.0182148 ],
       [ 1.15973   , -0.18839218, -0.0182148 ],
       [ 1.15973   , -0.25133982,  0.04473284],
       ...,
       [ 1.852154  , -0.37723509,  0.42241865],
       [ 1.852154  , -0.37723509,  0.48536629],
       [ 1.852154  , -0.37723509,  0.54831393]])

In [7]: voxel2_surface.is_filled([[0.2,0.2,0.2]])
Out[7]: array([False])

In [8]: voxel2_surface.is_filled([v1_points[0]])
---------------------------------------------------------------------------
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/aidrive1/miniconda3/envs/dr/lib/python3.8/site-packages/trimesh/voxel/runlength.py", line 378, in sorted_rle_gather_1d
    index = next(index_iter)
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/aidrive1/miniconda3/envs/dr/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/aidrive1/miniconda3/envs/dr/lib/python3.8/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/data/colin/DRNet/net_utils/eval_det.py", line 350, in eval_det_cls_wrapper_w_mesh
    (rec, prec, ap), (rec_mesh, prec_mesh, ap_mesh) = eval_det_cls_w_mesh(pred, gt, ovthresh, use_07_metric, get_iou_func, get_iou_mesh)
[...]

PR #1339, opened 2021-09-04
fix the same issue #1315

Co1lin:


mikedh:
Thanks for the PR! I wonder if there's a way to unit test for this. No need to change (I'll merge this week) but another option for handling this is passing a default value to `next`:

```
In [6]: a = [1, 2]

In [7]: i = iter(a)

In [8]: next(i, None)
Out[8]: 1

In [9]: next(i, None)
Out[9]: 2

In [10]: next(i, None)

```

PR #1354, opened 2021-09-23
Fix a bug loading color of dae files.

rshanor:
Code was crashing for 4 channel colors.
    
trimesh/exchange/dae.py", line 203, in _parse_node
        len(color_index) * 3, 3)
    ValueError: cannot reshape array of size 10020 into shape (2505,3)


mikedh:
Awesome, thanks for the PR!

PR #1370, opened 2021-10-04
Submesh support for `slice_mesh_plane()`

### LINKED ISSUE #1369: Submesh support for `slice_mesh_plane()`
Hi Mike,

I've run into a use case with the `slice_mesh_plane()` function where I need to slice through only _part_ of the mesh.

One approach would be to take the submesh, slice it, and then rejoin it to the rest of the mesh. However that gets a little messy since it requires some extra bookkeeping of the vertices and the faces. It seems like it would be better, for my application, to provide a way to specify the triangles to which the slicing operation should apply.

I propose adding a `faces_to_slice` parameter to `slice_mesh_plane()` and `.slice_plane()`, which is an optional boolean mask. When the mask is provided, the slicing operation would only apply to the specified faces, with the remaining faces preserved regardless of where they lie in relation to the slicing plane.

What do you think about adding this? Would a PR be welcome?

paulmelnikow:
I've added a test, which is working well! However it doesn't have any assertions. I started writing a couple, but it feels like a lot of code. Curious what you'd suggest there (or if you want to fill some in).

Closes #1369

### DIFF
--- a/trimesh/intersections.py
+++ b/trimesh/intersections.py
@@ -430,6 +430,7 @@ def slice_faces_plane(vertices,
                       faces,
                       plane_normal,
                       plane_origin,
+                      face_index=None,
                       cached_dots=None):
     """
     Slice a mesh (given as a set of faces and vertices) with a plane, returning a
@@ -446,6 +447,9 @@ def slice_faces_plane(vertices,
         Normal vector of plane to intersect with mesh
     plane_origin :  (3,) float
         Point on plane to intersect with mesh
+    face_index : ((m,) int)
+        Indexes of mesh.faces which to slice. When no mask is provided, the
+        default is to slice all faces.
     cached_dots : (n, 3) float
         If an external function has stored dot
         products pass them here to avoid recomputing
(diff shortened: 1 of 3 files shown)


trimesh/intersections.py
```diff
@@ -698,10 +716,14 @@ def slice_mesh_plane(mesh,
...
                                             plane_normal=normal,
                                             plane_origin=origin,
+                                            face_index=face_index,
                                             cached_dots=dots)
 
         # check if cap arg specified
         if cap:
+            if face_index:
+                # This hasn't been implemented yet.
+                raise NotImplementedError("face_index and cap can't be used together")
```
paulmelnikow, line 726: I think this probably could be implemented in the future.

trimesh/intersections.py
```diff
@@ -446,6 +447,9 @@ def slice_faces_plane(vertices,
         Normal vector of plane to intersect with mesh
     plane_origin :  (3,) float
         Point on plane to intersect with mesh
+    face_index : ((m,) int)
+        Indexes of mesh.faces which to slice. When no mask is provided, the
+        default is to slice all faces.
```
paulmelnikow, line 452: Adding this parameter to the helper seems like the simplest way to accomplish this.

mikedh:

PR #1381, opened 2021-10-17
Fix AttributeError around self._line_offset

wkentaro:
To avoid AttributeError: 'SceneViewer' object has no attribute '_line_offset'


mikedh:
Thanks for the PR! Seems reasonable to always populate that.

PR #1400, opened 2021-11-02
moves .lower() in texture file naming for consistency within material

### LINKED ISSUE #1396: texture image name doesn't match material name
If a material is named eg. `MaterialA`, then when exported as obj the mtl file will be named `MaterialA.mtl` but the corresponding image texture will be named `materiala.png`. This comes from the `.lower()` here:

https://github.com/mikedh/trimesh/blob/5c8891defc1967cfcb9fd7c0c908b4f666364091/trimesh/visual/material.py#L128

Is there a reason for using `.lower()` here? It causes some unwanted discrepancy.

kampelmuehler:
resolves #1396

### DIFF
--- a/trimesh/visual/material.py
+++ b/trimesh/visual/material.py
@@ -124,8 +124,7 @@ class SimpleMaterial(Material):
             # what is the name of the export image to save
             if image_type is None:
                 image_type = 'png'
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! I'll merge before next release.

PR #1405, opened 2021-11-03
[WIP] face_subset

### LINKED ISSUE #1403: incorrect uv map after mesh smoothed
I'm using the excellent implementation of `Trimesh.smoothed` to remove some ugly artefacts in my assets, however it seems that the uv map is either lost (error in concatenating visuals if images are None) or gives the wrong UVs.

- w.r.t. the error in concatenating visuals, which occurs here https://github.com/mikedh/trimesh/blob/main/trimesh/path/packing.py#L429
where None image gives no size, therefore it fails (and continues without Texture materials). Is it possible to ignore the fact that some images are None? the uvs should could be still concatenated.

- w.r.t. to the wrong UVS in the end for the example below, i have a mesh with 2098 vertices and 191015 uvs 
my 2 cents go on this line https://github.com/mikedh/trimesh/blob/main/trimesh/util.py#L1590, which could instead select the right uvs. However i can't see an obvious way in the short time i've looked at this

any hints?

code below:

```
import numpy as np
import trimesh
from PIL import Image

mesh = trimesh.load("./teapot.obj").unwrap(Image.fromarray(np.empty([512, 512, 3], dtype=np.uint8)))

mesh = mesh.smoothed()
print("vertices:", len(mesh.vertices))
print("uvs:", len(mesh.visual.uv))
```

> vertices: 2098
> uvs: 191015

[teapot.zip](https://github.com/mikedh/trimesh/files/7468978/teapot.zip)

michele-arrival:
fixes #1403?

### DIFF
--- a/tests/test_smooth.py
+++ b/tests/test_smooth.py
@@ -1,0 +1,25 @@
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+
+class SmoothTest(g.unittest.TestCase):
+
+    def test_smooth(self):
+        m = g.get_mesh('chair_model.obj', force='mesh')
+        s = m.smoothed()
+
+        ori = g.np.hstack((m.visual.uv, m.vertices))
+        check = g.np.hstack((s.visual.uv, s.vertices))
+
+        tree = g.spatial.cKDTree(ori)
+        distances, index = tree.query(check, k=1)
+        assert distances.max() < 1e-8
+
+        # g.texture_equal(m, s)
+
+
+if __name__ == '__main__':
+    g.trimesh.util.attach_to_log()
+    g.unittest.main()
--- a/tests/test_visual.py
+++ b/tests/test_visual.py
@@ -1,0 +1,59 @@
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+
+class VisualTest(g.unittest.TestCase):
+
+    def test_face_subset_texture_visuals(self):
+        m = g.get_mesh('fuze.obj', force='mesh')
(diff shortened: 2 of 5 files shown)


mikedh:
Thanks for the PR! I'll merge before next release. One question, is there any way to unit test this? Just spitballing, maybe something that checks to see every original position-uv has a matching value in the smoothed mesh. I suppose smoothing will add duplicates at the boundary, so every original position-uv should have one or more exactly identical position-uv pairs in the new mesh? It's been a while since I looked at this though haha. 

```
m = trimesh.load(...)
s = m.smoothed()

ori = np.hstack((m.visual.uv, m.vertices))
check = np.hstack((s.visual.uv, s.vertices))

tree = scipy.spatial.cKDTree(ori)
distances, index = tree.query(check, k=1)
assert distances.max() < 1e-8 
```

michele-arrival:
this works right out of the box:

```
class SmoothTest(g.unittest.TestCase):

    def test_smooth(self):
        m = g.get_mesh('chair_model.obj', force='mesh')
        s = m.smoothed()

        g.texture_equal(m, s)
```

just wondering if it needs its own test file or it can go under some other test suite/file?

michele-arrival:
there you have 2 kinds of tests. One for smooth, and one for face_subset (TextureVisuals + ColorVisuals).

Btw, Why is VertexColor not exposed in https://github.com/mikedh/trimesh/blob/main/trimesh/visual/__init__.py#L29 ?
same test as above for VertexColor is commented out for now because it can't see the class

Issue #1422, opened 2021-12-06
Cache incorrect for `scene.graph.nodes` after using `scene.graph.transforms.remove_node`


clemense:
Hi!
After using `scene.graph.transforms.remove_node(..)`, the value of `scene.graph.nodes` is incorrect / not updated. I assume this is due to the caching mechanism. How can I explicitly invalidate the cache such that the function gets the most recent value from `self.transforms.nodes`?

Thanks!

mikedh:
Oh that's odd, it looks like the `EnforcedForest` is clearing *it's* cache:

https://github.com/mikedh/trimesh/blob/46f48f1df0bbba304b1ef9673417d3ede5256172/trimesh/scene/transforms.py#L564-L565

But the upper level `SceneGraph` is using `graph.modified` as it's cache-key, so it isn't checking to see if anyone else modified the EnforcedForest:
https://github.com/mikedh/trimesh/blob/46f48f1df0bbba304b1ef9673417d3ede5256172/trimesh/scene/transforms.py#L150-L156

Seems like to fix we could add a `modified`/`_current` flag on changes, or add hashing of some form to EnforcedForest and check that from the scenegraph. PR's welcome! I have an outstanding release I need to get to at some point soon.

clemense:
Not sure if this is what you thought about, I just copied the scheme the scene graph uses. Let me know about any improvements.

mikedh:
Thanks for the fix! Also amusingly there was a hardcoded test which incorrectly asserted that the number of nodes didn't change after running `remove_node` 👀

PR #1424, opened 2021-12-07
Add cache modifier for enforced forest

clemense:
Addresses #1422


mikedh:
Thanks for the PR! Added a quick test, and should be released now.

PR #1429, opened 2021-12-13
Fix and test for reflection

### LINKED ISSUE #1417: Rigid registration returns a reflected mesh even if the optional parameter `reflection`, of the `procrustes` function, is set to `False`
Hi!
As usual, awesome job! 🥇

I am trying to perform a rigid registration of a mesh with respect to a target mesh (the `other_mesh`) using the `mesh_other` function. However, the rototranslation matrix that is returned from the function is such that to reflect the mesh.

I have tried to change the default value of the `reflection` parameter of the `procrustes` function to `False` but I still get a reflected mesh.

I am using Trimesh  3.9.25.

Unfortunately, I cannot share this mesh, but I am wondering If you have ever experienced that problem and how I can prevent the reflection of the mesh.

Thanks!

### LINKED ISSUE #1438: wireframe flag causes segfault
Hi there, 

I was looking for a way to enable wireframe mode programmatically, found https://github.com/mikedh/trimesh/issues/1365 which describes <trimesh object>.show( flags={'wireframe'=True} ).  When I use that, it generates a segfault.

Weirdly, if I just do <trimesh object>.show() I get the fully rendered object, I can hit 'w' to go into wireframe mode with no crashes.  I also see even weirder behavior if I first run a show() followed by a show(flags ... ).  Here's a minimal reproducer that exhibits the three cases I explored.

This is with trimesh==3.9.36 freshly installed from pip.  In case it's important, I'm on a mac with pyglet==1.5.21.  I updated pyglet to the latest to verify it wasn't an old version of pyglet which might have been causing the issue.

Josh

```
import numpy as np
import trimesh as tm

box = tm.creation.box()

# Set this value to 1, 2, or 3 to exercise one of the cases below
case = 2 

if case == 1:
    # This case doesn't segfault, I can hit 'w' within the viewer and get wireframe mode
    box.show( )

if case == 2:
    # This case causes a segfault at the command line
    box.show( flags={'wireframe':True} )

if case == 3:
    # In this case, if I first run a normal "show" followed by the wireframe
    # flag, I see the normal behavior on the first pyglet window that pops up..
    box.show( )

    # And after closing the first pyglet window, a second pyglet window pops up
    # during this show call.  The initial display/rendering is not wireframe,
    # but if I left-click the mouse in the window, it generates a refresh event
    # (it looks like) and the window is suddenly then in wireframe mode.
    box.show( flags={'wireframe':True} )
```

mikedh:
- fix #1417 
- mitigate or possibly fix #1438


fabricecarles:
I found again issue of reflexion in trimesh 3.20.1
```
transform_to_occ, cost = trimesh.registration.mesh_other(target_mesh, source_mesh, samples=500, scale=False, icp_first=10, icp_final=50, reflection=False)
```
output can be flipped (or not) even if reflection=False, this is sample dependant

PR #1437, opened 2021-12-15
Update path.py

thachdo:
Correct importing `trimesh.path` bug in python 2


mikedh:
Hey, yeah the `from . import` import style is the preferred method for modules and works in Python 2, I think this is unrelated to the issue you were seeing in the render import.

PR #1452, opened 2022-01-07
Fix bug with negative indices in OBJ

friggog:
Apologies, realized that I missed the same thing somewhere else

Not totally sure how to catch this with the test?


mikedh:
No worries, thanks for the followup! Hmm yeah I guess to test we'd need an example model that had normals and texture and negative indices? We might be able to just tweak `fuze.obj` to change a few indices to negative, but it may not be worth the trouble.

PR #1456, opened 2022-01-12
fix revolve strict check

khameeteman:
Code falls into strict check even if `tol.strict == False`


mikedh:
Great catch, thanks for the PR!

PR #1469, opened 2022-01-26
obj header as parameter

JWDobken:
this change adds the OBJ header as an optional parameter... to avoid clumsy post-processing


mikedh:
Looks good to me, thanks for the PR!

PR #1476, opened 2022-01-31
Configure build with pyproject.toml

EpicWink:
Also known as PEP 517 (and PEP 518), explicitly specify build backend and requirements. Also use `build` in CI to build package for upload to PyPI.

See [the pip docs on pyproject.toml](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/) for more details.


mikedh:
Looks great, thanks for the fix!

PR #1477, opened 2022-01-31
Remove setuptools unused dependency

### LINKED ISSUE #1475: Remove setuptools requirement
`setuptools` is listed in the (default) requirements for Trimesh, but isn't used anywhere. In the `setup.py`, the comment says it's for packaging, but this is not a runtime thing anyway

EpicWink:
Closes #1475


mikedh:
Awesome thanks for the fixes! I should probably also migrate most of `setup.py` to `setup.cfg` at some point as well.

PR #1479, opened 2022-02-01
Organize faces by meaningful material names

RoyJames:
I needed this feature to access faces by material, so sharing it in case others need it too. It makes more sense to use material names as group index when the `group_material` flag is True. This does not change the behavior of the code when that flag is False.


mikedh:
Thanks for the PR! Yeah it definitely makes more sense to use the material name if we are grouping by it, and the material names are all guaranteed to be unique in this instance, so it doesn't need the "make sure this name isn't already used" check? I've definitely had the bug where the names are all the same so you only get the last one haha. 

This looks good and I'll merge probably tomorrow, if you get a chance to add a simple test that would be good (one of the examples might have multiple materials) but if not no worries I'll try to throw together something easy before releasing.

RoyJames:
I can help with the test too. I've used this fork on a bunch of obj files that each has around 10 materials and it worked fine. I'm not sure if you described a bug - if all faces have the same material name, there should just be one material returned to us right? I imagine the new test should be in `test_scene.py`, in which I can try loading a .obj file that has multiple materials and check the right material names are loaded. Does such a test sound legitimate?

mikedh:
>  I imagine the new test should be in test_scene.py, in which I can try loading a .obj file that has multiple materials and check the right material names are loaded. Does such a test sound legitimate?

Yes this sounds perfect! Thanks!

mikedh:

PR #1482, opened 2022-02-03
Triangulation engine handling fixes in library and tests

musicinmybrain:
In `test_path_sweep` and `test_triangulate_plumbing`, use all available triangulation engines (and no unavailable ones). This specifically allows the tests to run when the `earcut` engine is available but the non-free `triangle` engine is not. It also provides coverage of the `earcut` engine even when `triangle` (the implicit default) is *also* available.

This came to light since I just added [mapbox-earcut](https://src.fedoraproject.org/rpms/python-mapbox-earcut) to Fedora Linux—so these tests are no longer skipped—but [triangle](https://pypi.org/project/triangle/) is not packaged because it requires the non-free [Triangle](https://www.cs.cmu.edu/~quake/triangle.html) library.

In fixing the tests to work in this configuration, I found that `trimesh.creation.sweep_polygon` was calling `triangulate_polygon` without passing along keyword parameters, which meant `sweep_polygon(engine=…)` was only partially respected. I fixed this in a separate commit.

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -142,8 +142,9 @@ class CreationTest(g.unittest.TestCase):
         path = g.np.c_[x, y, z]
 
         # Extrude
-        mesh = g.trimesh.creation.sweep_polygon(poly, path)
-        assert mesh.is_volume
+        for engine in self.engines:
+            mesh = g.trimesh.creation.sweep_polygon(poly, path, engine=engine)
+            assert mesh.is_volume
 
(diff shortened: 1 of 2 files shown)


mikedh:
Looks great, thanks for the fix and all the packaging work! Yeah it's too bad triangle doesn't have a more standard OSS license.

musicinmybrain:
Thanks for the quick review!

Issue #1480, opened 2022-02-02
Empty `mesh.visual.uv` creates invalid obj file


clemense:
Hi!
Sometimes I get meshes that have the `mesh.visual.uv` variable but it's empty.  This creates problems during export, where an invalid   obj file is created: The file will contain a `vt\n` line.
One way to resolve this is to add `and mesh.visual.uv.size > 0` to this if-clause:
https://github.com/mikedh/trimesh/blob/4aea783457c06ebe5a457a4da070dd22668db923/trimesh/exchange/obj.py#L852

Any thoughts? If that's ok, I can create a PR.

Thanks!

mikedh:
Ah yeah looks good, PR's welcome! What program barfs on loading those just out of curiosity? Could we run it with subprocess in unit tests to check all of our exports?

clemense:
E.g. Meshlab 1.3.2  and Blender 2.79 won't import those files (a more recent version of Blender has no problems).

PR #1488, opened 2022-02-08
Add additional uv texture test to avoid invalid obj

clemense:
Relates to #1480


mikedh:
Ah looks like it's sometimes None, we might try instead of `hasattr(mesh.visual, 'uv') and mesh.visual.uv.size > 0:`
`getattr` with a default and `np.shape` works on None:
`len(np.shape(getattr(mesh.visual, 'uv', None))) == 2`

mikedh:
Awesome thanks for the fix!

PR #1497, opened 2022-02-15
support faces in 3dxml

### LINKED ISSUE #1493: Extend 3dxml loader to support Face triangles attribute
The current 3dxml implementation only supports the `strips` attribute (which my files unfortunately do not have). [The format](https://www.3ds.com/3ds-passport/downloads/) also allows for `triangles` and `fan` attributes to define the Face.

https://github.com/mikedh/trimesh/blob/e9c6327230b64f176e9e176f58a0f4a3d0334620/trimesh/exchange/threedxml.py#L151

Thanks!

mikedh:
- fix #1493


trimesh/exchange/threedxml.py
```diff
@@ -144,15 +145,19 @@ def load_3DXML(file_obj, *args, **kwargs):
...
+            if 'strips' in faces.attrib:
+                # triangle strips, sequence of arbitrary length lists
+                # np.fromstring is substantially faster than np.array(i.split())
+                # inside the list comprehension
+                strips = [np.fromstring(i, sep=' ', dtype=np.int64)
+                          for i in faces.attrib['strips'].split(',')]
+
+                # convert strips to (m,3) int
+                mesh_faces.append(util.triangle_strips_to_faces(strips))
+            elif 'triangles' in faces.attrib:
```
costerwi, line 157: The [3D XML User Guide 4.3](http://media.3ds.com/3dxml/3dxmldoc/3D_XML_User_Guide_4.3.pdf) has an example on page 63 with both 'strips' _and_ 'triangles' attributes on the same Face element so `if` may be better than `elif` here.

Issue #1220, opened 2021-04-20
Is there a way to sample a surface to a PointCloud with Colors?


sai-cherukumilli-enview:
I currently use `trimesh.sample.sample_surface` method to sample the surface and generate a PointCloud. But this PointCloud is missing the color attributes. 

Is there a way to get the surface color for these points in the PointCloud?

mikedh:
Hey, I would get the face indices from `sample_surface`, and then either take the face colors, or do barycentric interpolation on vertex colors. `sample_surface` is already using barycentric sampling internally, you could probably add a `vertex_attributes` or `return_barycentric` argument to it pretty easily that would make this very straightforward; PR's welcome!

PR #1502, opened 2022-02-23
Add sample surface with color function

nenoNaninu:
This PR is related to issue #1220.
I added a function `sample_surface_with_color` to sample colored points from a textured mesh.

### DIFF
--- a/tests/test_color.py
+++ b/tests/test_color.py
@@ -221,6 +221,27 @@ class VisualTest(g.unittest.TestCase):
 
         g.np.testing.assert_allclose(colors, colors_expected, rtol=0, atol=0)
 
+    def test_uv_to_interpolation_color(self):
+        try:
+            import PIL.Image
+        except ImportError:
+            return
+
+        uv = g.np.array([[0.25, 0.2], [0.4, 0.5]], dtype=float)
+
+        tmp = g.np.arange(32)
+        x, y = g.np.meshgrid(tmp, tmp)
+        z = x + 128
+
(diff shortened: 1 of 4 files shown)


mikedh:
Awesome thanks for the PR! Apologies for the delay reviewing haha. Looks great and thanks for adding a test as well! My only though is couldn't we have color as an option to the original `sample_surface` function to avoid duplicate code block? Something like:
`sample_surface_color` -> `sample_surface(... sample_color=False)`

nenoNaninu:
Of course it is possible, but I thought it would be better to avoid changing the contents of the return value depending on the option, so I implemented it as a separate function. If this is a needless concern, I would like to fix the function to pass option:)

mikedh:
> I thought it would be better to avoid changing the contents of the return value depending on the option

Yeah totally understood, changing return values based on a flag isn't wonderful. On the other hand `sample_surface` is pretty well tested, and duplicating code can make it tough to maintain as bugfixes have to go into multiple places so I generally prefer the flag over duplicating code blocks. And in defense of the `return_x` flag, `numpy` does it so it must be ok 😆

nenoNaninu:
I fixed it!

mikedh:
Looks perfect thanks for the PR and fix!

jiazhou2023:
@mikedh @nenoNaninu Thanks for this PR, is it possible that for some objects/texture it sample the uv correspondence where it is black? Did it happen to you? 

Generating these strange results with black? Screenshot 2023-03-29 at 10 47 37 PM

PR #1503, opened 2022-02-24
Small correction in documentation

Drvanon:


Drvanon:
I saw this and the fix was so easy I thought I'd just create a quick pull request for it.

mikedh:
Awesome thanks for the fix!

PR #1515, opened 2022-03-11
ci: update import path for scipy coo_matrix

mmatl:
Newer version of scipy prints error messages for the older path to `coo_matrix` that is currently used in trimesh. Simple fix in this PR.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.html


mikedh:
I started noticing those too, thanks for the fix!

PR #1528, opened 2022-03-22
projected only handles faces on front side of normal

YuanWenqing:
The normal argument in polygons.projected already determined faces on which side would be projected. Auto side-pick in the method will lead incorrect projection of non-watertight mesh that has a big front side but a small back side.

Actually, an argument to determine whether both sides or only front side may be necessary.


mikedh:
I managed to confuse the heck out of myself with the auto-picking logic. I changed to to an explicit flag: `ignore_sign`: https://github.com/mikedh/trimesh/pull/1560/commits/ac8da9aa0798d62b8d3f656eaa8c30301499c015

PR #1534, opened 2022-03-25
apad/rpad in polygons.projected

YuanWenqing:
In some cases, we want an absolute padding size in projection. The old `pad` parameter is a relative padding size and the final padding size applied is not predictable.


mikedh:
Good idea and thanks for the PR! I may change it to exactly match numpy behavior where it is `atol + (rtol * scale)` rather than atol vs rtol, and then if you want only atol you can explicitly set rtol to `None`.

Issue #1517, opened 2022-03-14
Setting seed for trimesh.util.unique_id()


clemense:
Hi!
I am trying to generate a scene in a reproducible/deterministic fashion, i.e., also with the geometry node names being the same between runs. Unfortunately, I couldn't figure out how to make `trimesh.util.unique_id()` deterministic; setting the seed is not enough since it also depends on the current timestamp. Any hints?

Thanks!

mikedh:
Ah yeah that's not ideal. I've always just made a new graph and either passed it to the `Scene` constructor, or possibly `scene.graph.from_edgelist`. You can see [what the scene loader is doing too](https://github.com/mikedh/trimesh/blob/main/trimesh/exchange/load.py#L421-L427), or if you have a better idea for a deterministic creation of node names that would be good too!

clemense:
I see. I'm just not sure how I can rename all nodes and geometry in a deterministic fashion, since they are coming from loading files, and I have no assumption about the particular structure of the graph.

It would be possible to make the `unique_id()` method deterministic if it wouldn't have the time-incrementing (`now()`) part in the beginning. What's the reason for that in the first place?  Because if it's only about having similar ID's for nodes that were added to the scene in sequence, the same could be accomplished by e.g. having an alphanumeric-incrementing part in the beginning.  Then the entire method could be seeded and made deterministic.

mikedh:
Ah yeah `unique_id` is basically using UUID logic (probably read a blog post [like this one](https://eager.io/blog/how-long-does-an-id-need-to-be/)) where adding a timestamp reduces the likelihood of collisions. Totally on-board with making this deterministic. Could we do something like this?
```
-> start with a hash of something
-> if hash is a node already, increment the hash/hash-seed until it produces a value that isn't in the graph
```

clemense:
Alright, here's my proposed new version of unique_id with a test. Let men know if that's too radical.

(btw: there are some tests that are very hard to satisfy, e.g. the format test fails if `now` is not used in util.py anymore but there are a lot of external files that use it, so simply removing it also fails)

PR #1536, opened 2022-03-28
Change unique_id() implementation. And add a test.

clemense:
Addresses #1517


mikedh:
Yeah looks good! I might change it to `UUID(..).hex` to avoid the arbitrary dashes but yeah I agree reimplementing UUID doesn't make a ton of sense. Do we even need the denylist? Seems like collisions are pretty freaking unlikely.

Issue #1518, opened 2022-03-14
Export mtl for obj without texture doesn't work


clemense:
Hi!

I am very convinced that this recent change https://github.com/mikedh/trimesh/commit/5abdb9fcdf0f860667980bf1fa27338d1a86a19a#diff-b0fd95d2b4cd76b3a3ba885804d392e28913e1e78423883a0881cc88054160f8 (and the commit before) created a new problem: If a mesh with material (but without texture) is exported to obj, no mtl will be created. This also relates to what I observed here: #1499 

I propose the following changes to fix this:
1. Return to the original condition: `if include_texture and hasattr(mesh.visual, 'uv'):` in `obj.py`.
2. A few lines afterwards add the condition: `if len(np.shape(getattr(mesh.visual, 'uv', None))) == 2:` to avoid writing an empty `vt` field in the obj file (which was the motivation for the original commit, see above).

Let me know if an MR is needed. Thanks!

mikedh:
I'm not totally sure I understand this issue but PR's with a test very welcome! The test matrix/correct behavior looks like this (but one of these conditions isn't true?):

- mesh without UV coordinates, but has a material: should create MTL but no VT
- mesh without UV coordinates, no material: no MTL, no VT
- mesh *with* UV coordinates, no material: no MTL, VT in file
- mesh with UV but with material: MTL and VT in file

clemense:
Correct, the following condition is not true:
> mesh without UV coordinates, but has a material: should create MTL but no VT

This is because https://github.com/mikedh/trimesh/blob/4671ead1432819e8da2f595c4636df429ee3acaf/trimesh/exchange/obj.py#L854 in this case is False (it used to be True before https://github.com/mikedh/trimesh/pull/1488 because mesh without UV coordinates but with material has an empty `uv` variable).
Because of that `tex_data` is None, which in turn will not enter the material writer https://github.com/mikedh/trimesh/blob/4671ead1432819e8da2f595c4636df429ee3acaf/trimesh/exchange/obj.py#L907

I can write a test, basically it would look like this:
```
sphere = trimesh.creation.uv_sphere()
sphere.visual = trimesh.visual.TextureVisuals(uv=None, material=trimesh.visual.material.empty_material())
output = sphere.export('sphere.obj')
assert('usemtl' in output)
```

mikedh:
Oh thanks, I understand now. Yeah PR's welcome!

PR #1537, opened 2022-03-28
Export MTL file for obj meshes even if they don't have uv coordinates

clemense:
Addresses #1518

### DIFF
--- a/tests/test_obj.py
+++ b/tests/test_obj.py
@@ -267,6 +267,12 @@ class OBJTest(g.unittest.TestCase):
             mesh.export(file_type='obj'), file_type='obj')
         assert g.np.isclose(mesh.area, rec.area)
 
+    def test_no_uv_but_mtl(self):
+        sphere = g.trimesh.creation.uv_sphere()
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the fix and test!

larsgrobe:
Dear all,

should materials be exported with scenes containing no uv data? I could not find a minimal working example to do this, but simply writing by
`trimesh.exchange.export.export_scene(scene, 'scene.obj', mtl_name = 'scene.mtl')`
gives me a material file with only one entry.

With glb export, the material names and properties are maintained.

Best, Lars.

larsgrobe:
So it is apparently still the case that I need to have uv coordinates (can be produced by unwarp() ) to get material properties exported to a mtl file. I have this now implemented by translating a trimesh.visual.material.PBRMaterial with .to_simple(), and the result with .to_obj(name='my_name'). However, the name 'my_name' shows not up in the mtl or obj files - here (unfortunately) the default material[0-] remain. This must be caused by the exporter, since printing the return values from .to_obj() gives the desired material name.

PR #1542, opened 2022-03-31
Add description for the repair argument in submesh

SergeiShumilin:


mikedh:
Looks good thanks for the fix!

Issue #1310, opened 2021-08-05
Question: Does trimesh do any kind of transformation on coordinates when loading an stl


Benjamin-Fouquet:
Hello,

When loading a stl file in world coordinate from 3D slicer, I do not find "back" the same points when accessing mesh.vertices. It seems there is some kind of coordinate transformation that I fail to understand, could you help me on that ?

Additional question, is there a way to ensure I stay in world coordinate system using trimesh?

mikedh:
Yeah by default it merges vertices and removes NaN/inf values, you can turn that it off by loading with `mesh = trimesh.load(file_name, process=False)`.

With regard to 'world coordinates' STL's don't have transforms so it should just be an array of points unless you create a `trimesh.Scene` to hold the mesh.

whateverforever:
Currently having the same issue. Blender and meshlab agree on the vertex ordering, but trimesh has a different one. It seems blender and meshlab go vertex by vertex in the file, only loading a new one, if it hasn't already been loaded. trimesh seems to load all vertices as-is (i.e. almost all of them are duplicates) and merges them later on in a second step (losing the ordering). Did I get that correctly @mikedh ?

PR #1546, opened 2022-04-02
Keeping STL Vertex Order

whateverforever:
Hey Mike,

thank you for this fantastic library. As mentioned by another user in #1310, STL vertex ordering can be important to some use cases. In this PR, I have implemented a `unique_ordered` function that also returns the `inverse` list, which means we can use it in `merge_vertices`.

I have attached some timings, with one of the test models, and a subdivided suzanne model that has ~30k vertices. What do you think about this?

```python
In [2]: import trimesh_fork.trimesh as tmf

In [3]: import trimesh as tm

In [4]: import numpy as np

In [5]: data = np.random.random((500_000, 3))

In [9]: %timeit -n 10 tm.grouping.unique_rows(data)
155 ms ± 5.79 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [10]: %timeit -n 10 tmf.grouping.unique_rows(data)
276 ms ± 8.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [12]: %timeit -n 10 tm.load("./trimesh_fork/models/featuretype.STL")
4.35 ms ± 1.61 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [13]: %timeit -n 10 tmf.load("./trimesh_fork/models/featuretype.STL")
4.02 ms ± 320 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [14]: %timeit -n 10 tm.load("...snip.../suzanne-30000v.stl")
66 ms ± 611 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [15]: %timeit -n 10 tmf.load("...snip.../suzanne-30000v.stl")
69.8 ms ± 1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

...

In [3]: %timeit -n 10 tm.load("...snip.../suzanne-500_000v.stl")
1.69 s ± 47 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit -n 10 tmf.load("...snip.../suzanne-500_000v.stl")
1.8 s ± 36.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

```

### DIFF
--- a/tests/test_grouping.py
+++ b/tests/test_grouping.py
@@ -287,7 +287,7 @@ class GroupTests(g.unittest.TestCase):
         # shouldn't error because there are no vertices
         # even though faces are wrong
         mesh.merge_vertices()
-    
+
     def test_unique_ordered(self):
         a = g.np.array([9, 9, 9, 8, 8, 7, 7, 6,
                         6, 5, 5, 4, 4, 3, 3, 0,
@@ -299,7 +299,7 @@ class GroupTests(g.unittest.TestCase):
             return_index=True,
             return_inverse=True
         )
-        
+
         # indices are increasing, because we kept original order
         assert (g.np.diff(ind) > 0).all()
         # we can reconstruct original data
--- a/tests/test_grouping.py
+++ b/tests/test_grouping.py
@@ -287,6 +287,23 @@ class GroupTests(g.unittest.TestCase):
         # shouldn't error because there are no vertices
         # even though faces are wrong
         mesh.merge_vertices()
+    
+    def test_unique_ordered(self):
+        a = g.np.array([9, 9, 9, 8, 8, 7, 7, 6,
+                        6, 5, 5, 4, 4, 3, 3, 0,
+                        2, 1, 1, 0, 0, -1, -1, -1],
+                       dtype=g.np.int64)
+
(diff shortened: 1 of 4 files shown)


mikedh:
Awesome thanks for the PR! I'll try to dig into this next week, my only concern is with performance so thanks for also doing the benchmarks! If it's in the same general performance neighborhood my inclination would be to remove the `keep_order` flags and just always return an ordered result (as it seems like the only reason you *wouldn't* want that is if it's way slower)? 

Correct me if I'm wrong but my read of those benchmarks is that `unique_rows` gets about 40% slower and loading stays roughly the same as the merge is pretty fast? That seems close enough that we can probably tweak it to be basically on-par through profiling.

whateverforever:
Yes, I think you are absolutely right. I updated my comment with a benchmark on a 500k vertices suzanne, and the difference in loading times is very small, compared to the raw 500k rows `unique_rows` call. If we could tweak the performance to be on par, I'd be absolutely delighted, just not sure how to do it

mikedh:
Thanks for the PR!

Issue #170, opened 2018-05-17
Documentation


norheim:
There used to be extensive documentation on the API, and this has for some reason disappeared. Would appreciate if it was brought back!

mikedh:
Yeah for some reason (probably related to `numpydoc` 🙄 ) the module reference stopped building. It's still building the basic docs on [readthedocs](https://trimesh.readthedocs.io/en/latest/index.html) but I know that's less useful. 

I'll try to poke at it in the next couple weeks, or PR's are always welcome 😄

norheim:
That would be amazing; just having the auto generated documentation on methods to search through with readthedocs used to be great :) Might not be able to do a PR, but will keep checking back ;)

mikedh:
Still needs some more work, but it's building and deploying to `gh-pages` via travis. I switched from readthedocs as it was pretty hard to debug their build process:
https://mikedh.github.io/trimesh

mikedh:
Closing as docs are now building with travis and deploying to https://trimsh.org/

PR #1552, opened 2022-04-05
Add support for exporting to 3mf.

luisbc92:


mikedh:
Awesome thanks for the PR! 2 quick thoughts, there's an f-string that the oldest Python we support doesn't like:
```
   File "/opt/hostedtoolcache/Python/3.5.10/x64/lib/python3.5/site-packages/trimesh/exchange/threemf.py", line 264
    "item", {"objectid": str(i + 1), f"{{{nsmap['p']}}}UUID": str(uuid.uuid4())}, nsmap=nsmap
```
Also, I wonder is there any way to batch the looping through faces/vertices? Creating those objects in a loop is going to get pretty slow, maybe something like:
 `parse_fragment('<vertex x={} y={} z={}/>'*len(mesh.vertices)).format(*mesh.vertices.flatten())`?

```
                            with xf.element("vertices"):
                                for v1, v2, v3 in m.vertices:
                                    xf.write(etree.Element("vertex", x=str(v1), y=str(v2), z=str(v3)))
                            with xf.element("triangles"):
                                for f1, f2, f3 in m.faces:
                                    xf.write(etree.Element("triangle", v1=str(f1), v2=str(f2), v3=str(f3)))
```

luisbc92:
You are right! I didn't consider the supported Python versions, easy fix! I will make sure to run testing on my end to catch these small issues first.

Regarding the second issue, the barrier I have run into is when dumping large models. Generating that huge XML string utilizes significant amounts of memory, which is why I'm also streaming node by node to the file. (However, I'm just realizing that I'm still running into this issue anyway because we are buffering to BytesIO; oops... I can still fix this by opening a filehandle on the generated ZipFile so that we compress as we write.)

An in-between solution might be to generate the nodes and write them in batches. Does that sound good?

luisbc92:
Ok, this last commit implements both changes. It creates a ZipFile first, then opens files within it to compress data as we go. The vertices and faces are batched according to **batch_size**.

I played around with a few settings from 128 up to 128k and there did not seem to be much of a difference after 4096, so I settled on that.

mikedh:
Awesome thanks for the updates! One final point I promise 😄 don't `build` tags support a transform as implemented in `threemf._attrib_to_transform`? If we exported those would we be able to fully round-trip a scene via 3MF? I'm a fan of unit tests like:

```
scene = trimesh.load('models/cycloidal.3DXML

rexp = trimesh.load(file_obj=trimesh.util.wrap_as_stream(scene.export(file_type='3mf')), file_type='3mf')

assert set(rexp.geometry.keys()) == set(scene.geometry.keys())
assert set(rexp.graph.node_names == set(scene.graph.node_names)

```

If it's not in scope for you happy to merge regardless, thanks again for all the work on this!

luisbc92:
I think it does. It should be pretty easy to implement. I'll confess that I had only implemented enough for my use case, but I'm happy to work on adding this too.

Btw, thanks for being so responsive! Trimesh has been of huge help to my project.

luisbc92:
I just noticed there would be an issue when comparing with a unit test like the one you propose. The following (line #170) on threemf.py enforces a unique name to nodes added to the scene when loading a 3mf:

```python3
# frame names unique
name = id_name[last] + util.unique_id()
```

Maybe we could change this to only add the ID if there will be a collision? Or possibly compare them a different way.

Anyway, just a heads up.

mikedh:
> Maybe we could change this to only add the ID if there will be a collision? Or possibly compare them a different way.

Yeah I think that would be better! This might have been there as one of my example files had collisions, but I think behaving deterministically for well-constructed files would be an improvement.

luisbc92:
This last commit should store 3MFs while preserving object hierarchy as described by the scene graph.

While working on it, I realized that the leaf nodes are not actually stored in the 3MF as they are only a reference to the actual geometry. This is why we need to generate ids for the graph nodes.

This could be fixed, but it requires all importers that behave like this to do it consistently (3DXML for example), or maybe find a different way to compare scenes and validate equality.

mikedh:
> While working on it, I realized that the leaf nodes are not actually stored in the 3MF as they are only a reference to the actual geometry. This is why we need to generate ids for the graph nodes.

Ah yeah that makes sense, it's been a while since I looked at 3MF. No worries about the multiple commits or linters, I usually take care of that when doing a release anyway. Thanks for all the work and followup! 

I'm happy to merge whenever you're ready and prep for release.

luisbc92:
I think it should be ready.

I might work on getting repeatable load and export when I get some more free time.

Thanks for your help!

mikedh:
Thanks for all your work on this!

luisbc92:
Mike, I just pushed one final change a few seconds ago.

Even though the 3mf spec does not require this, a few tools expect sequential model IDs so I changed the method I generate IDs to a list instead of hashing names.

Hopefully you can still get it on the release.

mikedh:
Sweet! I'll PR it into my release branch (https://github.com/mikedh/trimesh/pull/1561) I just started testing this.

luisbc92:
Oh good! I'm glad this is working well!

PR #1553, opened 2022-04-05
add return_index of subdivided old faces

steo85it:
Hi! Thanks for this very useful library. 
I needed for a project to relate the indexes of the old and new faces after a remesh.subdivide and implemented it. 
I tried to be minimalist with my changes, but please have a look and see if it is: 1. correct; 2. potentially interesting for other users; 3. not "dangerous".
Also, please let me know if this is not the correct branch or procedure for a PR.

Description: Add an option to return a dict providing for each old face index the set of newly derived face indexes (only for faces undergoing subdivision).

### DIFF
--- a/trimesh/remesh.py
+++ b/trimesh/remesh.py
@@ -15,7 +15,8 @@ from .geometry import faces_to_edges
 def subdivide(vertices,
               faces,
               face_index=None,
-              vertex_attributes=None):
+              vertex_attributes=None,
+              return_index=False):
     """
     Subdivide a mesh into smaller triangles.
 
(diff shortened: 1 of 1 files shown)


mikedh:
Awesome thanks for the PR! Looks good, my only "put" is could we have it return a numpy array version of this so it could be used with advanced indexing? I think since subdivision is uniform it could be a `(len(mesh.faces), 4) int` or `len(face_index), 4) int` right?

steo85it:
It could be `(len(face_index), 4) int`, with the first column indicating both the old face index and the first of the new ones (and with `face_index == mesh.faces` if a list of faces isn't indicated in the input). It might be a bit less clear than the dict, but maybe easier to use directly as input.

PR #1558, opened 2022-04-11
fix quadratic node_name lookup

Rennzie:
Fixes an issue when using `Scene.add_geometry()`. Because the set of node names is created each time `graph.nodes` (line 183 in scene.py) is called, the lookup speed slows down quadratically. This makes adding thousands of geometries expensive - see graph.
Screenshot 2022-04-11 at 09 00 43

The fix is to keep track of the nodes names continually rather that recreating the set each time.

### DIFF
--- a/trimesh/scene/transforms.py
+++ b/trimesh/scene/transforms.py
@@ -497,6 +497,7 @@ class EnforcedForest(object):
         self.edge_data = collections.defaultdict(dict)
         # {u: data}
         self.node_data = collections.defaultdict(dict)
+        self.node_names = set()
 
         # if multiple calls are made for the same path
         # but the connectivity hasn't changed return cached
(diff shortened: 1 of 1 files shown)


Rennzie:
No probs. 

Would using `__in__` from `node_data.keys()` still be constant time? I'm not sure who that is implemented in python. If it is then going that route would make more sense is think.

mikedh:
> Would using __in__ from node_data.keys() still be constant time? I'm not sure how that is implemented in python. If it is then going that route would make more sense is think.

I'm guessing the quadratic time is from all the copying of the keys into a set. In Python2 this will still be copying every time, but in Python 3 [the speed difference](https://stackoverflow.com/questions/31738912/python-is-key-in-dict-different-faster-than-key-in-dict-keys/31739052) was fixed with their "set-like-object." I don't think we have to worry about Python 2 too much these days.

PR #1564, opened 2022-04-24
bugfix: perpendicular vectors in merge_colinear

YuanWenqing:
The perpendicular vectors in `merge_colinear` miss changing sign of one axis, thus not perpendicular to the direction vectors.

Example to reproduce:
~~~python
num=100
dists=np.linspace(0,1000,num=num)
direction=trimesh.unitize([1,1])
points=direction*dists.reshape((-1,1))
print(points.shape)

merged=trimesh.path.simplify.merge_colinear(points,scale=1000)
print(merged.shape)
~~~
output: 
~~~
points: (100, 2)
merged: (100, 2)
~~~

### DIFF
--- a/tests/test_simplify.py
+++ b/tests/test_simplify.py
@@ -24,7 +24,7 @@ class SimplifyTest(g.unittest.TestCase):
             # the simplified version shouldn't have lost area
             assert g.np.allclose(path.area,
                                  simplified.area,
-                                 rtol=1e-3)
+                                 rtol=1e-2)
 
             # see if we fit as many arcs as existed in the original drawing
             new_count = sum(int(type(i).__name__ == 'Arc')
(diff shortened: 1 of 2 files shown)


YuanWenqing:
Additionally, `scale` arg in `merge_colinear` is kind of confusing and hard to explain the results. 
In case of 3 points A B C(see figure below), `projection_ratio` is actually max `sin` value between the angles BAC and BCA.
In my opinion, it's better to rename `scale` to `min_angle` or `min_sin`, thus used to calculate threshold for filtering `projection_ratio`.

image

mikedh:
Thanks for the PR! Would you mind adding a unit test ideally that failed before and passes with this fix?

YuanWenqing:
Broken by `tests.test_simplify.SimplifyTest.test_simplify` seems to be just a threshold issue.
Difference of simplified path by both old and new version is like this:
> red: original, blue: simplified by old version, yellow: simplified by new version

original path
image
compare original and simplified by old version:
image
compare original and simplified by new version:
image
compare simplified by old and new version:
image

By old version, changed area is not bigger than 0.1%. But by new one, changed area is not bigger than 1%.
Why is `rtol=1e-3` in the old version? I suppose it's just a value jumping out of brain, right?

mikedh:
Awesome the new behavior looks like a big improvement, thanks for the PR!

> Why is rtol=1e-3 in the old version?  
Yeah just chosen arbitrarily.

PR #1571, opened 2022-04-27
refine radial_sort by specify position of clock 12

YuanWenqing:

PR #1568, opened 2022-04-27
Add test and fix for preserving colors when adding point clouds

clemense:
This fixes the problem with:
```
pc1 = trimesh.PointCloud(np.random.rand(10, 3), colors=[[255, 0, 0, 255]] * 10)
pc2 = trimesh.PointCloud(np.random.rand(10, 3), colors=[[0, 255, 0, 255]] * 10)
(pc1 + pc2).show()   # this currently won't show any colored point clouds
```

PR #1580, opened 2022-05-02
Release: Scene Fixes

mikedh:
- release #1564 
- release #1568 
- release #1574
- release #1575  
- release #1571


trimesh/points.py
```diff
@@ -110,26 +111,33 @@ def radial_sort(points,
...
+        axis1 = np.cross(axis0, normal)
+    vectors = points - origin
     # calculate the angles of the points on the axis
-    angles = np.arctan2(pr0, pr1)
-
+    angles = np.arctan2(np.dot(vectors, axis0),
+                        np.dot(vectors, axis1))
     # return the points sorted by angle
-    return points[[np.argsort(angles)]]
+    return points[angles.argsort()[::-1]]
```
YuanWenqing, line 140: Sorting points as `angles` descending is not corresponding to the doc "Vector to specify position of 12 in clock viewing in direction of normal".

Is this a typo?

mikedh: Yeah I just changed that for the unit test so the result would match a theta `np.linspace(0, np.pi*2, 1000)` but don't have a strong opinion (increasing vs decreasing). Happy to match behavior of the docstring (descending theta) or change the docstring to "increasing" if you have a preference.

YuanWenqing: Changing the docstring seems more reasonable: #1582

mikedh: Thanks for the fix!

PR #1587, opened 2022-05-16
Fix: change early stop part to loop continue instead of return.

HyeonseoNam:
Hi, 
`slice_mesh_plane` seems to have been written considering the multi-plane case.
However, early stopping part(`return`) is for only one-plane case.
To support the multi-plane case with the for loop, I changed `return` to `continue`

### DIFF
--- a/trimesh/intersections.py
+++ b/trimesh/intersections.py
@@ -735,8 +735,7 @@ def slice_mesh_plane(mesh,
                 # if path is None it means this plane didn't
                 # intersect anything so we can exit early without
                 # doing anything to cap the result
-
-                return Trimesh(vertices=vertices, faces=faces, **kwargs)
+                continue
             # transform Path3D onto XY plane for triangulation
             on_plane, to_3D = path.to_planar()
             # triangulate each closed region of 2D cap


HyeonseoNam:
Additionally, this example code will show the failure case clearly.
```python
np.random.seed(5)
red_box = trimesh.creation.box(extents=[10, 10, 10], transform=translation_matrix((5,-5,0)), face_colors=[255, 0, 0])
green_box = trimesh.creation.box(extents=[10, 10, 10], transform=random_rotation_matrix(), face_colors=[0, 255, 0])

# intersection of (green_box, red_box)
sliced = trimesh.intersections.slice_mesh_plane(
    mesh=green_box,
    plane_normal=-red_box.face_normals,
    plane_origin=red_box.vertices[red_box.faces[:, 1]],
    cap=True)

# visualize
trimesh.Scene([red_box, green_box, sliced.apply_scale(1.1)]).show() # scale the result for visualize
```
and `red_box.face_normals` is like this.
```
array([[-1.,  0.,  0.], 
       [ 0., -1.,  0.], # bottom plane of red box, out of green box. slice_mesh_plane returns at this point.
       [-1.,  0.,  0.],
       [ 0.,  0., -1.],
       [ 0.,  0.,  1.],
       [ 0., -1.,  0.],
       [ 0.,  0.,  1.],
       [ 0.,  1.,  0.],
       [ 0.,  0., -1.],
       [ 0.,  1.,  0.],
       [ 1.,  0.,  0.],
       [ 1.,  0.,  0.]])
```
image
(gray is the slightly scaled sliced result)
In this case, planes(normals, origins) of red_box pass to the `slice_mesh_plane` and after reach to the loop, it early stops at the second iteration. This is because the second plane(bottom of red_box) is out of the green_box.

The visualization for right case is like this: 
image

mikedh:
Awesome thanks for the fix and example! Would it be possible to add your example as a unit test as well?

mikedh:
Thanks for the fix!

HyeonseoNam:
I've been too busy to reflect your feedback for a while. Thank you for reflecting the testing code !

Issue #1603, opened 2022-06-10
`trimesh.primitives.Sphere` has no `transform` attribute


clemense:
Hi,
The `Sphere` seems to be the only primitive in trimesh that doesn't have a `transform` attribute. (but instead only a `center`)
Is this on purpose?

It creates a lot of if-then-else constructs when dealing with primitives.  Or is there another way to access this property in a primitive-agnostic way?

Thanks!

mikedh:
Ah yeah I probably just did that because having a full matrix for a sphere seemed silly at the time, but the lack of consistency isn't wonderful either. Maybe `center` should be a `property/setter` and `transform` should be a required property for all primitive attributes (via `abc`?).

PR #1604, opened 2022-06-10
Add transform property to primitives

clemense:
Addresses issue #1603


mikedh:
Awesome, thanks for the PR!

PR #1609, opened 2022-06-21
Non Rigid Iterative Closest Point

Kiord:
Trimesh already features standard ICP for rigid mesh alignment. I think [Non rigid ICP](https://gravis.dmi.unibas.ch/publications/2007/CVPR07_Amberg.pdf) can totally fit in this library.

NRICP allows to register a template mesh (typically with a well designed topology) on another mesh or point cloud.
It is extensively used for building [3D Morphable Models](https://www.youtube.com/watch?v=MlGkzFeyCYc).

Implemented features :
- Algorithm at the end of page 3 (but with a maximum iterations per step)
- Stiffness weight and landmark weight (Eq. 6)
- Distance treshold for outlier matches
- Handles multiple landmark formats
  - vertex index
  - triangle index + barycentric coordinates
  - 3D position (casted to triangle index + barycentric coordinates)
- Normal similarity weighting
  - Implemented as a multiplication by a power of cosines (the power being the weight)
- Source mesh must be a `Trimesh` with faces
- Target shape can be a `Trimesh` or a `PointCloud`. If it has no faces, it is considered as a point cloud
  - Normal estimation for registration to a point cloud

Misc :
- Needed to adapt `trimesh.points.plane_fit` to fit planes to multiple point sets at once
-  Example : Registration of a ball to a teeth in `examples/nricp.py`
- Tests 
  - `tests.test_registration.test_nricp`
  - `tests.test_points.test_plane_fit`

The source mesh and target shape must be well aligned at first.
Typical pipeline : procrustes --> ICP --> NRICP

---
Base implemention inspired from https://github.com/shubhamag/non_rigid_icp


mikedh:
Awesome thanks for the PR! This looks really cool, it might take me a little while to go through it haha. 

- Don't worry about the flake8/python2 errors I can fix those very easily
  - Since we still support Python2 it's getting mad at the f-strings, `"{} {}...".fomat(**kwargs)"`
  - If you want to run the autoformatter there's intructions in the docs: https://trimsh.org/contributing.html  
- Is the `while` loop unbounded? Would it be possible to replace that with a `for _ in range(max_iter)....break`? 
- How is the performance? the Trimesh implementation of `closest_point` might be pretty slow in a loop.
- Is there an even simpler test we could add? Maybe using deformed icospheres, or icospere-capsule?

Kiord:
Thanks for reading so quickly !
- Good for the flake8 errors, now I know what is the issue.
- The loop in the original algorithm is indeed unbounded. 
image
 but I also added a `max_iter` option that bounds the loop. If `max_iter` is negative (`None` would be more clear !), the loop takes the original unbounded behavior.
- The performance is great but `trimesh.proximity.closest_point` is indeed the bottleneck. I think the search for nearest points on target surface is the most costly operation in the algorithm anyways, just like the standard ICP algorithm. With `use_faces=False` a KDTree is used then it is much faster, but less accurate. 
- Yes definitely.

I had another thought : what about defining a class `QueriableGeometry` or whatever from which one can query nearest points? It could be instanced from a `(n, 3)` array, a `Trimesh` or a `PointCloud` so that the query is fully transparent. Options in the `QueriableGeometry.query(points, **options)` method could bring details about the query : return normals if possible ? return triangle index if possible ? Barycentric coordinates ?
I think this could factorize lot of code for ICP-based algorithms (I am working on [another one](https://people.csail.mit.edu/sumner/research/deftransfer/Sumner2004DTF.pdf) at the moment by the way).

Kiord:
Update on the PR : 
- Add  a `query` abstract method to `Geometry3D` and `query_from_points` and `query_from_mesh` in `trimesh.proximity`. 
- Add a `NearestPointQuery` dataclass to store the result of a nearest point query + attributes. List of all attributes so far :
  - `distances`
  - `normals`
  - `vertex_indices`
  - `triangle_indices`
  - `barycentric_coordinates`
  - `interpolated_normals`
- Integrate this mechanism in `nricp` function to remove some boiler plate

Next, the management of landmark types should be addressed

Kiord:
#### (Last ?) Update on the PR

- Implemented the correspondence part of [[Sumner and Popovic 2004 : Deformation Transfer for Triangle Meshes]](https://people.csail.mit.edu/sumner/research/deftransfer/Sumner2004DTF.pdf) in `registration.nricp_sumner`
- Renamed `nricp` to `nricp_amberg`
- Added `Trimesh.face_neighborhood` similar to `Trimesh.face_adjacency` but including face pairs that share at least one vertex (used in `nricp_sumner`)
- For simplicity, landmarks allowed formats are now restricted to :
  - source landmarks 
    - vertex indices (n,) int
    - triangle indices (n,) int +  barycentric coordinates (n, 3) float
  - target landmarks (now _target positions_)
    - 3d positions (n, 3) float
- Updated the example (face registration !) with a new obj in trimesh/models, and updated the tests. 


#### Remaining weak points
- Again, instability in `nricp_amberg` if they are multiple landmarks in the same source triangle
- Normal weighting in `nricp_sumner` comes at the cost of recreating a `Trimesh` at each step to calculate its vertex normals               
  ```python
  from .base import Trimesh
      source_normals = Trimesh(transformed_vertices[:nV],
      source_mesh.faces,
      process=False).vertex_normals
  ```
  Also, `Trimesh` is imported here because importing in the header leads to a circular import... Not clean.
- `Geometry3D.query`, `proximity.query_from_points` , `proximity.query_from_mesh` and `proximity.NearestQueryResult` seem redundant with existing mechanism for closest point query. Still, it factorizes a lot of code in this PR.

---

The implementation of Sumner's Deformation Transfer is inspired from [this repository](https://github.com/mickare/Deformation-Transfer-for-Triangle-Meshes).

Should we put paper references of implemented algorithms in the `README.md` ?

Kiord:
Thank you Mike for the feedback,

No worries this is a big chunk of code and there are other PRs to manage 👍 

To me, it was interesting to transparently query the nearest points of any 3d geometry, but ok.
However `proximity.query_from_points`, `proximity.query_from_mesh` and `proximity.NearestQueryResult` could remain (this logic belongs more to `proximity` than to `registration`).

Regarding the summaries/guides, I think it is a good idea but I might do that later (I will be quite busy until august).

mikedh:
Thanks again for all the work on this! I'm going through it, the visualization for the example is really neat too 😄

Issue #1055, opened 2020-11-15
min_distance_single not doing what it says on the can and broken faces


jonnymaserati:
Hi there

Looks like I've bumped into an issue with the `min_dist_single` method in `trimesh.collision.CollisionManager` class where the method is ***not*** returning the minimum distance.

I've attempted to illustrate the issue in this picture.

![image](https://user-images.githubusercontent.com/41046859/99181923-d11ab000-2731-11eb-9279-6e6cb760dc5a.png)

The red mesh is the external collision object and the blue meshes are the objects in the `CollisionManager` object. I've created a function that systematically slices the red object (adding end caps to each slice), the analogy would be the red object is a salami which I cut into slices and then use the `min_dist_single` method on each slice against the `CollisionManager` class.

The black lines are lines connecting the pairs of minimum distance `points` given for each salami slice. You can see that groups of these lines are visibly connecting to meshes that are further away than others.

Below is an example of a *salami slice* mesh. I've colored the `faces` that `trimesh.repair.broken_faces` returns. For some reason these meshes are not constructing properly, despite having a full complement of vertices and faces (i.e. these are not holes).
Solid | Wireframe
--- | ---
![image](https://user-images.githubusercontent.com/41046859/99184507-a89bb180-2743-11eb-9f16-91f7ee86427f.png) | ![image](https://user-images.githubusercontent.com/41046859/99184520-bbae8180-2743-11eb-90ca-413e37062977.png)

Here's the properties for these meshes:
```
is_convex: False
is_empty: False
is_volume: False
is_watertight: False
is_winding_consistent: False
```
The winding can be fixed with `trimesh.repair.fix_winding`, but `trimesh.repair.fill_holes` returns `False` because I'm pretty sure there aren't any holes. The other repair fix functions return `None`.

It's got me scratching my head this one... any suggestions? I was thinking that the red face on the cylinder end that has an issue might be because of the angles, but then why does the identical face (you can see it in the wireframe) look okay?

Is it a coincidence that all the faces with issues are touching one another? Do I need to introduce another vertex on the end faces, since I noticed that the way I construct the ends means that one of the vertexes have many more neighbors than the others that might be causing the problem (do they all need the same number)?

I'd really appreciate any help/steer.

jonnymaserati:
Okay... I've managed to fix the mesh by passing it through `trimesh.convex.convex_hull`.
Solid | Wireframe
---|---
[...]

PR #1621, opened 2022-07-02
use convex fcl class when mesh is convex, add signed distance flag

mjd3:
This PR incorporates the changes from [my PR in python-fcl](https://github.com/BerkeleyAutomation/python-fcl/pull/50) and adds/fixes a few main points around collision checking and minimum distance calculations:

1. python-fcl now supports bindings for the `Convex` primitive, which is much more efficient and accurate than BVH models both for collision checking and min distance computation if a mesh is convex (see [here](https://github.com/flexible-collision-library/fcl/blob/97455a46de121fb7c0f749e21a58b1b54cd2c6be/include/fcl/narrowphase/distance_request.h#L66) - signed distance calculations can be computed with convexity-based methods and nearest point computations are guaranteed to be on the surface of the mesh). This PR adds a check in the `CollisionManager` class for convexity and uses the `Convex` primitive if the mesh is convex.
3. The signed distance flag is used by default for min distance queries - this flag will have no effect on queries with BVH models  but allows for signed distance computation for convex models.
4. Exposed the `normal` field in the `ContactData` class.
5. Implemented the cache for BVH/Convex models so they aren't recomputed (it looked like this was unfinished from an earlier commit).

Additionally, the linked python-fcl PR fixes an issue where nearest points during distance queries were often incorrect and did not match the returned min distance value; this may have been the issue in #1055.

Hoping you might have time to review this and the linked python-fcl PR - I think they will greatly help out with mesh-mesh collisions and signed distances! Thanks Mike!

### DIFF
--- a/tests/test_collision.py
+++ b/tests/test_collision.py
@@ -118,12 +118,18 @@ class CollisionTest(g.unittest.TestCase):
         tf4 = g.np.eye(4)
         tf4[:3, 3] = g.np.array([-2, 0, 0])
 
+        tf5 = g.np.eye(4)
+        tf5[:3, 3] = g.np.array([5.75, 0, 0])
+
         # Test one-to-many distance checking
         m = g.trimesh.collision.CollisionManager()
         m.add_object('cube1', cube, tf1)
 
         dist = m.min_distance_single(cube)
         assert g.np.isclose(dist, 4.0)
+
+        dist = m.min_distance_single(cube, tf5)
+        assert g.np.isclose(dist, -0.25)
 
         dist, name = m.min_distance_single(cube, return_name=True)
         assert g.np.isclose(dist, 4.0)
@@ -168,15 +174,23 @@ class CollisionTest(g.unittest.TestCase):
         n = g.trimesh.collision.CollisionManager()
         n.add_object('cube0', cube, tf2)
 
-        dist, names = m.min_distance_other(n, return_names=True)
+        dist, names, data = m.min_distance_other(n, return_names=True, return_data=True)
         assert g.np.isclose(dist, 4.0)
         assert names == ('cube0', 'cube0')
+        assert g.np.isclose(
+            g.np.linalg.norm(data.point(names[0]) - data.point(names[1])),
+            dist
+        )
 
         n.add_object('cube4', cube, tf4)
 
-        dist, names = m.min_distance_other(n, return_names=True)
+        dist, names, data = m.min_distance_other(n, return_names=True, return_data=True)
         assert g.np.isclose(dist, 1.0)
         assert names == ('cube0', 'cube4')
+        assert g.np.isclose(
+            g.np.linalg.norm(data.point(names[0]) - data.point(names[1])),
+            dist
+        )
 
     def test_scene(self):
         try:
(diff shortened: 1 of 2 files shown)


mikedh:
Nice this looks awesome thanks for the PR! Quick question before I go through, is there any reason not to automatically switch to using the FCL signed distanced over the `mesh.proximity.signed_distance` if FCL is available?

mjd3:
Great question! My understanding of the `trimesh.proximity` methods is that they are mostly used for checking mesh<->point distances, whereas the `trimesh.collision` classes/methods are for collision checking and min distance queries between two meshes. You can represent point clouds as BVH models in FCL, but then you are back to the current case (i.e., nearest points aren't guaranteed to be on the object, negative distances are not computed via GJK/EPA). That said, there are probably some improvements that can also be made to the `proximity` methods for convex meshes.

mikedh:
Oh yeah good point it's pretty much all `mesh-point` queries in `proximity`. But yeah I think special casing convex there would be great too in the future!

So I installed the FCL from the matching upstream PR, and on Python 3.8 I was seeing this:


```
ERROR: test_collision (__main__.CollisionTest)
----------------------------------------------------------------------
  File "test_collision.py", line 33, in test_collision
    ret, names, data = m.in_collision_single(cube,
  File "/home/mikedh/trimesh/trimesh/collision.py", line 346, in in_collision_single
    contact_data.append(ContactData(names, contact))
  File "/home/mikedh/trimesh/trimesh/collision.py", line 30, in __init__
    names[0]: contact.b1,
TypeError: 'reversed' object is not subscriptable

=======
```

It looks like something's going on with the names, when I embed where that crash happens:
```
In [1]: names
Out[1]: <reversed at 0x7fd867526d90>

# this was created with set(names)
In [5]: self.names
Out[5]: {'__external', 'cube1'}

# but it looks like the items in the reversed were a generator or something that's been consumed
In [6]: list(names)
Out[6]: []
```

The obvious fix would be to at the top of the function add a `names = list(names)` to consume the iterator-like-thing, which does fix the runtime error but then results in failed asserts (probably because it's not doing the right thing anymore haha).

mjd3:
Ah shoot; this is happening because of the new caching actually. It's because both objects passed to the manager are actually the same FCL collision geometry; when we are checking for which object is `o1` in FCL's `Contact`, both of them are the same geometry (i.e., multiple instances of the same cube). I will investigate if there is a better way to check the name order from the returned contact data from FCL; if not I may just abandon the cache idea. It honestly seems kind of weird that the `Contact` returns the geometries instead of the collision objects (which would be unique since they also include the transform), but that's an FCL choice I guess.

mikedh:
Looks like that fixed it! On a test scene of 1000 spheres this sped up the instantiation and check from 5.5 -> 3.5s even with the relatively expensive `is_convex` check, nice!

I'll merge this to my release branch and then merge to main when we get new wheels for FCL up? Thanks for the work on this! I've been hoping to over the long term make collision/ray/scene stuff a little more useful and I think this goes a long way. It also unearthed at least one bug in `is_convex` haha.

mjd3:
Great! Yeah I did some profiling on my end too and I think the biggest improvements are for the min distance checks; I was seeing up to 40x faster for those. I also added a PR in python-fcl for upgrading to FCL 0.7.0 - we can move the conversation over there for that but figured it might be nice to just update/upgrade everything at once.

PR #1625, opened 2022-07-07
Fix normal property of ContactData

mjd3:
Oops I must have copy-pasted some of this from the point method and thus the normal was actually just returning the point again. This should fix that and also extends the tests a bit to verify.


mikedh:
Looks good, thanks for the follow up and test!

PR #1630, opened 2022-07-12
fix "reversed not subscriptable" bug

mjd3:
I think this may not be easily testable so not caught, but it seems like `reversed` should be wrapped in `tuple`, otherwise it is not subscriptable since it creates an iterator.
image


mikedh:

PR #1638, opened 2022-07-18
rename blender_boolean.py template to avoid byte-compilation errors

mgorny:
Rename the blender_boolean.py template to blender_boolean.py.tmpl
to avoid it being considered a regular Python module and Python
attempting to byte-compile it.  The most visible problem with that is
that the installer attempts to byte-compile it and hits a syntax error:

```pytb
$ python -m installer -d /tmp/y dist/trimesh-3.12.9-py3-none-any.whl
*** Error compiling '/tmp/y/usr/lib/python3.11/site-packages/trimesh/resources/templates/blender_boolean.py'...
  File "/usr/lib/python3.11/site-packages/trimesh/resources/templates/blender_boolean.py", line 39
    mesh_pre = $MESH_PRE
               ^
SyntaxError: invalid syntax
```

However, we do have a check for missing .pyc files in Gentoo to detect
packaging errors, and this file triggers it as well.


mikedh:
Looks good, thanks for the PR!  I think `pyinstaller` hits this too. The only argument for the original name is syntax highlighting in vscode haha.

mgorny:
Thanks!

PR #1643, opened 2022-07-22
bugfix: find ply textures with spaces in filepath

jjcmoon:
I found a small bug where trimesh can't load the texturefile of a ply if there's a space in the filepath.

For example, if the header contains:
`comment TextureFile some file.jpg`
The header parser will return `"some"` instead of `"some file.jpg"` for `image_name`.

P.S. thanks for all the work on trimesh, it's an amazing library.


mikedh:
Awesome thanks for the fix! I added a space to the test cases `fuze uv.jpg` which surfaced this same bug in OBJ too.

Issue #1521, opened 2022-03-16
trimesh.exchange.load.load() breaks a glb file


gleb-an:
I execute just this code:
```
scene = trimesh.exchange.load.load(self.input_path, process=False)
file_name = 'test.glb'
full_path = join(self.store.directory, file_name)
glb_data = trimesh.exchange.gltf.export_glb(scene)
with(open(full_path, 'wb')) as glb_file:
    glb_file.write(glb_data)
```

origin: 
![Screenshot from 2022-03-16 22-32-08](https://user-images.githubusercontent.com/8939518/158685685-4a71ef30-0a09-4280-90f0-5299cf1bf2ae.png)
I got:
![Screenshot from 2022-03-16 22-32-25](https://user-images.githubusercontent.com/8939518/158685717-edc5b3cb-0f2a-4578-9407-447918557dbc.png)

mikedh:
Hey, does this still happen if you use the object's export method:

```
scene = trimesh.load(self.input_path, process=False)
scene.export('hi.glb')
```

If that doesn't fix it can you post a link to the file and I can take a look?

gleb-an:
> Hey, does this still happen if you use the object's export method:
> 
> ```
> scene = trimesh.load(self.input_path, process=False)
> scene.export('hi.glb')
> ```
> 
> If that doesn't fix it can you post a link to the file and I can take a look?
I didn't fix it.
[Medical_gas.zip](https://github.com/mikedh/trimesh/files/8288258/Medical_gas.zip)

mikedh:
Hey, yeah it looks like that exporter is using a `byteStride` which we're not handling at the moment:
```
In [5]: view
Out[5]: 
{'buffer': 0,
 'byteOffset': 4152,
 'byteLength': 19008,
 'byteStride': 24,
 'target': 34962}
```

gleb-an:
Do you have any workarounds?

gleb-an:
I updated my origin glb file:
[med_gas_2.zip](https://github.com/mikedh/trimesh/files/8290022/med_gas_2.zip)
But I lost materials and colors:
![Screenshot from 2022-03-17 21-47-18](https://user-images.githubusercontent.com/8939518/158883717-7f62effb-b864-4d5f-b161-938eb45b8713.png)

mikedh:
Yeah unfortunately there's no easy workaround other than supporting `byteStride`, PR's welcome!

PR #1644, opened 2022-07-22
Add support for byte stride

gleb-an:


mikedh:
This looks awesome, thanks! Don't worry about the tests, but would you happen to have a small example model that uses `byteStride` we could add to `models`?

gleb-an:
> This looks awesome, thanks! Don't worry about the tests, but would you happen to have a small example model that uses `byteStride` we could add to `models`?

I believe it will close this issue
https://github.com/mikedh/trimesh/issues/1521

[Medical_gas.zip](https://github.com/mikedh/trimesh/files/9183964/Medical_gas.zip)

mikedh:
Thanks for the PR!

PR #1647, opened 2022-07-28
Weighted Procrustes Analysis

tillns:
I added an optional weight parameter to the procrustes() method in registration.py. It allows to compensate for a few noisy points to make the transformation more robust. Didn't have the time to adjust the test file, but I tested the method visually for myself. Idea to test: 
Add some noise to a subset of the source points, which should increase the returned cost. Now weigh the points that noise was added to less depending on the intensity of the noise. While this will probably increase the average cost even further, it should decrease the cost for the remaining points. (Option that I didn't implement, but I'm not sure if it's good: Instead of returning average cost, return weighted average cost).

### DIFF
--- a/trimesh/registration.py
+++ b/trimesh/registration.py
@@ -174,6 +174,7 @@ def mesh_other(mesh,
 
 def procrustes(a,
                b,
+               weights=None,
                reflection=True,
                translation=True,
                scale=True,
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! I added a quick test, and I noticed the `np.diag(w).dot(...)` was popping up on profiles as quite slow as it's making a large `(n, n)` matrix. I *think* I maintained the logic correctly as this appears to be just doing `(stuff * w.reshape((-1, 1))` 

If I screwed up the logic let me know haha.

tillns:
Thank you for this great open source project! Been using it a lot lately. That's a good catch by you. It shouldn't change the logic. I would only adjust the comment from lines 268 and 269 to not say w_mat anymore but w instead and maybe move it into the else clause.

PR #1656, opened 2022-08-10
Update openctm.py

chriscramer:
fix issue where find_library throws an exception, causing an uncaught exception at line 163 when _ctm_lib_name is not defined


chriscramer:
I'm not sure why we're seeing an exception on finding a library, but rather than getting a None in _ctm_lib_name, we aren't defining it, so a quick fix

mikedh:
Awesome thanks for the fix!

PR #1659, opened 2022-08-11
fix export PBR alphaMode seperated from alphaCutoff

LHolst:
When importing a `glb` file that contained PBR materials with `BLEND` alpha mode and exporting the scene with `file_type='glb'`, the exported material had no alpha mode set.

PR fixes the export for materials with `alphaMode` and unset `alphaCutoff`.
Previously, the export of `alphaMode` was dependent on a defined `alphaCutoff`:
* `alphaMode` 'BLEND' 
* `alphaCutoff` `None`

if alphaCutoff is `None` leading to `alphaMode` not exporting at all.

partially reverting bbed0734ea6ef4d15f647784c3cd4726fc385e59


mikedh:
Thanks for the PR! Would you be able to add a quick test that roundtrips some test data and asserts that it did the right thing?

LHolst:
i'm getting warm with the tests, and i'll try to add a test for exporting gltf files with alphaMode / alphaCutoff settings

LHolst:
added  `test_alphamode` and used the `python setup.py --format` command

used Model comes from https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTest

mikedh:

PR #1681, opened 2022-09-09
Added jupyter notebook example of saving an image of a 3D geometry and test script

justinsavage35:
Added jupyter notebook example of saving an isometric image of a 3D geometry to add to the examples on https://trimsh.org/examples.html .

Any notes would be greatly appreciated.


mikedh:
Awesome thanks for the PR! I might keep just the notebook one, there's [offscreen_render.py](https://github.com/mikedh/trimesh/blob/main/examples/offscreen_render.py) in there which is pretty similar to the script example.

Issue #1684, opened 2022-09-15
Fix FIPS error thrown on useage of MD5 hash function


DonMaEn:
I am unable to use Trimesh on my FIPS enabled gitlab runner due to the use of the MD5 hash. I am attempting to make a PR that fixes this.

The [FIPS](https://en.wikipedia.org/wiki/Federal_Information_Processing_Standards) does not allow use of the MD5 hash function, which is used in several places in Trimesh. The python hashlib library allows you to specify that the md5 function is not being used for security though, which should be the case in trimesh. In this case, use of the MD5 function should be allowed.

PR #1685, opened 2022-09-15
Change uses of hashlib.md5 to an FIPS compliant hash function

DonMaEn:
#1684

### DIFF
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -16,7 +16,7 @@ class CacheTest(g.unittest.TestCase):
         count = 10000
 
         g.timeit.timeit(setup=setup,
-                        stmt='t._modified_m=True;t.md5()',
+                        stmt='t._modified_m=True;t.hash()',
                         number=count)
         g.timeit.timeit(setup=setup,
                         stmt='t._modified_c=True;t.crc()',
@@ -27,8 +27,8 @@ class CacheTest(g.unittest.TestCase):
 
         m = g.get_mesh('featuretype.STL')
         # log result values
-        g.log.info('\nResult\nMD5:\n{}\nCRC:\n{}\nXX:\n{}'.format(
-            m.vertices.md5(),
+        g.log.info('\nResult\nhash:\n{}\nCRC:\n{}\nXX:\n{}'.format(
+            m.vertices.hash(),
             m.vertices.crc(),
             m.vertices.fast_hash()))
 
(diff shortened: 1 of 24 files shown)


mikedh:
Thanks for the PR! If the flag isn't supported by older Python versions, I'm also open to doing something similar to what [this comment on dask](https://github.com/dask/dask/issues/8608#issuecomment-1023329415) proposed and just switch the hash function: 

- on Python 3 use blake2b (if xxhash isn't installed)
- on Python 2 fall back to SHA256 (if xxhash isn't installed)

Looks like these are some options:
```
mikedh@luna:tests$ python2
Python 2.7.18 (default, Jul  1 2022, 12:27:04) 
[GCC 9.4.0] on linux2
>>> import hashlib
>>> hashlib.algorithms
('md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512')
>>> 
mikedh@luna:tests$ 
mikedh@luna:tests$ 
mikedh@luna:tests$ python
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
>>> import hashlib
>>> hashlib.algorithms_available
{'sha224', 'whirlpool', 'sha1', 'sha3_512', 'sha256', 'shake_256', 'sha3_256', 'md5-sha1', 'sm3', 'sha512', 'md4', 'ripemd160', 'sha384', 'sha512_256', 'sha512_224', 'sha3_224', 'blake2s', 'md5', 'shake_128', 'blake2b', 'sha3_384'}
```

DonMaEn:
@mikedh Thanks for the fast reply. Yeah I just realized the workflow for python 2.7 failed. I thought about using a different hash function and I'm open to that idea, but I wasn't sure what to do about the name of the md5 hash functions built into the various classes that use them.

I can make that change, but should I just call the md5 function in classes like the [TrackedArray](https://github.com/mikedh/trimesh/blob/3058152cec78a7c5a3475ddd3c8edf3081f5ec82/trimesh/caching.py#L125) for instance to "hash" or something?

mikedh:
Yeah, I kind of regret naming it `_md5` and I think it would be better as `_hash`. I'm not sure if renaming that totally breaks the test suite but I think it's probably OK to rename.

DonMaEn:
@mikedh Ok I think this should pass all the tests and stuff now.

What I ended up doing is adding a custom hash function to the trimesh/util.py file that is used everywhere the hashlib.md5 function was used. I also changed all the things that said md5 to say hash basically. Let me know if you want me to do anything differently.

mikedh:
Thanks for the updates this looks great! Hash functions definitely need a cleanup.

Quick question:
- Is there a flake8 rule or something for FIPS? I don’t mind adding fixes for it but if it’s not checked in CI unit tests the odds this stays in compliance are pretty low.

Additional stuff, I can make a pass on these before merging so no worries. Apologies in advance if this takes a while to merge, it’s touching a lot of files and I’d like to do some more profiling. Notes:  
-  I think there’s enough hash logic to move it into a submodule:
    - `trimesh/hashes/fast_hash`
    - `trimesh/hashes/fallback_hash`: blake2b/sha256 
- `(mesh, path).identifier_md5`: log a `DepricationWarning` for October 2023 before calling `identifier_hash`
    - Note this should always call the same (not the fastest available) hash function, I would just make this `sha256`
- I’m thinking about also deprecating all of the redundant and potentially confusing `mesh.md5`, `mesh.crc`, `mesh.fast_hash`, and just making `mesh.__hash__` call `fast_hash`
- I think `crc` isn’t used anywhere anymore? I had a hash collision bug somehow with CRC that scared me off despite it being fast, and it’s not really supposed to be used for that.



Quick checks of the functions in question:

```
In [14]: from hashlib import blake2b, md5, sha256

In [15]: data = np.random.random(10000).tobytes()

In [16]: %timeit md5(data).hexdigest()
95.6 µs ± 728 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [17]: %timeit sha256(data).hexdigest()
145 µs ± 492 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [18]: %timeit blake2b(data).hexdigest()
78.8 µs ± 756 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [19]: from xxhash import xxh64_hexdigest

In [20]: %timeit xxh64_hexdigest(data)
4.92 µs ± 26.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

DonMaEn:
I'm not aware of a flake8 rule to check FIPS compliance, but I haven't used flake8 very much. I bet we could enable FIPS mode in one of the test images used in github to ensure it stays FIPS compliant. If that solution sounds good to you I'll look into it.

I agree with your other points.

This PR is from a fork of Trimesh I made, so i'm not sure how you want to handle making changes. I guess you could just make your own branch and pull my changes into it? I could also give you access to my fork, either way.

PR #1697, opened 2022-09-25
add 'TestVHACD' to vhacd executable names

nicholscrawford:
VHACD compiles on ubuntu to TestVHACD as opposed to testVHACD or vhacd. Unsure if the other two are needed, but at least using the instructions from the vhacd repo, it compiles to TestVHACD on ubuntu, and then trimesh won't recognize the executable. This change fixed it for me.


mikedh:
Awesome thanks for the PR! Also I noticed VHACD rewrote their cli recently, has their new version been working for you with the trimesh interface? I haven't had a chance to test it.

Issue #1710, opened 2022-10-11
Adding Simple Line Segments to Scene, still required with scipy/shapely dependency?


Krande:
Hey,

I was just wondering if its possible to add simple line segment geometries to the Scene object without having to install dependencies such as Scipy and Shapely? 

My intention is to export the line segment geometries to glb/gltf. 

Best Regards
Kristoffer

Krande:
So I got it to work without scipy and shapely by simply moving a few dependencies into a try-except `ModuleNotFoundError` block.

Here's the edit's I made: https://github.com/mikedh/trimesh/compare/main...Krande:trimesh:reduce-dep-simple-lines

Basically I just moved the following import lines in ` trimesh/path/path.py`

```python
try:
    from . import repair
    from . import polygons
    from scipy.spatial import cKDTree
    from shapely.geometry import Polygon
except ModuleNotFoundError as e:
    logging.warning(e)
```

Then I did this:

```python
scene = trimesh.Scene()
points = [(0, 0, 0), (1, 0, 0), (1, 1, 0)]
path = trimesh.load_path(np.asarray(points))
scene.add_geometry(path)
scene.export(file_obj="temp/lines.glb", file_type=".glb")
```
And it successfully produced this:

![image](https://user-images.githubusercontent.com/9642232/195117796-42446d07-758a-4431-a709-dec13da1436c.png)

No shapely or scipy packages installed. 

Now all I need to figure out if it's possible to assign vertex coloring to my generated lines :)

Let me know if you would be interested in me creating a PR for the changes. I bet my solution breaks a lot of user generated warnings etc.. So I am happy to modify it to a less breaking proposal. 

Let me know nonetheless!

Best Regards
Kristoffer

mikedh:
Nice that's neat! I'm impressed it got that far haha. Yeah super open to PR's making path more usable with a minimal install! Yeah not sure how well tested vertex colors on paths are haha but fixes and tests also welcome for that!

I'd probably wrap the imports individually in this style so the user gets a nice error message if they really just need to install scipy or whatever:
- https://github.com/mikedh/trimesh/blob/20dca0bc96f0f12dccc0298a9080149bba64ecf2/trimesh/bounds.py#L12-L20
- Maybe add a test to `test_minimal` which is run without any deps in CI.

PR #1712, opened 2022-10-11
Reduce dependencies when adding simple line geometries

Krande:
Hey,

here's the PR with the exception handling and minimal unittest as mentioned in https://github.com/mikedh/trimesh/issues/1710

Let me know if you suggest any changes before merging :)


mikedh:
Thanks for the PR and test!

mikedh:
Looks like you can export to SVG and DXF too with a minimal install!

PR #1734, opened 2022-11-02
fix: GLTF materials are now imported despite not having PIL.Image installed

Krande:
Hey Mike!

I had an issue where my materials weren't properly roundtripped in my glb files. During some debugging I noticed that the  [_parse_materials()](https://github.com/mikedh/trimesh/blob/main/trimesh/exchange/gltf.py#L1138) function stops and returns None in the event `PIL.Image` is not installed (which I believe is only used to parse textures?). 

Anyways, here's a proposal for circumventing this. 
I haven't gotten around to test this yet but I thought I would just start the PR here and I'll fix it and add some extra tests once I have some more time I'll see what I am breaking with my proposal :)

Best Regards
Kristoffer


mikedh:
Awesome thanks for the PR! I'll merge into #1729 which adds testing on a bunch of sample models. Would you mind adding a PR that checks the new behavior on an example model?

Krande:
Sure, I can add a test + example file for checking the roundtripped material assignment on glb's :) Would you like to merge #1729 first before I continue on this PR? (I am assuming you ment for me to update this PR once you're finished merging?)

PR #1735, opened 2022-11-03
Fix kwargs not passed to cylinder

khameeteman:
Creating an `annulus` with r_min < tol.merge returns a `cylinder`, however kwargs are not passed (e.g. `visual`)


khameeteman:
Hi @mikedh I don't know why this is failing, do you?

mikedh:
Awesome thanks for the PR! No worries about the tests, those are broken for unrelated reasons (until #1729 is merged).

Issue #1385, opened 2021-10-20
`trimesh.Scene.scaled(scale)` with 3d `scale`


clemense:
Hi,
I would like to expand the `trimesh.Scene.scaled()` function to be able to use a 3d vector as argument (instead of a scalar) similar to what is possible in `trimesh.Geometry.apply_scale()`.

When looking at
https://github.com/mikedh/trimesh/blob/5c8891defc1967cfcb9fd7c0c908b4f666364091/trimesh/scene/scene.py#L989
it seems that the problem is that the scene could be 2D in which case the behavior would be ill-defined.
But just to clarify, can it be the case that a scene contains both, 2D and 3D geometry?

Thanks!

PR #1738, opened 2022-11-06
Non-uniform scaling 

clemense:
This addresses #1385 

It allows non-uniform scaling, such as `scene.scaled([0.3, 1.3, 10.0])`.

This PR contains also two new tests.

Under the hood, the implementation is a bit more involved, since:
a) I wanted to keep the scene graph structure and not flatten it (unlike the uniform scaling variant).
b) For non-uniform scaling, geometries cannot appear multiple times in the scene (and need to be duplicated first).

I basically kept the old implementation and wrapped it in a big if-else for the non-uniform/3D scaling case.


mikedh:
Awesome thanks for the PR! I'll try to get this in as part of #1732. This looked good in quick testing (after I stopped testing `scene.apply_scale` anyway :).

Issue #1557, opened 2022-04-07
Enhancement Ideas 2


mikedh:
PR's are always super welcome! This is a relatively small open source project and really benefits from the bugfixes, features, and other stuff the 100+ contributors have PR'd, so thanks! 

The previous issue (#199) got pretty out of date and a lot of the stuff on it was done so here's a new thread. Feel free to suggest things!

- Implement an actually correct and maybe even faster OBB algorithm (#1544)
- Finally get the `embree3` stuff over the line to be merged
   - Wheels for embree3 are [done-ish](https://github.com/sampotter/python-embree/pull/17) but probably need some more tests
   - A number of people have worked on this ray branch https://github.com/mikedh/trimesh/pull/1108 
- Support Shapely 2.0 when it releases (#1556)
- Wishlist: find some way to make `scene.save_image` always work. We get a ton of issues about this and the answer is always pretty much "sorry your graphics driver hates you," or "try the docker image." I guess it's just a hard problem that depends a lot on platform and I'm not sure there's any way to solve it from trimesh
- Polish the wart where OBJ sometimes return a `Scene`. Maybe now that multi-material single meshes were PR'd in https://github.com/mikedh/trimesh/pull/1246  we could have OBJ always return a `Trimesh`.

leonmkim:
Firstly, want to say thank you @mikedh for the simply amazing library! Incredibly impressive how much functionality trimesh contains. 

Just wanted to suggest, it'd be nice to seeing some of the functionality, particularly under proximity and collisions, leveraging primitives when possible rather than treating everything as arbitrary meshes. I've been using the primitive generation trimesh provides, but as far as I know, things like signed distance queries and collision queries are not taking advantage of the metadata associated with the primitive geometries. Please correct me if I'm wrong! I've written my own set of SDF queries for the primitives I'm working with right now (boxes, cylinders, spheres) and found the speedup for my use case considerable (just adapted the very comprehensive set of examples from the wonderful [Inigo Quilez' site](https://iquilezles.org/articles/distfunctions/)).  

```
def sdf_box(locations, world_tf_box, box_half_dims):
    # locations is a Nx3 array
    # world_tf_box is a 4x4 transformation matrix
    # box_dims is a 3x1 array of [length, width, height]
    # returns a Nx1 array of sdf values
    # get the inverse transformation matrix
    box_tf_world = invert_transform(world_tf_box)
    # transform the locations to the box frame
[...]

Issue #1004, opened 2020-09-21
Looking for more advanced funtion for mesh subdivision


Xin-gif:
Hi,

I'm now trying to refine and decimate the meshes in order to get them ready for further process. I used mesh.subdivide() function to subdivide the shapes that don't have enough vertices and faces. But this function uses mid-point subdivision. It works well but only if initially shapes have polygons having roughly the same size. Otherwise, some (already) small faces would get extremely finely subdivided.  I'm wondering is there any more advanced remeshing functions that i can use to fix this? Also i couldn't find any function to do the decimation, could you please let me know if there's a function for that?

Looking forward to hearing from you.
Thank you

YuanWenqing:
Same problem! Mid-point subdivision will produce four new faces for the original one. So a long-narrow face will be subdivided to four more long-narrow faces.

Here is my modified version of original `trimesh.remesh`. It only subdivides the longest edge of each face and thus produces only two new faces. It works fine in my project.

~~~python
# coding: utf8
"""
modified from trimesh.remesh, only subdivide the longest edge
"""
import numpy as np
from trimesh import grouping
from trimesh import util


def subdivide(vertices,
              faces,
              face_index=None):
    """
    Subdivide a mesh into smaller triangles.

    Note that if `face_index` is passed, only those
    faces will be subdivided and their neighbors won't
    be modified making the mesh no longer "watertight."

    Parameters
    ------------
    vertices : (n, 3) float
      Vertices in space
    faces : (m, 3) int
      Indexes of vertices which make up triangular faces
    face_index : faces to subdivide.
      if None: all faces of mesh will be subdivided
      if (n,) int array of indices: only specified faces

    Returns
    ----------
    new_vertices : (q, 3) float
      Vertices in space
    new_faces : (p, 3) int
      Remeshed faces
    """
    if face_index is None:
        face_index = np.arange(len(faces))
    else:
        face_index = np.asanyarray(face_index)

    # the (c, 3) int array of vertex indices
    faces_subset = faces[face_index]  # (F,3)

    # find max edge of each face
    face_edges = faces_subset[:, [0, 1, 1, 2, 2, 0]].reshape((-1, 3, 2))  # (F,3,2)
    face_edges_length = ((np.diff(vertices[face_edges], axis=2) ** 2).sum(axis=3) ** 0.5).reshape((-1, 3))  # (F,3)
    face_edges_argmax = np.argmax(face_edges_length, axis=1)  # (F,)
[...]

PR #1750, opened 2022-11-16
Add loop subdivision algorithm

HyeonseoNam:
Hi, I added an loop subdivision algorithm as the need for the advanced subdivision method mentioned in #1004.

I tried to implement the Catmull-Clark method as you mentioned in #1557, but I found that this method is suitable for quad mesh and requires additional triangulation. Also, loop subdivision is more preferred in triangle mesh (these are disccussed in https://gamedev.net/forums/topic/684378-does-catmull-clark-subdivision-work-for-meshes-consisting-of-only-triangles/5322569/). So I implemented loop method instead of catmull.

My implementation is based on this reference: p70 in http://multires.caltech.edu/pubs/sig00notes.pdf/

1. Calculate odd vertices: Assign a new odd vertex on each edge and calculate the value for the boundary case and the interior case.
2. Calculate even vertices: Calculated new even vertices from the existing vertices and their adjacent vertices.
3. Compose new faces.

### DIFF
--- a/tests/test_remesh.py
+++ b/tests/test_remesh.py
@@ -117,7 +117,7 @@ class SubDivideTest(g.unittest.TestCase):
             assert len(sub.faces) > len(m.faces)
             # subdivided faces are smaller
             assert sub.area_faces.mean() < m.area_faces.mean()
-   
+        
     def test_loop_multibody(self):
         mesh = g.get_mesh('cycloidal.ply')  # a mesh with multiple bodies
         sub = mesh.loop(iterations=1, multibody=True)
@@ -125,7 +125,7 @@ class SubDivideTest(g.unittest.TestCase):
         assert len(sub.faces) > len(mesh.faces)
         # subdivided faces are smaller
         assert sub.area_faces.mean() < mesh.area_faces.mean()
-
+        
     def test_loop_correct(self):
         box = g.trimesh.creation.box()
         big_sphere = g.trimesh.creation.icosphere(radius=0.5)
@@ -135,8 +135,8 @@ class SubDivideTest(g.unittest.TestCase):
         assert big_sphere.contains(sub.vertices).all()
         # bigger than 0.4 sphere
         assert (~small_sphere.contains(sub.vertices)).all()
-
-    def test_loop_bound(self):
+        
+    def test_loop_bound(self): 
         def _get_boundary_vertices(mesh):
             boundary_groups = g.trimesh.grouping.group_rows(
                 mesh.edges_sorted, require_count=1)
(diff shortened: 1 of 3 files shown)


mikedh:
This looks awesome thanks for the PR! Good catch with catmul looking lousy on triangle meshes, this seems like a much better solution and it's sweet you were able to vectorize it! I'll try to test and merge later this week, but my initial notes are: 

- don't worry about the formatting, I can make a pass pretty easily. 
- I think I'll rename `Trimesh.loop` -> `Trimesh.subdivide_loop` to match the other functions. 
- is the multibody flag necessary since it's checking `graph.neighbors` inside? Or does it get wacky for multiple bodies if you don't split?

HyeonseoNam:
I'm very happy that I can contribute to this thankful trimesh :) 

I also thought it is quite strange about the name of function, `loop`. `subdivide_loop` seems to be much more appropriate.

Yes, since `graph.neighbors` just gives all neighbors.
In the algorithm, to calculate even vertices, summation should be performed by each vertex's neighbors for same body, but `graph.neighbors` gives all neighbors from all shared bodies, then, it is quite ambiguous to classify where each neighbor belongs to. 
Calculating odd vertices(before calculating evens) is in the same context. An edge should be shared with 1 (boundary case) or 2 (interior case) faces for same body. If more than 2 (e.g., `grouping.group_rows(edges, require_count=4)`), it is also ambiguous to figure out whether each face is in the same body or not. 

So I handled this case (more than 2 faces share one edge) as an error, and made it run for each splitted body. This has possible to be slow if there are many bodies, so I think there will be a better way.

HyeonseoNam:
Oh, if `split` and `concatenate` are performed at the beginning of the code for all cases, the multibody flag seems to be unnecessary. If you want to me to make changes, I'll send a commit (to remove multibody flag), or refer to it when you reflect code !

```python
mesh = trimesh.load("trimesh/models/cycloidal.ply")
# sub = mesh.loop(iterations=1, multibody=False) # will return error
# sub = mesh.loop(iterations=1, multibody=True) # will use for loop

# <-- replace multibody flag with
splited_meshes = mesh.split(only_watertight=False)
if len(splited_meshes) > 1: 
    mesh = trimesh.util.concatenate(splited_meshes)
# --> 
sub = mesh.loop(iterations=1, multibody=False) # no for loop, no error
```


In addition, I put some rendered results. Maybe it will be helpful for reference when people use.

`target_mesh = trimesh.load("trimesh/models/cycloidal.ply")`
image

`target_mesh = trimesh.creation.cone(1, 1.5)`
image

a custom mesh
image

PR #1771, opened 2022-12-01
Fix broken link in README.md

mavimo:
SSIA


mikedh:
Awesome thanks for the fix!

PR #1782, opened 2022-12-12
Extends arguments list of update_vertices

patrick-nicodemus:
All the child classes take 'mask' as an argument, so the parent class should too.


mikedh:

PR #1793, opened 2022-12-29
Fix name of quadric decimation method

nickswalker:
Simple typo, but makes the functionality harder to discover


mikedh:
Looks great, thanks for the PR! I'll probably add a `DeprecationWarning` to the base method for a year or so just to give people time to migrate from my baked in typo 😄.

Issue #1640, opened 2022-07-19
transform_inertia: add translation contribution


danielcostanzi18:
Hi, I'm working on a design project and I need to compute the inertia tensor of some meshes, but not w.r.t. their center of mass (CoM), with respect to a different frame, with the origin **not overlapped** with the CoM.
Given a `mesh`, seems to me that the property `moment_inertia` is the inertia tensor computed w.r.t. a frame with the axes aligned with the global frame and the origin in the CoM of the mesh, right?

The method `transform_inertia` only project the inertia matrix in a new frame. Would be very useful to write a method that computes the inertia matrix w.r.t an arbitrary origin applying the parallel axis theorem.

I'd be glad to contribute to this improvement if you think it is worthwhile and you are not already working on it.

Regards
DC

mikedh:
That would be awesome, PR's welcome! Yeah as you mentioned `transform_inertia` just applies the rotation.

With the inertia stuff my main concern is with tests given that my intuitive understanding isn't great haha. A few undergrad-textbook-examples as unit tests and simple sanity checks would be really nice.

danielcostanzi18:
Ok great! I implemented an example to translate it, now I have to organize it in a simple function and, as you said, find a simple example to validate the method.
I'll do a PR as soon as I have time to work more on it.

PR #1827, opened 2023-02-03
Inertia tensor w.r.t to arbitrary frame (Issue #1640)

Raserino:
Hi,
I wanted to get started with contributing to Trimesh because it is a very nice project on an interesting topic. Thank you so much for your work on it!

I saw the issue #1640 is still open, so I implemented a function to give the inertia tensor w.r.t to a frame using parallel axis theorem (Steiner’s theorem).

I added a test testing the function on different frames on a cuboid using textbook formulas to calculate the inertia tensors. It tests both translation and rotation of the input frame.

This is my first ever PR, so I hope I did everything right. I am looking forward to your feedback!

### DIFF
--- a/tests/test_inertia.py
+++ b/tests/test_inertia.py
@@ -239,6 +239,110 @@ class InertiaTest(g.unittest.TestCase):
 
             assert g.np.allclose(MI_gt, MI)
 
+    def test_frame_inertia_box(self):
+        # set up a rectangular cuboid, vertices in this order:
+        # Two edges have the lengths 1,2 and 3
+        #           ^ x0, x2
+        #    y2 ^  /    b
+        #       | 1-----------2
+        #       |/|          /|
+        #       0-+---------3-+----> y0, x1, z2
+        #      /| |         | |
+        #     / | |         | |
+        # y1 v  | |         | |h
+        #       | |         | |
+        #       | |         | |
+        #       | 5---------+-6
+        #       |/          |/ d
+        #       4-----------7
+        #       |
+        #       v z0, z1
+
+        h = 3
+        b = 2
+        d = 1
+        # definition in frame 0
+        vertices = g.np.float32([[0, 0, 0],
+                                 [d, 0, 0],
+                                 [d, b, 0],
+                                 [0, b, 0],
+                                 [0, 0, h],
+                                 [d, 0, h],
+                                 [d, b, h],
+                                 [0, b, h]])
+
+        # 6 quad faces for the cube
+        quads = g.np.int32([[3, 2, 1, 0],
+                            [0, 1, 5, 4],
+                            [1, 2, 6, 5],
+                            [2, 3, 7, 6],
+                            [3, 0, 4, 7],
+                            [4, 5, 6, 7]])
+
+        # create a trimesh cube from vertices and faces
+        tm_cube = g.trimesh.Trimesh(vertices, quads)
+
+        # set density of 1 and calculate mass
+        tm_cube.density = 1
+        mass = tm_cube.density*h*b*d
+        assert tm_cube.mass == mass
+
+        def parallel_axis_theorem(inertia, mass, a1, a2, a3):
+            """
+            ApplY parallel axis theorem
+
+            Parameters
+            ------------
+            inertia: (3,3) float
+              inertia tensor around centre of mass
+            mass : float
+            a1, a2, a3: float
+              difference between desired rotation center and center of mass in x,y,z
+
+            Returns
+            -----------
+            rotated_inertia : (3,3) float
+            Inertia tensor
+
+            """
+            # copy from wikipedia
+            return inertia + mass * g.np.array([[a2**2 + a3**2, -a1*a2, -a1*a3],
+                                                [-a1*a2, a1**2 + a3**2, -a2*a3],
+                                                [-a1*a3, -a2*a3, a1**2 + a2**2]])
+
+        # CHECK FRAME 0
+        # manual analytical calculations of inertia tensor
+        inertia0 = 1/12 * mass * g.np.diag([h**2 + b**2, h**2 + d**2, b**2 + d**2])
+        a1 = -d/2
+        a2 = -b/2
+        a3 = -h/2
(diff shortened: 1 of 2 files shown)


mikedh:
This looks awesome, thanks for the PR!!

I was trying to validate the logic by using this to compose the inertia tensor of a scene, and compare it to the "dumped mesh" version of that scene:
```
import numpy as np
import trimesh

if __name__ == '__main__':
    # load a scene with the same mesh instanced multiple times
    s = trimesh.load('models/cycloidal.3DXML')

    trimesh.util.attach_to_log()

    moments_geom = {k: g.moment_inertia for k, g in s.geometry.items()}
    # collect the moments of inertia in different frames
    moments = []

    for node in s.graph.nodes_geometry:
        # get the transform that moves each mesh into
        # it's position in the scene
        matrix, geom = s.graph[node]

        # I think we want the inertia
        moments.append(trimesh.inertia.frame_inertia(
            mesh=s.geometry[geom], transform=np.linalg.inv(matrix)))

    # sum the moments we calculated
    total_scene = np.sum(moments, axis=0)

    # apply the transforms dumping instances into a single mesh
    # as a "truth" moment of inertia using the surface integral
    ms = s.dump(concatenate=True)
    total_dump = ms.moment_inertia

    # compare the two calculation methods by percent
    diff_percent = np.abs((total_dump - total_scene) / total_dump)
    print('looks like the matrices differ percent-wise by:\n', diff_percent)
```
Outputs:
```
looks like the matrices differ percent-wise by:
 [[0.18876 0.00246 0.58601]
 [0.00246 0.18876 2.81289]
 [0.58601 2.81289 0.     ]
```

It looks like it's the same order of magnitude but I played with it a bit and couldn't get them to match exactly within float precision (which I think they should?) I'm probably doing something wrong here haha. Other quick questions:
- this line is currently multiplying element-wise, not a dot product, is that right? `aligned_inertia = center_aligned_inertia + mesh.mass * M`
- seems like we could move the parallel axis as-implemented here into `transform_inertia`? It looks like that function silently ignores translation haha.

PR looks great thanks for implementing this!

Raserino:
Thank you so much for your extensive feedback and your thoughts!! Your work on the whole project is really incredible. Also your idea to test is very smart. It didn't allow me to load the 3Dxml file, but I created a similar scene with np.random.

The small differences in your code occur because for `total_scene` you get the moment of inertia around the base_frame origin. For `total_dump` you get the momemt of inertia around the center of mass of all the meshes combined. 
As the meshes are randomly placed around the base frame origin, the center of mass is close to the origin and the differences are only small. 
I modified the code to also get the moment of inertia around the center of mass for `total_scene`. Then the differences are zero.
The documentation does not tell right now that the moment of inertia as a mesh property is always around the center of mass. This could be a helpful addition.

```
import numpy as np
import trimesh

if __name__ == '__main__':
    # load a scene with the same mesh instanced multiple times
    mesh = trimesh.load('models/cycloidal.ply')
    s = trimesh.scene.scene.Scene(geometry=mesh)
    for i in range(0,10):
        transform = trimesh.transformations.rotation_matrix(
                                                        2*np.pi*np.random.rand(), 
                                                        np.random.rand(3) , 
                                                        50*np.random.rand(3))
        s.add_geometry(geometry=mesh,
                        node_name='node' + str(i),
                        geom_name='geom' + str(i),
                        transform=transform
                        )

    trimesh.util.attach_to_log()

    moments_geom = {k: g.moment_inertia for k, g in s.geometry.items()}
    # collect the moments of inertia in different frames
    moments = []

    # apply the transforms dumping instances into a single mesh
    # as a "truth" moment of inertia using the surface integral
    ms = s.dump(concatenate=True)
    total_dump = ms.moment_inertia

    # get center mass of dumped meshes
    center_mass = ms.center_mass

    # transformation from origin to center_mass
    origin_T_center_mass = np.eye(4)
    origin_T_center_mass[:3,3] = center_mass

    for node in s.graph.nodes_geometry:
        # get the transform that moves each mesh into
        # it's position in the scene
        matrix, geom = s.graph[node]

        # transformation from mesh to center_mass of all meshes together
        transform = np.dot(np.linalg.inv(matrix), origin_T_center_mass)

        # I think we want the inertia
        moments.append(trimesh.inertia.frame_inertia(
            mesh=s.geometry[geom], transform=transform))

    # sum the moments we calculated
    total_scene = np.sum(moments, axis=0)

    # compare the two calculation methods by percent
    diff_percent = np.abs((total_dump - total_scene) / total_dump)
    print('looks like the matrices differ percent-wise by:\n', diff_percent)
```

Output:
```
looks like the matrices differ percent-wise by:
 [[0. 0. 0.]
 [0. 0. 0.]
 [0. 0. 0.]]
```

- Yes the multiplication should be element-wise. I think it can only be element-wise as `mesh.mass` is not a matrix

- `transform_inertia` actually rotates the inertia tensor by a rotation matrix R. 
In my function the base gets rotated, not the inertia tensor.
The difference is in the order of the dot product: (I: Inertia tensor, R: Rotation matrix)
`I_rotated = R*I*R_transposed`
`I_baserotated = R_transposed*I*R` (both matrix multiplication)
 It actually took me quite some time to figure out this makes a difference. First I used `transform_inertia` to rotate in `inertia_frame` and was wondering why the test kept failing. The name `transform_inertia` is a bit misleading in my opinion. I would keep both functions, but maybe rename `transform_inertia` to `rotate_inertia`.

mikedh:
> It actually took me quite some time to figure out this makes a difference. First I used transform_inertia to rotate in inertia_frame and was wondering why the test kept failing. The name transform_inertia is a bit misleading in my opinion. I would keep both functions, but maybe rename transform_inertia to rotate_inertia.

Great point this confused me too haha. Right, since the moments of inertia are axis-aligned at the mesh center of mass, if you transform a mesh you only have to worry about rotation to get the new center-mass moment of inertia. I agree that function should really be named `rotate_inertia`, or possibly include parallel axis logic if it stays named `transform_inertia`. I'll try to clarify the docstrings before releasing and add a shortcut for this new logic to mesh objects, maybe `mesh.moment_inertia_frame(transform)`. Thanks again for the PR!

Raserino:
Thank you for merging!

PR #1830, opened 2023-02-07
NRICP guide

Kiord:
As discussed in #1609, here is a small guide on non-rigid registration techniques `nricp_sumner` and `nricp_amberg`.

I am not 100% sure of the technical details provided, but it should give more insights about NRICP. 
A review of the quality of English would be appreciated.

- Added a `docs/guides/nricp.md` markdown file.

(PS : some commits from my previous PR are mentioned here I don't know how to get rid of them)


mikedh:
Looks great, thanks for the PR!! I'll try to make an edit pass before releasing but it looks good. I've also been meaning to move some stuff from the readme into `docs/guides` like you've done here, I think I'll also need to set up the `toctree` so the docs build generates links at some point.

PR #1833, opened 2023-02-10
Second moment

Raserino:
Hi,
I added a new feature regarding the second moment of area for polygons. 

I added a `centered` input flag, which can center the polygon before the second moments are calculated. 
If the `centered` flag is set, the function also performs a principal axis transform. This is pretty similar to inertia tensors. Ixy should be 0 after the transformation.

The function outputs the angle by which the polygon needs to be rotated (around the centroid because `centered` must be set True) and a transformation matrix to perform that rotation and a translation to move the origin to the centroid. 
Normally `alpha` is the angle by which the frame needs to be rotated, not the shape. I decided to return `-alpha` which is the angle by which the polygon needs be rotated. I felt this is more intuitive in context of the other polygon functions. 

Before my changes the function also allowed (n,2) numpy arrays as input instead of a polygon. I removed this because the function needs the `Polygon.centroid`. For the numpy array there had to be another function able to do that. 
And for the centroid you need to calculate the area which also would require another function. It wouldn’t be overly complicated to implement this but I thought the use of the shapely Polygon is exactly because to prevent implementing all this simple functions. Also, all the other functions in `polygon.py` only take polygon inputs so it is more consistent now. But if you think otherwise, let me know, I can re-add support for numpy arrays.

I used my university textbook as source, but the content is the same as on this website: https://leancrew.com/all-this/2018/01/transforming-section-properties-and-principal-directions/ I was surprised there is rather little reference on the internet on this as this was quite a big thing in my mechanics class. 

I added a test on a double cornered rectangle: One rectangle is still in the first quadrant of the coordinate system, but there is another rectangle rotated by 180 deg. The idea is to have a polygon that has `Ixy != 0` with the centroid at the origin. This polygon is then rotated by the outputted alpha and the the test checks if after that `Ixy == 0`. 

I hope you like it, I appreciate your Feedback!

### DIFF
--- a/tests/test_polygons.py
+++ b/tests/test_polygons.py
@@ -154,7 +154,7 @@ class PolygonTests(g.unittest.TestCase):
         def poly_doublecorner(bh):
             # returns two equal sized rectangles as one polygon
             # Same as poly_corner(), but the rectangle gets
-            # mirrored by 180° at the origin
+            # mirrored by 180 deg at the origin
             # This puts the centroid in the origin with Ixy != 0
             shell_1 = rectangle(bh)
             shell_1 += shell_1.min(axis=0)
@@ -189,6 +189,7 @@ class PolygonTests(g.unittest.TestCase):
 
         from trimesh.path.polygons import second_moments
         from trimesh.path.polygons import transform_polygon
+
         from shapely.geometry import Polygon
 
         heights = g.np.array([[0.01, 0.01],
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome thanks for the PR!! I'll try to check it against https://en.wikipedia.org/wiki/List_of_second_moments_of_area


BTW `scene.moment_inertia` and `scene.moment_inertia_frame` with logic from your previous PR are huge speedups over naively dumping the scene, as well as it being nice to get it in any frame, thanks!

Raserino:
Hi I am glad to hear that about the previous PR! 
And thank you for looking into it the new second moment function!

If you want to try to test it yourself I recommend the "equal legged angle" example from the wikipedia page you linked. The moments Ia and Ib listed should be the principal moments of inertia returned by the function. Alpha should be 45°. You can see the principal axis as a and b on the wikipedia picture, they are rotated by 45° to the x- and y-Axis.

The rest of the examples are not too helpful in my opinion because they just give formulas for the basic second moments, but the calculation of these is the part of the function I didn't change.

PR #1835, opened 2023-02-10
fix dummy Profiler class to be a proper context manager

mgorny:
The dummy `Profiler` class that is used whenever `pyinstrument` is not available did not return a object in its `__enter__` method, effectively breaking statements like:

    with g.Profiler() as P:
        pass
    print(P.output_text())

Make it return itself to fix that.  This fixes 4 test failures for us.

### DIFF
--- a/tests/generic.py
+++ b/tests/generic.py
@@ -49,8 +49,8 @@ try:
 except BaseException:
     # make a dummy profiler which does nothing
     class Profiler(object):
-        def __enter__(*args, **kwargs):
-            pass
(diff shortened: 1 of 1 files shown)


mikedh:

mikedh:
Merged into the wrong branch this fix was so good 😆

mgorny:
Thanks. I really appreciate thinking about us who don't need profiling to run tests, this is definitely better than the `sed` we used to have ;-).

PR #1838, opened 2023-02-11
added lxml import error message print

Raserino:
If the lxml package for loading 3DXML/XAML/3MF is not installed, the `ImportError` is not displayed but only `ValueError: File type: 3dxml not supported`. This is not too helpful because it does not say that you just need to install lxml to make it work.


mikedh:
Thanks for the PR! Yeah just omitting formats from the loader dict is a little gnarly, I think the way I'd like to handle this is to re-raise the `ImportError`. I think my refactor #1839 will accomplish the same thing (`trimesh.load('hi.3dxml')` will raise the `lxml` `ImportError` to make debugging more transparent.

Issue #1829, opened 2023-02-07
Error when loading an OBJ having face like data in a g tag


clemgaut:
Trimesh version: 3.17.0

When trying to load an obj, I get the error
`ValueError: invalid literal for int() with base 10: 'Three'`

It originates from the file trimesh/exchange/obj.py line 504 `v.append(int(split[0]))` in the function `_parse_faces_fallback`

The OBJ that produces this issue is:

> g An f Three Other Thing
v 388.018188 778.077576 347.121552

It seems that trimesh considers the part after the "f" as being a face. I managed to produce the minimal example above showing the issue.
When removing one of the words after the "f", trimesh is able to load the obj:

> g An f Two Things
v 388.018188 778.077576 347.121552`

I therefore suspect an error related to trimesh considering that whatever is after an "f" and has 3 components is considered a face, even though that text is within an existing "tag" (in this cas the "g" tag)

After investigating, the issue was likely introduced in the commit https://github.com/mikedh/trimesh/commit/c0cc6bba08f4dad8ce3e634ebf5f41865ac7c086

PR #1841, opened 2023-02-14
Removes g tag from the list of faces when loading an obj

clemgaut:
Fixes the issue https://github.com/mikedh/trimesh/issues/1829 where loading an obj could fail if a 'g ' tag contains a face like name.


mikedh:
Looks great, thanks for the fix and test!! Makes sense to do it in the same place we're tracking objects/materials.

clemgaut:
You are welcome, thanks for sharing such a nice and useful library :)

PR #1842, opened 2023-02-14
Ensure that packed UV output order matches original order for multiple materials with duplicates.

overlind:
This is intended to fix a specific corner case I ran into where the returned array of UVs of a mesh using multiple textures (of which two or more are identical) does not match the order of the original array of UVs after the deduplication of the textures is applied in this pack function.  As a result of this issue, downstream processes end up with a scrambled texture applied.  I am relatively new both to Python and to the Trimesh library, so I apologize if my proposed solution introduces new complications.


mikedh:
Awesome thanks for the fix! Would it be possible to add a small model demonstrating this to unit tests?

mikedh:
I'll play with this in a release branch, thanks for the PR!

Issue #1846, opened 2023-02-20
stl.py -> Binary header "face count" should be unsigned int


mkaiser94:
https://github.com/mikedh/trimesh/blob/556c2b41071f5cc727a2584df193e8967aa1f657/trimesh/exchange/stl.py#L20

Hey everybody, I got a header error while loading a rather large STL:
 `trimesh.exchange.stl.HeaderError: Binary STL has incorrect length in header: 2500000800 vs [-1794966496]`

and found that in line 20 of stl.py (linked above), the binary header has a data type of "<i4", which is a signed int, defined for the face count, while it should be an unsigned int. 
https://en.wikipedia.org/wiki/STL_(file_format)

Is that intentional or a mistake?

Cheers,
Martin

Ps: Tagging @mikedh and @hroncok because they worked on the file

mkaiser94:
I figured that it is a mistake and made a mini PR :)

PR #1847, opened 2023-02-21
Change face_count data type from signed to unsigned int in binary stl header

mkaiser94:
In stl.py, the data type for face_count in the binary stl header should be an unsigned int and not a signed int, as mentioned in #1846. This creates potential problems when opening STLs with a face_count thats higher than 2147483647. This mini-PR simply changes "<i4" numpy-datatype to "<u4" to fix this issue.

### DIFF
--- a/trimesh/exchange/stl.py
+++ b/trimesh/exchange/stl.py
@@ -17,7 +17,7 @@ _stl_dtype = np.dtype([('normals', '<f4', (3)),
                        ('attributes', '<u2')])
 # define a numpy datatype for the header of a binary STL file
 _stl_dtype_header = np.dtype([('header', np.void, 80),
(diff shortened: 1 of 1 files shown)


mikedh:
Awesome thanks for the fix!! Wow did it work without crashing on a 2 billion face mesh? I bet that surfaced some performance issues haha any fixes very welcome!

mkaiser94:
Worked like a charm! And gave me a good excuse to get a second cup of coffee while loading the mesh :D Jokes aside, it only takes 43s to open this 2.3Gb STL that I use for "stress testing", which I think is not bad at all considering the processing that trimesh does on this nasty file.    

Thanks for your effort, great library

Issue #1848, opened 2023-02-22
bounding_box_oriented does not work on 2D meshes


pchorak:
Qhull throws an error when given 3D points that form a less than 3 dimensional convex hull (used in `bounding_box_oriented`).

Commands to reproduce the error
~~~
import trimesh
mesh = trimesh.Trimesh(vertices=[[0, 0, 0], [0, 1, 1], [1, 0, 0], [0, 0.5, 0.5]], faces=[[0, 1, 2], [1, 3, 2]])
mesh.bounding_box_oriented
~~~

Environment
- Python 3.10.6
- SciPy 1.10.1

Qhull error message
~~~
QhullError: QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)

While executing:  | qhull i Qt Pp QbB
Options selected for Qhull 2019.1.r 2019/06/21:
  run-id 951852643  incidence  Qtriangulate  Pprecision-ignore
  QbBound-unit-box 0.5  _pre-merge  _zero-centrum  _max-width  1
  Error-roundoff 6.9e-16  _one-merge 4.9e-15  _near-inside 2.4e-14
  Visible-distance 1.4e-15  U-max-coplanar 1.4e-15  Width-outside 2.8e-15
  _wide-facet 8.3e-15  _maxoutside 5.5e-15

The input to qhull appears to be less than 3 dimensional, or a
computation has overflowed.

Qhull could not construct a clearly convex simplex from points:
- p3(v4):  -0.5     0     0
- p1(v3):  -0.5   0.5   0.5
- p2(v2):   0.5  -0.5  -0.5
- p0(v1):  -0.5  -0.5  -0.5

The center point is coplanar with a facet, or a vertex is coplanar
with a neighboring facet.  The maximum round off error for
computing distances is 6.9e-16.  The center point, facets and distances
to the center point are as follows:

center point    -0.25   -0.125   -0.125

facet p1 p2 p0 distance=    0
facet p3 p2 p0 distance=    0
facet p3 p1 p0 distance= -0.25
facet p3 p1 p2 distance=    0

These points either have a maximum or minimum x-coordinate, or
they maximize the determinant for k coordinates.  Trial points
are first selected from points that maximize a coordinate.

The min and max coordinates for each dimension are:
  0:      -0.5       0.5  difference=    1
  1:      -0.5       0.5  difference=    1
  2:      -0.5       0.5  difference=    1

If the input should be full dimensional, you have several options that
[...]

PR #1849, opened 2023-02-22
Handle coplanar points in oriented_bounds

pchorak:
A potential solution to issue https://github.com/mikedh/trimesh/issues/1848

These changes check for this specific issue in `oriented_bounds` by catching Qhull errors and confirming that the points are all co-planar (rank 2). If this is the case, the bounds are computed with a dedicated function that transforms the points onto the xy plane and then uses `oriented_bounds_2D` to find the bounds.

This solution feels a little excessive, so I would be very much open to suggestion. The Qhull error could also be avoided by adding `QJ` to the `qhull_options` of `trimesh.convex.convex.hull`, but I'm not sure we want to joggle points by default.

### DIFF
--- a/tests/test_bounds.py
+++ b/tests/test_bounds.py
@@ -98,7 +98,7 @@ class BoundsTest(g.unittest.TestCase):
             points = g.np.zeros((5, 3))
             points[:, :2] = g.np.random.random((points.shape[0], 2))
             rot, _ = g.np.linalg.qr(g.np.random.randn(3, 3))
-            points = points @ rot
+            points = g.np.matmul(points, rot)
             to_origin, extents = g.trimesh.bounds.oriented_bounds(points)
 
             assert g.trimesh.util.is_shape(to_origin, (4, 4))
--- a/tests/test_bounds.py
+++ b/tests/test_bounds.py
@@ -81,14 +81,41 @@ class BoundsTest(g.unittest.TestCase):
                 transformed_bounds = [transformed.min(axis=0),
                                       transformed.max(axis=0)]
 
-                for i in transformed_bounds:
+                for j in transformed_bounds:
                     # assert that the points once our obb to_origin transform is applied
                     # has a bounding box centered on the origin
-                    assert g.np.allclose(g.np.abs(i), extents / 2.0)
+                    assert g.np.allclose(g.np.abs(j), extents / 2.0)
 
                 extents_tf = g.np.diff(
                     transformed_bounds, axis=0).reshape(dimension)
                 assert g.np.allclose(extents_tf, extents)
+
+    def test_obb_coplanar_points(self):
+        """
+        Test OBB functions with coplanar points as input
+        """
+        for i in range(5):
+            points = g.np.zeros((5, 3))
+            points[:, :2] = g.np.random.random((points.shape[0], 2))
+            rot, _ = g.np.linalg.qr(g.np.random.randn(3, 3))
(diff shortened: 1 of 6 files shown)


mikedh:
This looks great thanks for the PR! I'll try to run and merge next week, my initial thoughts:

- I've actually been thinking about whether we should be storing `_convex_hull_raw` for meshes with just the raw output from `spatial.ConvexHull` because `oriented_bounds` doesn't need the face winding or normals to be consistant and repairing it is expensive. With recent speedups repairing the convex hulls is like 30% of `oriented_bounds` [benchmarked using this gist](https://gist.github.com/mikedh/e8043836b72931202b23ff183365b576).
- I like the catching `QHullError`!
- I'd probably move the coplanar case inside of `oriented_bounds` or rename it with a `_` prefix as end users should probably never be calling it?
- Are there major downsides to `QJ`? Open to adding it to the default args. The description seems pretty reasonable. I wasn't quite clear if it joggles it only when a precision issue occurs, or always:
> Option 'QJ' or 'QJn' joggles each input coordinate by adding a random number in the range [-n,n]. If a precision error occurs, It tries again. If precision errors still occur, Qhull increases n ten-fold and tries again. The maximum value for increasing n is 0.01 times the maximum width of the input. Option 'QJ' selects a default value for n. The header file [user_r.h](http://www.qhull.org/src/libqhull_r/user_.h#JOGGLEdefault) defines these parameters and a maximum number of retries. See [Merged facets or joggled input](http://www.qhull.org/html/qh-impre.htm#joggle).

pchorak:
Adding `_convex_hull_raw` sounds like a nice speedup. I took a stab at this [here](https://github.com/mikedh/trimesh/commit/ff906cadb4c0e005fece566596ec431e67343292), but something went wrong because that version is running way slower. It looks like `oriented_bounds` uses the normals, such as [here](https://github.com/mikedh/trimesh/blob/main/trimesh/bounds.py#L198) where if they are not repaired there end up being many more edges.

Moving the coplanar case inside `oriented_bounds` sounds good; I've made that change.

I don't think there are _major_ downsides to `QJ`. The description made it sound like it always joggles inputs, although in some quick testing, the results seem exact even with `QJ`. My only concern was that, since I don't know how people are using trimesh, there might be use cases where someone needs accuracy to more than 7 significant digits and joggling could affect it.

mikedh:
> Adding _convex_hull_raw sounds like a nice speedup. I took a stab at this [here](https://github.com/mikedh/trimesh/commit/ff906cadb4c0e005fece566596ec431e67343292), but something went wrong because that version is running way slower. It looks like oriented_bounds uses the normals, such as [here](https://github.com/mikedh/trimesh/blob/main/trimesh/bounds.py#L198) where if they are not repaired there end up being many more edges.

Oh that's a great point, computing normals is often surprisingly expensive and definitively not the sort of thing you want to be doing twice haha, I'll play with this as well.

> I don't think there are major downsides to QJ. The description made it sound like it always joggles inputs, although in some quick testing, the results seem exact even with QJ. My only concern was that, since I don't know how people are using trimesh, there might be use cases where someone needs accuracy to more than 7 significant digits and joggling could affect it.

Oh nice. Yeah have to think about it, 7 significant digits is pretty close to what the default `tol.merge` is, and having `convex_hull` work in more edge cases seems like a pretty reasonable tradeoff.

mikedh:
Thanks for the fix and followups!

PR #1851, opened 2023-02-22
Reduce name collisions to speed up loading of gltfs

pchorak:
`unique_name` can become incredibly slow because of [this loop](https://github.com/mikedh/trimesh/blob/main/trimesh/util.py#L2437) if given many colliding names. For example, if loading a model without geometry or node names, or with many repeated names, `unique_name` will identify that there is a conflict and then try appending and then incrementing an integer until the name is unique. For later names, when there are already many with appended integers, this approaches a double for loop over all the names.

The change here adds an optional argument to `unique_name` that can allow it to keep track of how far it has incremented for different input names and start from this point, significantly speeding it up in certain cases.

Description from first attempt
~~The changes here append a unique integer to each name before checking it's uniqueness, to significantly reduce the number of name collisions. The side effect is that all the geometry and node names have an integer appended. If it is important to preserve the original names if at all possible, another option would be to create a unique name generator that keeps track of base names that have been seen before with a counter for each.~~

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -1365,7 +1365,7 @@ def _read_buffers(header,
     # load data from accessors into Trimesh objects
     meshes = collections.OrderedDict()
 
-    names_original = collections.defaultdict(list)
+    start_increments = {}
 
     for index, m in enumerate(header.get("meshes", [])):
 
@@ -1389,9 +1389,8 @@ def _read_buffers(header,
                 attr = p['attributes']
                 # create a unique mesh name per- primitive
                 name = m.get('name', 'GLTF')
-                names_original[index].append(name)
                 # make name unique across multiple meshes
-                name = unique_name(name, meshes)
+                name = unique_name(name, meshes, start_increments)
 
                 if mode == _GL_LINES:
                     # load GL_LINES into a Path object
@@ -1463,7 +1462,6 @@ def _read_buffers(header,
                     # each primitive gets it's own Trimesh object
                     if len(m["primitives"]) > 1:
                         kwargs['metadata']['from_gltf_primitive'] = True
-                        name = unique_name(name, meshes)
                     else:
                         kwargs['metadata']['from_gltf_primitive'] = False
 
@@ -1545,9 +1543,13 @@ def _read_buffers(header,
     # to a dict comprehension as it will be checking
     # the mutated dict in every loop
     name_index = {}
+    start_increments = {}
     for i, n in enumerate(nodes):
         name_index[unique_name(
-            n.get('name', str(i)), name_index)] = i
+            n.get('name', str(i)),
+            name_index,
+            start_increments
+        )] = i
     # invert the dict so we can look up by index
     # node index (int) : name (str)
     names = {v: k for k, v in name_index.items()}
--- a/trimesh/util.py
+++ b/trimesh/util.py
@@ -2389,7 +2389,7 @@ def is_ccw(points):
     return ccw
 
 
-def unique_name(start, contains):
+def unique_name(start, contains, start_increments={}):
     """
     Deterministically generate a unique name not
     contained in a dict, set or other grouping with
@@ -2402,6 +2402,9 @@ def unique_name(start, contains):
       Initial guess for name.
     contains : dict, set, or list
       Bundle of existing names we can *not* use.
+    start_increments : dict
+      Maps name starts encountered before to increments in
+      order to speed up finding a unique name.
 
     Returns
     ---------
@@ -2414,13 +2417,13 @@ def unique_name(start, contains):
             start not in contains):
         return start
 
-    # start checking with zero index
-    increment = 0
+    # start checking with zero index unless found
+    increment = start_increments.get(start, 0)
     if start is not None and len(start) > 0:
         formatter = start + '_{}'
         # split by our delimiter once
         split = start.rsplit('_', 1)
-        if len(split) == 2:
+        if len(split) == 2 and increment == 0:
             try:
                 # start incrementing from the existing
                 # trailing value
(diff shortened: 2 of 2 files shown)


mikedh:
Thanks for the PR! Yeah `unique_name` can be very `n^2` unfortunately as it can't track the prefixes used already as a function. I'd prefer to keep the existing behavior on names if possible. One other option is if you switch the function to an object that can count up, testing on some synthetic data with duplicates you get a huge speedup:

```

import trimesh
import collections

from uuid import uuid4
from trimesh.util import unique_name
from pyinstrument import Profiler


def unique_new(start, contains):
    """
    """
    if len(start) > 0 and start not in contains:
        return start

    if len(start) == 0:
        start = 'geometry'

    precheck = [start + '_' + str(i) for i in range(len(contains))]

    from IPython import embed
    embed()


class UniqueName(object):
    """
    Keep track of names that have been assigned with a
    prefix scheme of `{name}_{integer}`
    """

    def __init__(self, contains=None):

        self.contains = set()
        if contains is not None:
            self.contains.update(contains)
        self.prefix = collections.Counter()

    def unique(self, name):
        if len(name) == 0:
            name = 'geometry'
        contains = self.contains

        if name not in contains:
            contains.add(name)
            return name

        # if the name is already of the
        split = name.rsplit('_', 1)
        if len(split) == 2 and str.isnumeric(split[1]):
            name = split[0]

        # keep track of incremented prefixes
        prefix = self.prefix
        prefix[name] += 1

        result = name + '_' + str(prefix[name])
        assert result not in contains
        self.contains.add(result)

        return result


if __name__ == '__main__':

    trimesh.util.attach_to_log()

    # get some random strings
    names = [uuid4().hex for _ in range(20)]
    # make it a whole lotta duplicates
    names = names * 1000

    # add a non-int postfix to test
    names.extend(['suppp_hi'] * 10)

    assigned = set()
    with Profiler() as P:
        for n in names:
            unique = unique_name(n, assigned)
            assigned.add(unique)
    P.print()

    assigned_new = set()
    tracker = UniqueName()
    with Profiler() as P:
        for n in names:
            assigned_new.add(tracker.unique(n))
    P.print()

    # new scheme should match the old one
    assert assigned_new == assigned
```

```

3.987 <module>  err.py:1
└─ 3.971 unique_name  trimesh/util.py:2405
   ├─ 2.065 [self]  None
   └─ 1.896 str.format  None
         [2 frames hidden]  <built-in>



  _     ._   __/__   _ _  _  _ _/_   Recorded: 16:48:51  Samples:  35
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.037     CPU time: 0.037
/   _/                      v4.4.0

Program: /media/psf/Dropbox/robotics/trimesh/err.py

0.036 <module>  err.py:1
├─ 0.030 UniqueName.unique  err.py:38
│  ├─ 0.024 [self]  None
│  ├─ 0.003 set.add  None
│  │     [2 frames hidden]  <built-in>
│  ├─ 0.002 len  None
│  │     [2 frames hidden]  <built-in>
│  └─ 0.001 str.rsplit  None
│        [2 frames hidden]  <built-in>
├─ 0.003 [self]  None
└─ 0.003 set.add  None
      [2 frames hidden]  <built-in>
```

pchorak:
Hi! Thanks for the quick response. Using an object also crossed my mind. I think the behavior of `UniqueName` and my change to `unique_name` are pretty similar, although I am counting individually for each name stem, which I think is closer to the original behavior. When you say you'd like to keep the existing behavior on names, do you mean to leave `unique_name` untouched or just maintain the behavior? I ditched my initial idea that changed the behavior when I realized it failed a bunch of the unit tests. The current version is not meant change the behavior of `unique_name`.

trimesh/exchange/gltf.py
```diff
@@ -1463,7 +1462,6 @@ def _read_buffers(header,
                     # each primitive gets it's own Trimesh object
                     if len(m["primitives"]) > 1:
                         kwargs['metadata']['from_gltf_primitive'] = True
-                        name = unique_name(name, meshes)
```
pchorak, line 1466: `name` is computed earlier in the loop and it doesn't look like there are any changes to `name` or `meshes` in between line 1394 and here

trimesh/exchange/gltf.py
```diff
@@ -1365,7 +1365,7 @@ def _read_buffers(header,
     # load data from accessors into Trimesh objects
     meshes = collections.OrderedDict()
 
-    names_original = collections.defaultdict(list)
```
pchorak, line 1368: `names_original` looked unused to me

mikedh:
Awesome thanks for the fix and followup! Tracking counts in an extra dict seems like a great solution, keeps the API the same but adds an optional speedup which we can use internally. Here's `unique_name` original vs. tracking the counts:


```
  _     ._   __/__   _ _  _  _ _/_   Recorded: 21:04:02  Samples:  3907
 /_//_/// /_\ / //_// / //_'/ //     Duration: 3.914     CPU time: 3.893
/   _/                      v4.4.0

Program: test_util.py

3.913 UtilTests.test_unique_name  test_util.py:191
`- 3.889 unique_name  trimesh/util.py:2405
   |- 2.035 [self]  None
   `- 1.841 str.format  None
         [2 frames hidden]  <built-in>



  _     ._   __/__   _ _  _  _ _/_   Recorded: 21:04:06  Samples:  58
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.059     CPU time: 0.058
/   _/                      v4.4.0

Program: test_util.py

0.058 UtilTests.test_unique_name  test_util.py:191
|- 0.051 unique_name  trimesh/util.py:2405
|  |- 0.038 [self]  None
|  |- 0.006 str.format  None
|  |     [2 frames hidden]  <built-in>
|  |- 0.003 str.rsplit  None
|  |     [2 frames hidden]  <built-in>
|  |- 0.003 len  None
|  |     [2 frames hidden]  <built-in>
|  `- 0.001 dict.get  None
|        [2 frames hidden]  <built-in>
|- 0.006 [self]  None
`- 0.001 set.add  None
      [2 frames hidden]  <built-in>

```

PR #1852, opened 2023-02-22
Pass through node extension data from gltfs

pchorak:
It would be great to pass through any node extension data from gltf files. Putting the extension data in `extras` seems simplest since the field appears unused for gltfs (it looks like the actual extras are put in `metadata`, is this intended?) but is [passed on when constructing the graph](https://github.com/mikedh/trimesh/blob/main/trimesh/scene/transforms.py#L75).

### DIFF
--- a/tests/test_gltf.py
+++ b/tests/test_gltf.py
@@ -585,12 +585,14 @@ class GLTFTest(g.unittest.TestCase):
             sphere1,
             node_name="Sphere1",
             geom_name="Geom Sphere1",
-            transform=tf1)
+            transform=tf1,
+            extras={'field': 'extra_data1'})
         s.add_geometry(sphere2,
                        node_name="Sphere2",
                        geom_name="Geom Sphere2",
                        parent_node_name="Sphere1",
-                       transform=tf2)
+                       transform=tf2,
+                       extras={'field': 'extra_data2'})
 
         # Test extras appear in the exported model nodes
         files = s.export(None, "gltf")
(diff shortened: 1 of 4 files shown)


mikedh:
Awesome thanks! Yeah the `scene.metadata` == `gltf.extras` is a little wonky, but it is nice because it preserves `scene.metadata` through a GLTF round-trip. 

Agreed it would be nice to preserve `extensions` and ideally it might even round-trip through gltf->trimesh->gltf. My only hangup is that the (`extensions == extras, extras == metadata`) mapping is already confusing me haha. 

I also don't totally recall where the `kwargs['extras']` end up, but I don't see them in the `trimesh.Scene` constructor or in `trimesh.exchange.load.load_kwargs`? One option is we could map `extensions` to `Scene.metadata['extensions']` (although then we have to worry about if there is an extra named extensions :).

pchorak:
Thanks for the review and thoughts! I'll take a look again and try to see where might be good to put 'extensions'. `Scene.metadata` would also be convenient, but you raise a good point about potential issues if there's an extra with that name.

pchorak:
The scene and mesh extras are making it into the scene and geometry metadata and through round-trips saving/loading .gltf files. However, it looks like the node extras [here](https://github.com/mikedh/trimesh/blob/main/trimesh/exchange/gltf.py#L1642) are not making it into the scene graph and thus through round-trips (lost when loading). I augmented one of the unit tests to fail because of this in 073ab83. I fixed the round-trip by putting the node extras in "extras" in 9c3858b, so that when the node `kwargs` are passed to `SceneGraph.update` by [handle_scene](https://github.com/mikedh/trimesh/blob/main/trimesh/exchange/load.py#L431), the extras are preserved by [this]([](https://github.com/mikedh/trimesh/blob/main/trimesh/scene/transforms.py#L76)) existing code.

Lastly, I added some new arguments and code to handle node `extensions` in a similar way to node `extras`.

pchorak:
Hi! Any thoughts/input on the current version of this? It should allow gltf _node_ extras and extensions to be loaded and saved, neither of which are supported in the current version (not be confused with the _scene_ and _mesh_ extras, which are supported in the current version via metadata).

pchorak:
I like the idea of unifying everything in metadata and made changes to do this. The node/edge metadata are still handled a little differently; since the edge data is a dictionary, the metadata are an item in the dictionary instead of an object attribute. I did change the key from 'extras' to 'metadata', and also changed the last argument of `Scene.add_geometry` accordingly.

mikedh:
Awesome thanks for the fixes!! Yeah having it as an item in the dictionary makes total sense. It's also super nice to round-trip extensions. I'll merge and then release in #1870

mikedh:
Any thoughts on putting extension in `scenes[0]` vs the root of the file as proposed in #1973?

pchorak:
Thanks for pinging me! I replied with some input in the new PR but am ok either way.

PR #1881, opened 2023-03-28
fix for ValueError: assignment destination is read-only

gabrielggg:
Hi, i'm getting this error,
    self.scene.camera_transform[...] = self.view['ball'].pose
ValueError: assignment destination is read-only

So the fix is to set the array to be writable. Please check this out. Thanks a lot in advance!

### DIFF
--- a/examples/widget.py
+++ b/examples/widget.py
@@ -1,5 +1,5 @@
 """
-Glooey widget example. Only runs with python>=3.6 (because of Glooey).
+Glooey widget example. Only runs with python>=3.6 and pyglet<=1.5.27 (because of Glooey).
 """
 
 import io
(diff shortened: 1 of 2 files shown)


mikedh:
Hey, I think this will confuse the scene-graph hashing as this will alter the camera transform array but without clearing cached transforms. Can you use the setter instead of altering the matrix in-place, i.e.:
```
self.scene.camera_transform = self.view['ball'].pose
```

Which is a shortcut for:
```
self.scene.graph[self.scene.camera.name] = self.view['ball'].pose
```

gabrielggg:
Ok, thanks, it works!

mikedh:
Great thanks for the fix!!

Issue #1880, opened 2023-03-28
PLY Number of vertices incorrect


whateverforever:
Hi there! I have a 3D-scanned ply file whose header specifies X number of vertices. However when loading the file in trimesh, the number of vertices is much greater, even with `process` and `validate` disabled. The number of faces is almost correct, but off-by-one.

The file is binary, and has per-wedge UV coords.

Any idea what is going on here, is this as intended?

```
ply
format binary_little_endian 1.0
comment TextureFile some_texture.png
element vertex 100793
property float x
property float y
property float z
property float nx
property float ny
property float nz
element face 200000
property list uchar int vertex_indices
property list uchar float texcoord
end_header
...snip...
```

```python
In [13]: obj_base = trimesh.load("model.ply")

In [14]: obj_base.vertices.shape
Out[14]: (145102, 3)

In [15]: obj_proc = trimesh.load("model.ply", process=False, validate=False)

In [16]: obj_proc.vertices.shape
Out[16]: (145102, 3)

In [17]: obj_base.faces.shape
Out[17]: (199999, 3)

In [18]: obj_proc.faces.shape
Out[18]: (199999, 3)
```

mikedh:
Hey, I think you may want `fix_texture=False`. If I recall correctly the PLY format texture is per-face and that required unmerging some vertices to make it not look gnarly with a per-vertex UV. For `fuze.ply` with this header:
```
ply
format binary_little_endian 1.0
comment VCGLIB generated
comment TextureFile   fuze uv.jpg
element vertex 502
...
element face 1000
```

It matches if you don't mind mangled texture:
```
In [13]: trimesh.load('fuze.ply', process=False)
Out[13]: <trimesh.Trimesh(vertices.shape=(664, 3), faces.shape=(1000, 3))>

In [14]: trimesh.load('fuze.ply', process=False, fix_texture=False)
Out[14]: <trimesh.Trimesh(vertices.shape=(502, 3), faces.shape=(1000, 3))>

```

whateverforever:
thank you for the quick reply! I'll have a look what it does to the texture, but that sounds exactly like what I'm looking for :)

jandromani:
Based on the header information you provided, the PLY file should have 100,793 vertices and 200,000 faces. However, when you load the file using trimesh, you get 145,102 vertices and 199,999 faces.

[...]

PR #1882, opened 2023-03-29
Fix ply uv export

whateverforever:
Hey,

while working on the same project as mentioned here https://github.com/mikedh/trimesh/issues/1880 , I noticed that the PLY export ignores the texture coordinates. This PR should fix that.


mikedh:
Looks great, thanks!

PR #1886, opened 2023-03-31
Changes how faces are parsed in obj to take into account edge cases

clemgaut:
g tag group can be included within the list of faces, which can cause parsing issues in rare cases

### DIFF
--- a/models/face_in_group_name_mid_file.obj
+++ b/models/face_in_group_name_mid_file.obj
@@ -1,0 +1,9 @@
+v 10 20 30
+v 15 5 25
+v 1 5 10
+v 5 15 20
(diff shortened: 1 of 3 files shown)


mikedh:
Looks great, thanks for the fix and test! My only concern was relative speed but this looked roughly identical on a 1mb OBJ.

clemgaut:
FYI I just did a test on a 42 Mb OBJ. The new method can be a bit slower: with trimesh version before my change it took 3.8s (average over 7 runs) while with my change it took 4.3s (average over 7 runs). The difference is not that big (about 15%) so I would think this is reasonable too.

Issue #1902, opened 2023-04-13
Dumping scenes with non-mesh geometries fails on concatenation 


thegnarwhals:
If a scene has non-mesh geometries, for example Path3Ds, `scene.dump(concatenate=True)` will fail. It is useful to have a way to still concatenate the meshes, skipping the others with a warning, and returning the result. I will implement this and create a PR.

thegnarwhals:
Relates to #1900 and #1901 somewhat

Issue #1900, opened 2023-04-13
Concatenation does not assert on type as intended


thegnarwhals:
In `util.concatenate()`, there is a comment that says we are asserting that the meshes have trimesh type. However, we are only asserting on the first mesh. This leads to unhelpful attribute errors when dumping scenes that contain e.g. Path3D geometries. I will create a PR for this.

PR #1903, opened 2023-04-13
Skip non-mesh geometries in scene dump concatenation

thegnarwhals:
See #1902

Relates to #1900 and #1901

If a scene has non-mesh geometries, for example Path3Ds, scene.dump(concatenate=True) will fail. It is useful to have a way to still concatenate the meshes, skipping the others with a warning, and returning the result. This change implements this.

### DIFF
--- a/trimesh/scene/scene.py
+++ b/trimesh/scene/scene.py
@@ -843,6 +843,16 @@ class Scene(Geometry3D):
             transform, geometry_name = self.graph[node_name]
             # get a copy of the geometry
             current = self.geometry[geometry_name].copy()
+
+            if concatenate:
+                try:
+                    util.type_named(current, 'Trimesh')
+                except ValueError:
+                    util.log.warning(
+                        'Skipping non-mesh geometry in concatenation'
+                    )
+                    continue
+
             # move the geometry vertices into the requested frame
             current.apply_transform(transform)
             current.metadata['name'] = geometry_name


mikedh:
Awesome thanks for the two fixes! I think concatenate might predate the existence of scenes with other geometries haha. Seems like the desirable behavior for concatenate would be:

- all one type `(Trimesh, Path3D, Path2D, PointCloud)` objects: returns that type
- mix of types: returns of the type first in the above ordered list, drops others, logs a warning. Slightly janky but seems unavoidable.
- mix of `(Path2D, Path3D)`: returns `Path3D`

Does that sounds right?

thegnarwhals:
Hi @mikedh - thanks for your response. I'm a little confused by the 

> mix of (Path2D, Path3D): returns Path3D

are you suggesting that we cast the Path2Ds into Path3Ds prior to concatenation? Is there an existing way to do this?

Also, can Path3Ds and PointClouds easily be concatenated?

mikedh:
> are you suggesting that we cast the Path2Ds into Path3Ds prior to concatenation? Is there an existing way to do this?

Yeah, every Path2D has a `to_3D` method which just puts it on the XY plane as a Path3D

> Also, can Path3Ds and PointClouds easily be concatenated?

Not currently, and it seems like they probably aren't concatenatable.

Thanks for the work on this!! I can make a pass on this after merging your fixes, I probably won't be able to get to it until late this week.

thegnarwhals:
Gotcha! Sounds good to me! I might have a chance to get on this tomorrow if you like

mikedh:
No worries I'll make a pass, thanks for the PR!

PR #1904, opened 2023-04-14
fix FSLeyes broken link

myd7349:
The original link of [fsleyes](https://users.fmrib.ox.ac.uk/~paulmc/fsleyes/userdoc/latest/quick_start.html) in README.md doesn't exist any more.


mikedh:

PR #1907, opened 2023-04-18
Release: Concat + Fixes

mikedh:
- release #1912 fixing caching bug on `TrackedArray`
  - added test which runs through all methods and tries to break our hashing, which caught the same bug on `{'put', 'itemset', 'byteswap', 'sort', 'fill', 'partition', 'setflags'}` which are now fixed. 
  - switched hash of `hash_function(object)` to `hash_function(object.tobytes(order='C'))` which seems slightly slower in some cases, but I'm a little unclear as to what properties the hash function was actually hashing before even though it worked. This also removes a conditional inside our check block as we're always hashing on the C-order version of the data.
- release #1904 fixing readme 
- release #1917 and also convert `trimesh.creation.uv_sphere` to use the `revolve` method which fixes a bookkeeping issue. 
- release #1903 and #1901 which update the concatenate logic to handle mixed values


louisnovy:
Ah nice test. I didn't realize some of those existed as in-place methods.

PR #1912, opened 2023-04-22
Ensure TrackedArray tracks in-place .sort() and returns a new hash

louisnovy:


mikedh:
Good catch, thanks for the fix!!

PR #1914, opened 2023-04-23
Caught another caching bug

louisnovy:
I did some minor cleanup to keep things more consistent and after running tests/test_cache.py a few times I've also realized it's non-deterministically failing. Setting the numpy.random seed doesn't change this. Any ideas?


mikedh:
Looks great thanks for the followup! 
> I've also realized it's non-deterministically failing. Setting the numpy.random seed doesn't change this. Any ideas?

Oh interesting, does this still happen non-deterministically if we replace `np.random.random` -> `g.random`? Which would probably be better even if it doesn't fix it haha.

louisnovy:
Still failing randomly for me. Printing out the args used shows they are identical...

louisnovy:
Looking at numpy docs I also found .setfield() which isn't getting caught...
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.setfield.html

PR #1917, opened 2023-05-01
[WIP] Convexity Check Patch

Eric-Vin:
The current `is_convex` check for meshes returns true for trimesh objects containing disjoint meshes that are individually convex. I've added a test case for this in `test_convex.py`. I also tried to do a quick patch for the problem by just checking `body_count` of the mesh, and returning false if it is greater than 1. This also causes an issue as it looks like the uv_sphere in the test has a body count of 125 (even though it appears to be a single body). Any advice for how to fix this issue?

### DIFF
--- a/tests/test_convex.py
+++ b/tests/test_convex.py
@@ -18,7 +18,10 @@ class ConvexTest(g.unittest.TestCase):
                   (True, g.trimesh.creation.box()),
                   (True, g.trimesh.creation.cylinder(
                       radius=1, height=10)),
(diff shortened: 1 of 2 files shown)


mikedh:
Looks great, thanks for the fix!! Good catch on the body count that's definitely a bug haha.  I just pushed a change to the release branch that changes `trimesh.creation.uv_sphere` from the old logic to use `trimesh.creation.revolve` which all the newer radially symmetric creation functions use and does the bookkeeping much more carefully.

PR #1918, opened 2023-05-03
Checks that only chunks having faces are added to faces tuples

clemgaut:
Covers a rare case where a chunk could contain 'f ' without having a face when loading an obj.

This rare case happens when a comment in a group having no faces contains 'f '.

The fix checks that at least one line in the chunk starts with 'f ', which is the proper definition of a face.

An alternative solution would be to remove comment lines (starting with #) but checking for the actual occurence of at least one face felt more appropriate.


mikedh:
Looks great, thanks for the fix and test!

PR #1921, opened 2023-05-06
Two problems were solved about exporting a mesh into a GLTF format.

tanupoo:
## version

>>> trimesh.__version__
'3.21.5'

## Problem 1

when I tried to export a mesh into a GLTF format, it failed.  This is because export_gltf() in gltf.py always return a dict.  But, export_mesh() expectes to have a string.

```
  File "/tmp/trimesh/test1.py", line 5, in <module>
    trimesh.exchange.export.export_mesh(mesh_tri, "mesh.gltf")
  File "/home/sakane/.local/lib/python3.10/site-packages/trimesh/exchange/export.py", line 86, in export_mesh
    result = util.write_encoded(file_obj, export)
  File "/home/sakane/.local/lib/python3.10/site-packages/trimesh/util.py", line 2099, in write_encoded
    file_obj.write(stuff)
TypeError: a bytes-like object is required, not 'dict'
```

The reproduction code is like below.

```
import trimesh

mesh_tri = trimesh.Trimesh(vertices=[[0, 0, 0], [0, 0, 1], [0, 1, 0]],
                           faces=[[0, 1, 2]])
trimesh.exchange.export.export_mesh(mesh_tri, "mesh.gltf")
```

## Problem 2

export_gltf() doesn't support to embed the buffer into the Uri.

## an idea to solve for the problem 1

Added a flag into export_gltf() to return a JSON string.  The default of the flag is False.  It means to return a dict to preserve the original behavior.

## an idea to solve for the problem 2

Added a flag into export_gltf() to embed the buffer into the URI.  The default of the flag is False.  It means to put the buffer into a different place in the dict to preserve the original behavior.

## proposal fix and the flags.

I made a pull request that solves above two problem.

When return_json_strin is true, it can export GLTF into a file.

```
import trimesh

mesh_tri = trimesh.Trimesh(vertices=[[0, 0, 0], [0, 0, 1], [0, 1, 0]],
                           faces=[[0, 1, 2]])
trimesh.exchange.export.export_mesh(mesh_tri, "mesh.gltf",
                                    return_json_string=True)
```

When embed_buffer is true, it embeds the buffer into the URI.

```
import trimesh

mesh_tri = trimesh.Trimesh(vertices=[[0, 0, 0], [0, 0, 1], [0, 1, 0]],
                           faces=[[0, 1, 2]])
trimesh.exchange.export.export_mesh(mesh_tri, "mesh.gltf",
                                    embed_buffer=True,
                                    return_json_string=True)
```

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -75,7 +75,9 @@ def export_gltf(scene,
                 include_normals=None,
                 merge_buffers=False,
                 unitize_normals=False,
-                tree_postprocessor=None):
+                tree_postprocessor=None,
+                embed_buffer=False,
+                return_json_string=False):
     """
     Export a scene object as a GLTF directory.
 
@@ -94,6 +96,10 @@ def export_gltf(scene,
       If passed will use to write each file.
     tree_postprocesser : None or callable
       Run this on the header tree before exiting.
+    embed_buffer : bool
+      Embed the buffer into the uri.
+    return_json_string : bool
+      return a JSON string if True, a dict if False.
 
     Returns
     ----------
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR!! Yeah probably a bunch of ways to do it but `mesh.export('hi.gltf')` should probably just write the buffer files next to the requested export, I added a quick fix and test here: https://github.com/mikedh/trimesh/pull/1907/commits/fc652034adfafed19787d07186c7f96cc761ed49

tanupoo:
Confirmed the problem 1 seems solved.  How about supporting embedding buffers into URI ?  Should I make another PR ?

mikedh:
Great! Oh yeah we support `base64` in the URI for load, you mean enabling it for export as well? Sure PR's welcome! I'd test with a round trip export -> import.

tanupoo:
submitted.  https://github.com/mikedh/trimesh/pull/1923

PR #1931, opened 2023-06-02
deterministic sample under trimesh

gMonika15:


mikedh:
Thanks for the PR, looks good!! Passing in a `seed=None` makes total sense. I think I will however switch the behavior here so that it always passes seed to `default_rng`, which would replicate the current non-deterministic behavior if seed is not passed, and will only be deterministic if you explicitly pass a `seed=10` or whatever.

mikedh:
Releasing in https://github.com/mikedh/trimesh/pull/1936 thanks for the PR!

I don't think `remove_close` is deterministic, so `sample_surface_even` then isn't either. `sample_surface` now is fully deterministic *if* you explicitly pass `seed=10` or other integer. If `remove_close` picked the minimum index or something it could probably be made to be deterministic.

PR #1939, opened 2023-06-09
Hot fix

gMonika15:
Check for the deterministic  over the code and it looks fine.

### DIFF
--- a/tests/test_sample.py
+++ b/tests/test_sample.py
@@ -76,8 +76,10 @@ class SampleTest(g.unittest.TestCase):
         m = g.get_mesh('featuretype.STL')
 
         # Deterministic execution
-        even_first, index_first = g.trimesh.sample.sample_surface_even(m, 1000)
-        even_last, index_last = g.trimesh.sample.sample_surface_even(m, 1000)
+        even_first, index_first = g.trimesh.sample.sample_surface_even(
+            mesh=m, count=1000, seed=42)
+        even_last, index_last = g.trimesh.sample.sample_surface_even(
+            mesh=m, count=1000, seed=42)
         
         assert (even_first == even_last).all()
         assert (index_first == index_last).all()
--- a/trimesh/sample.py
+++ b/trimesh/sample.py
@@ -52,7 +52,7 @@ def sample_surface(mesh, count, face_weight=None, sample_color=False, seed=None)
 
     # seed used to provide deterministic values
     if seed is None :
-        seed = np.random.randint(1)
+        seed = np.random.randint(100)
     # cumulative sum of weights (len(mesh.faces))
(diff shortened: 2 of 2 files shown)


mikedh:
Hey, if you want the deterministic behavior you have to explicitly pass a value to seed. I don't think that being deterministic all the time is desirable, and by passing `seed: Optional[int] = None` through to the RNG we match the behavior of many numpy functions that use random. To get deterministic results from `sample_surface`: 

```
In [1]: import trimesh

In [2]: m = trimesh.creation.box()

In [4]: trimesh.sample.sample_surface?
Signature:
trimesh.sample.sample_surface(
    mesh,
    count,
    face_weight=None,
    sample_color=False,
    seed=None,
)
Docstring:
Sample the surface of a mesh, returning the specified
number of points

For individual triangle sampling uses this method:
http://mathworld.wolfram.com/TrianglePointPicking.html

Parameters
-----------
mesh : trimesh.Trimesh
  Geometry to sample the surface of
count : int
  Number of points to return
face_weight : None or len(mesh.faces) float
  Weight faces by a factor other than face area.
  If None will be the same as face_weight=mesh.area
sample_color : bool
  Option to calculate the color of the sampled points.
  Default is False.
seed : None or int
  If passed as an integer will provide deterministic results
  otherwise pulls the seed from operating system entropy.

Returns
---------
samples : (count, 3) float
  Points in space on the surface of mesh
face_index : (count,) int
  Indices of faces for each sampled point
colors : (count, 4) float
  Colors of each sampled point
  Returns only when the sample_color is True
File:      /media/psf/Dropbox/robotics/trimesh/trimesh/sample.py
Type:      function

In [5]: trimesh.sample.sample_surface(m, count=2, seed=1)
Out[5]: 
(TrackedArray([[0.44864945, 0.35584039, 0.5       ],
               [0.5       , 0.18816855, 0.07667355]]),
 array([ 6, 11]))

In [6]: trimesh.sample.sample_surface(m, count=2, seed=1)
Out[6]: 
(TrackedArray([[0.44864945, 0.35584039, 0.5       ],
               [0.5       , 0.18816855, 0.07667355]]),
 array([ 6, 11]))

In [7]: trimesh.sample.sample_surface(m, count=2)
Out[7]: 
(TrackedArray([[-0.4429452 ,  0.5       , -0.11927735],
               [ 0.11918517, -0.28645715, -0.5       ]]),
 array([7, 3]))

In [8]: trimesh.sample.sample_surface(m, count=2)
Out[8]: 
(TrackedArray([[-0.31725414,  0.5       ,  0.45224329],
               [-0.47652398,  0.25286158,  0.5       ]]),
 array([7, 4]))


```

PR #1940, opened 2023-06-09
Add Support for Empty OpenSCAD Outputs

Eric-Vin:
OpenSCAD currently raises an error if the intersection of two meshes is empty. This PR changes the way subprocess is used, capturing the output of the call and including it in any CalledProcessError that is raised. This allows better handling of the above error, replacing it with an empty Trimesh object (similar to how Blender handles this). I also added a test to confirm the crash no longer occurs.

### DIFF
--- a/tests/test_boolean.py
+++ b/tests/test_boolean.py
@@ -78,6 +78,20 @@ class BooleanTest(g.unittest.TestCase):
             assert g.np.isclose(r.volume,
                                 8.617306056726884)
 
+    def test_empty(self):
+        engines = [
+            ('blender', g.trimesh.interfaces.blender.exists),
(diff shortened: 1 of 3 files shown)


Eric-Vin:
Whoops, looks like there are some formatting issues. I'll fix those.

Eric-Vin:
@mikedh I just wanted to follow up on this PR. Does this look ok?

mikedh:
Thanks for the fix this looks great! I'll merge on next release after testing. 

I have been hoping to start de-scoping the `trimesh.interfaces` submodule if [some of the things here](https://github.com/mikedh/trimesh/issues/1947) start becoming real options for booleans but for now OpenSCAD is the only somewhat reliable method haha.

PR #1946, opened 2023-06-15
Added method for planar intersection remeshing

kleinsimon:
this method allows to remesh at the intersection with a plane. This is useful for deformation and segmentation algorithms, i.e. morphing a mesh bound to a spline

### DIFF
--- a/tests/test_remesh.py
+++ b/tests/test_remesh.py
@@ -273,6 +273,24 @@ class SubDivideTest(g.unittest.TestCase):
                 assert (bary > -eps).all()
         test([0, 1])
         test([1, 0])
+        
+    def test_plane(self):
+        # try plane remesh on some primitives
+        meshes = [g.trimesh.creation.box(),
+                  g.trimesh.creation.icosphere()]
+
+        for i,m in enumerate(meshes):
+            s = m.subdivide_plane((1,0,0), (0,0,0))
+            # shouldn't have subdivided in-place
+            assert len(s.faces) > len(m.faces)
+            # area should be the same
+            assert g.np.isclose(m.area, s.area)
+            # volume should be the same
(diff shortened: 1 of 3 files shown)


mikedh:
Awesome thanks for the PR and test! The logic seems pretty similar to [`trimesh.intersections.slice_faces_plane`](https://trimsh.org/trimesh.intersections.html#trimesh.intersections.slice_faces_plane), I was wondering if it would be possible to use the outputs of that to simplify this function, or split out the logic common to the two functions? Then we get the benefits of testing `slice_faces_plane` in this function as well.

kleinsimon:
Hi thanks for your response. Yes, it is a modified version of this method. I first tried to insert this as a flag like "keep_both", but since the sequence is rather different, it was merely two methods with a switch. If you want to, I could give it a second try, now that it seems to be functional.

mikedh:
Awesome yeah if you wouldn't mind factoring it use `slice_plane` (maybe pulling some of the internals into a simpler function that returns only numpy arrays or something?). From a maintenance perspective people find bugs in `slice_plane` all the time hahah and it would be really great to have the fiddly logic in one place. 

As a cheesy implementation this works on a simple test data:
```
    m = trimesh.load('models/machinist.3DXML', force='mesh')
    m.process(merge_tex=True, merge_norm=True)

    assert m.is_volume

    # random vector                                                                                 
    vector = np.array([-0.88284,  0.42343, -0.20321])
    origin = m.center_mass

    r = trimesh.util.concatenate([
	m.slice_plane(origin, vector), m.slice_plane(origin, -vector)])
    # this is necessary because we're running slice twice independently and indexes 
    # aren't preserved
    r.process(merge_norm=True, merge_tex=True)

    assert r.is_volume
```

PR #1954, opened 2023-06-22
fix: re-enable passing facet color by color keyward

HiroIshida:
After this commits, we cannot create icosphere with `color` keyward.  
https://github.com/mikedh/trimesh/commit/147b1774da51c735d5fa4daa02260ea146b5dfd7

Although, we can do that with `facet_colors` keyward arg thanks to https://github.com/mikedh/trimesh/pull/1942, it still has been lose backward compatibility. 

This PR recover the compatibility with 3.22.0.


mikedh:
Sounds good, happy to do the prior deprecation procedure used for hashes, where it'll throw a `DeprecationWarning` if color is passed, and then remove it in a year or so.

PR #1957, opened 2023-06-26
add draco support for gltf

seed93:
only support for Linux currently


mikedh:
Thanks for the PR! Unfortunately trimesh is a pure Python library, so I can't merge this without breaking that promise to the users. 

For compiled stuff I try to only add a dependency if it has its own package with wheels for Mac/Win/Linux Python 3.6-present and is maintained. I recently set up a `cibuildwheel` pipeline for [embreex](https://github.com/mikedh/embreeX/blob/main/pyproject.toml) as we were getting a ton of issues about being able to install it. That might be a good template if you want to set up a self contained `gltf_draco_primitive` package which we could import. 

So one path to getting this in would be to:
- Create a new package, something like `gltf_draco_primitive`
- Set up cibuildwheel Mac/Win/Linux Python 3.6-3.11
- Add an import in the trimesh gltf loader wrapped in a try block.

PR #1960, opened 2023-06-28
Expand uesr and vars in path

trhermans:
There's an issue in searching for the v-hacd executable if the directory in `$PATH` is defined using any environment variables or the `~` abbreviation for the home user.

E.g. I put `TestVHACD` in `~/bin` and had it listed in my `$PATH` variable as such but trimesh couldn't find it. 

It looks like this also effects other external interfaces as well, but I haven't tried to test those.

I just through together a quick 1-line fix for v-hacd, but it would be easy to extend to the others if desired.

### DIFF
--- a/trimesh/interfaces/vhacd.py
+++ b/trimesh/interfaces/vhacd.py
@@ -5,10 +5,8 @@ from .generic import MeshScript
 from ..constants import log
 from ..util import which
 
-_search_path = str.join(
-    ":",
-    [os.path.expanduser(os.path.expandvars(p)) for p in os.environ["PATH"].split(":")],
-)
+_search_path = ":".join([os.path.expanduser(os.path.expandvars(p)) for p in os.environ["PATH"].split(":")])
+
 
 if platform.system() == 'Windows':
     # split existing path by delimiter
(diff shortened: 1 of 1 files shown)


mikedh:
Awesome thanks for the PR!! Do the current bindings for vhacd still work? I saw they re-wrote much of it and I was wondering if vhacd in trimesh should be either updated or deprecated, as it's not tested at all currently.

trhermans:
The bindings do not work with the updated vhacd, which I found out this morning after resolving this issue.

I wrote a totally hacked version that showed it's possible to run it, but I now see I never pushed it.
The main issue is you can't set the output file location with the command line app, so reading it in would require some reworking of the `MeshScript` class to get working. 

It's not super critical for my workflow right now, I was just trying to figure out why a unit test was failing, but it's probably worth making an issue about it and I might be able to make a fix at some point this summer.

mikedh:
Sounds good, thanks for the PR! I wonder if `pyVHACD` would be a better dep than calling the CLI, just checking quickly they set up `cibuildwheel` and it meets the "have wheels for mac/win/linux Py3.6+" test: https://pypi.org/project/pyVHACD/0.0.2/#files

I haven't installed or tested however, and I'm not sure if they plan to maintain it going forward.

PR #1973, opened 2023-07-11
bubble up the 'extension' to root

ankuPRK:
# Overview

This change will write the `extensions` provided with `scene.metadata` at the top-most level, so that assets such as GLTF lights are written at the correct level.

https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_lights_punctual/README.md#adding-light-instances-to-nodes

This enables writing lights which are exportable in Blender.

# Example code:

We try to add a single red point-light in an empty scene, and then import it in Blender. The code:

```
import trimesh
import numpy as np

scene = trimesh.Scene()
scene_metadata = {"gltf_extensions": {"KHR_lights_punctual": {"lights": []} }}

# add a light
scene_metadata["gltf_extensions"]["KHR_lights_punctual"]["lights"].append(
    {
        "name": "light1",
        "color": [1.0,0.0,0.0],
        "intensity": 50,
        "type": "point",
        "range": 10
    })
light_metadata = {"gltf_extensions": {"KHR_lights_punctual": {"light": 0}}}
mesh = trimesh.Trimesh() # a placeholder empty mesh for the light
scene.add_geometry(mesh, node_name="light1", transform=np.eye(4), metadata=light_metadata)

# update the scene with 
scene.metadata.update(scene_metadata)

# write the scene
scene.export("./single_light.gltf")
```

# Output of the current main
```
{
    "scene": 0,
    "scenes": [
        {
            "nodes": [
                0
            ],
            "extras": {},
            "extensions": {
                "KHR_lights_punctual": {
                    "lights": [
                        {
                            "name": "light1",
                            "color": [
                                1.0,
                                0.0,
                                0.0
                            ],
                            "intensity": 50,
                            "type": "point",
                            "range": 10
                        }
                    ]
                }
            }
        }
    ],
    "asset": {
        "version": "2.0",
        "generator": "https://github.com/mikedh/trimesh"
    },
    "nodes": [
        {
            "name": "world",
            "children": [
                1
            ]
        },
        {
            "name": "light1",
            "extensions": {
                "KHR_lights_punctual": {
                    "light": 0
                }
            },
            "extras": {}
        }
    ],
    "extensionsUsed": [
        "KHR_lights_punctual"
    ]
}
```
When we import this gltf file in Blender, we get this error message:
image

Which means that extensions needs to be at the top-most level of the tree.

# Output of this branch

This GLTF json is shown below, followed by the visualization on importing in Blender:
```
{
    "scene": 0,
    "scenes": [
        {
            "nodes": [
                0
            ],
            "extras": {}
        }
    ],
    "asset": {
        "version": "2.0",
        "generator": "https://github.com/mikedh/trimesh"
    },
    "extensions": {
        "KHR_lights_punctual": {
            "lights": [
                {
                    "name": "light1",
                    "color": [
                        1.0,
                        0.0,
                        0.0
                    ],
                    "intensity": 50,
                    "type": "point",
                    "range": 10
                }
            ]
        }
    },
    "nodes": [
        {
            "name": "world",
            "children": [
                1
            ]
        },
        {
            "name": "light1",
            "extensions": {
                "KHR_lights_punctual": {
                    "light": 0
                }
            },
            "extras": {}
        }
    ],
    "extensionsUsed": [
        "KHR_lights_punctual"
    ]
}
```
This GLTF is nicely importable in Blender, and produces the expected result:

image

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -650,7 +650,7 @@ def _create_gltf_structure(scene,
             extensions = tree['scenes'][0]['extras'].pop(
                 'gltf_extensions', None)
             if isinstance(extensions, dict):
-                tree['scenes'][0]['extensions'] = extensions
+                tree['extensions'] = extensions
         except BaseException:
             log.debug(
                 'failed to export scene metadata!', exc_info=True)
@@ -703,9 +703,9 @@ def _create_gltf_structure(scene,
 
     extensions_used = set()
     # Add any scene extensions used
-    if 'extensions' in tree['scenes'][0]:
+    if 'extensions' in tree:
         extensions_used = extensions_used.union(
-            set(tree['scenes'][0]['extensions'].keys()))
+            set(tree['extensions'].keys()))
     # Add any mesh extensions used
     for mesh in tree['meshes']:
         if 'extensions' in mesh:
@@ -1742,8 +1742,7 @@ def _read_buffers(header,
         # use a try except to avoid nested key checks
         if "metadata" not in result:
             result["metadata"] = {}
-        result['metadata']['gltf_extensions'] = header['scenes'][
-            header['scene']]['extensions']
+        result['metadata']['gltf_extensions'] = header['extensions']
     except BaseException:
         pass
 


mikedh:
Awesome thanks for the fix! Seems like this gets a little tricky since extensions can be on any element in a gltf file at any level, and presumably the original PR adding extension support needed an extension on the GLTF `scene`? I wonder if we could record the original key on load somehow so the extensions fully passed through a load-export cycle?

It's a shame `gltf_validator` or `jsonschema` don't catch this for whatever reason. Happy to merge this, although is there a small sample model we could add in a unit test to verify the extension makes it through a round trip in the right place?

ankuPRK:
Hi @mikedh sure I can add a simple unit-test to ensure that round-trip is happening correctly.

I would need some time to look at the other extensions (I tested it only on `KHR_lights_punctual`) to make sure that the change isn't too specific to Blender. 

The extensions for other items (like meshes, materials etc) shouldn't get affected by this change though.

pchorak:
> presumably the original PR adding extension support needed an extension on the GLTF `scene`

I actually moved away from using the scene extensions in favor of something else, so changing the extension behavior isn't a problem for me.

That said, putting extensions from `scene_metadata` in the top-level of the gltf file could cause confusion since both exist in the specification:
- Scene extensions: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-scene
- Top-level extensions: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-extension

PR #1977, opened 2023-07-17
fix: add fallback if PATH is not set

tpank:
Hey @mikedh,

we encountered a problem regarding the PATH variable: In the `interfaces` module, the content of the PATH environment variable is used as a base for the search path for executables (as it should be). However, on systems where the variable is not set, this behavior leads to a `KeyError` when loading the module.

I replaced `os.environ["PATH"]` with `os.environ.get("PATH", "")` to fix this.

### DIFF
--- a/trimesh/interfaces/blender.py
+++ b/trimesh/interfaces/blender.py
@@ -7,7 +7,7 @@ from ..constants import log
 import os
 import platform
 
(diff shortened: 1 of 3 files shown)


mikedh:
Thanks for the fix good catch! I added a test and will release shortly.

PR #1980, opened 2023-07-18
Specular/Glossiness to Metallic/Roughness pbr material conversion

dabeschte:
Hi!
I implemented a conversion from specular glossiness PBR materials to metallic roughness.

After trying out on two different GLB files, the code got a little more messy than I hoped for.
But this is mostly caused by non-standard textures being used in those models. 

I followed the information from [here](https://kcoley.github.io/glTF/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/) as a reference, but most of the time, the actual textures are a bit different, e.g. using 1, 2, or 3 channels for specularGlossinessTexture instead of 4 as defined in this reference document. Because of that, I am not really sure if I always read every file correctly, but it did work fine for me in these examples.

Still, I would like to ask you for a review, because I think that this PR might come in quite handy for a lot of us people who use trimesh to parse 3d scenes.

### DIFF
--- a/tests/test_gltf.py
+++ b/tests/test_gltf.py
@@ -300,6 +300,37 @@ class GLTFTest(g.unittest.TestCase):
     def test_specular_glossiness(self):
         s = g.get_mesh('pyramid.zip')
         assert len(s.geometry) > 0
+        assert 'GLTF' in s.geometry
+
+        mat = s.geometry['GLTF'].visual.material
+        assert isinstance(mat, g.trimesh.visual.material.PBRMaterial)
+
+        color = g.np.array(mat.baseColorTexture)[:, :, :3]
+        assert color.shape[0] == 84 and color.shape[1] == 71
+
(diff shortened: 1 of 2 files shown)


dabeschte:
Thanks for the quick reply. 
I implemented a test case and it fortunately already lead me to a bug in the conversion script which I also fixed now.

pillow is now imported locally as done in other functions - including a simple fallback.

mikedh:
Sweet thanks for the tests! Looks like `test_corpus` [caught something](https://github.com/mikedh/trimesh/actions/runs/5597278003/jobs/10246350202?pr=1980) on [this test model](https://github.com/assimp/assimp/tree/master/test/models/glTF2/BoxTextured-glTF-pbrSpecularGlossiness).

dabeschte:
Fixed it. Was a simple shape mismatch for grayscale textures since they are of shape WxH instead of WxHxC which I expected.

To be honest, it is likely that there are still some bugs in there which I just did not run into yet with my own test models. If some issues pop up after this PR, you can always reach out to me.

And thanks for the great work and the quick replies! :)

mikedh:
Awesome thanks for the feature and fix!!

PR #1981, opened 2023-07-19
Fix wrong depth range

ciscorn:
I was having trouble with some inexplicable z-fighting and found an unnecessary use of `glDepthRange`.

`glDepthRange` is not about clipping, but about mapping the z range [-1, 1] in NDC space to the z buffer value range [0, 1].

The default values [0, 1] are fine for normal use, so this line can be removed entirely :)


mikedh:
Looks good, thanks for the PR! Yeah probably better for windowed to use the defaults.

PR #1982, opened 2023-07-22
Fix properties in PBRMaterial not in _data

liamjwang:
Some of the properties of PBRMaterial were not being stored in the _data field. 
This caused these properties to not be copied to the new object in the copy() method.


mikedh:
Awesome thanks for the PR!! If you have a chance to add a test in `tests/test_gltf.py` that would be sweet otherwise I can do it no worries and I'll include in #1986

Issue #1511, opened 2022-03-06
Read texture mapped vertex colors from 3dxml


costerwi:
The example models/cube1.3dxml includes 1D TextureCoordinates within AbaqusGeometry.3drep referring to colors stored in image.bmp.

It would be extremely useful to be able to read these vertex colors into the model and ultimately export as glb.

Thanks for considering!

costerwi:
Completed with 2617bd4

PR #1985, opened 2023-07-26
3dxml texture coordinates

costerwi:
This PR adds texture mapping to the 3dxml format file import as requested in #1511. The example models/cube1.3dxml includes this feature.

It also adds some flexibility to the format, such as allowing multiple Faces and Face elements to be parsed.

Thanks for your consideration!
Carl


mikedh:
Awesome thanks for the PR! Would you happen to have a small model we can add to unit tests which checks this code path by any chance?

costerwi:
Thanks! I added one model and a few more tests to the existing test_3dxml.py.

Issue #1951, opened 2023-06-19
Dropping Python 2.7 For 4.0.0 Release


mikedh:
My preference would be to keep 2.7 compatibility going forward as there was no real reason to drop it, but in the last few months the ecosystem around it has been increasingly on its last legs. With [this change from Github](https://github.com/actions/setup-python/issues/672) someone would need to do a fair bit of CI work to keep things going on 2.7. According to [these stats](https://pypistats.org/packages/trimesh) Python 2.7 is currently around 0.75% of downloads for trimesh. And thus I think the time has come to make Python 3.6 the minimum supported version of Python sometime in the next month or two. 

- I'll probably sooner rather than later drop Python 2.7 from CI as all releases are blocked until that happens. 
- Release a `py2` version to PyPi so `pip install trimesh` on Python2 works going forward with the current version being the last. 
- Add type hints semi-automatically using [monkeytype](https://github.com/Instagram/MonkeyType) and [pytest-monkeytype](https://pypi.org/project/pytest-monkeytype/). Downside so far is they add a ton of ugly imports some of which are circular and break the library.
- Switch everything possible to `pyproject.toml` and remove most logic from `setup.py`

aqeelat:
How can I help here?

PR #1990, opened 2023-08-02
fix: use format instead of fstring to compatible with 2x

HiroIshida:
This PR recovers compatibility with python 2.x


mikedh:
Thanks for the hotfix! You may want to add `<4` to your version filter for the upcoming https://github.com/mikedh/trimesh/issues/1951

Issue #1989, opened 2023-07-28
Bug when copying trimesh.primitives: Density is not copied


clemense:
In previous versions (e.g. `3.15.5`) this worked:
```
import trimesh
a = trimesh.primitives.Box()
a.density = 2.0
print(a.density)
print(a.copy().density)
```
Output:
```
2.0
2.0
```
It doesn't work in newer versions (e.g. the latest `3.22.5`, and also `3.20.1`). The same code produces the following output:
```
2.0
1.0
```

clemense:
The same is true for `metadata`, the behavior also changed recently:
```
import trimesh
a = trimesh.primitives.Box()
a.metadata['foo'] = 'bar'
print(a.metadata)
print(a.copy().metadata)
```
Output:
```
{'foo': 'bar'}
{}
```

mikedh:
Yeah the behavior of primitives changed in https://github.com/mikedh/trimesh/pull/1942 due to some hard to track bugs for example when someone applied `face_colors=...` in the constructor.

Looking at the `center_mass` and `density` overrides they aren't using either the `_cache` or `_data` system as they were special cased (although they probably shouldn't be just assigning to an internal variable) which also confuses this issue. Seems like a fix might be:

- `density` when overridden gets applied to `data` which dumps the cache (or selectively dumps just mass_properties), same for the `center_mass` override
- `primitive.copy` gets a test to make sure density override survives

PR #1999, opened 2023-08-09
Add density, center_mass, and metadata during primitive copy

clemense:
Addresses #1989 

* New test for `Primitive.copy()` that includes `density`, `center_mass`, and `metadata`
* New setter for `density` and `center_mass` to add to `_data`
* `metadata` is copied in `Primitive.copy()`  (shallow)


mikedh:
Looks great thanks for the fix! I like the strategy of adding it to data when set otherwise getting from mass_properties, I think I'll propagate that to `Trimesh` as well.

PR #2006, opened 2023-08-18
Preserve texture on slice_plane

ibrarmalik:
For the `slice_plane` method, I interpolate the UVs for the new vertices by retrieving the barycentric coordinates of the point wrt. the original face.

I'm not sure what to do when `cap=True`. For now, I'm just ignoring it and not returning any UVs when this happens.

An example:

Screenshot 2023-08-18 at 16 34 53
Screenshot 2023-08-18 at 16 35 43

### DIFF
--- a/tests/test_section.py
+++ b/tests/test_section.py
@@ -291,6 +291,32 @@ class SliceTest(g.unittest.TestCase):
 
         assert len(sliced.faces) > 0
 
+    def test_textured_mesh(self):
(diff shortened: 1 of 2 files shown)


mikedh:
This looks great thanks for the PR and test! Yeah I think ignoring `cap=True` makes sense, or you could stack zero uv's for the cap part. I'll release in #2008

PR #2015, opened 2023-08-30
specular_to_pbr: fixed perceived brightness calculation

dabeschte:
I corrected the perceived brighntness calculation and now use the correct reference solution's brightness from Babylon.
It slipped through the test case, because the difference is quite small


mikedh:
Thanks for the followup fix!! Should we tighten the `atol` in the test maybe or add an example that's further from zero so it hits this?

mikedh:
Sounds good, thanks for the fix!

PR #2016, opened 2023-08-30
keep vertex normals when fusing meshes

dabeschte:
Hi!

I am not sure if there are any potential failure cases for this. 
But I am surprised that vertex normals (and other vertex attributes as well actually) are discarded when fusing meshes.

Anyways, it does do the trick for my use case. But let me know if there are some corner cases that could make issues!

### DIFF
--- a/trimesh/util.py
+++ b/trimesh/util.py
@@ -1503,6 +1503,10 @@ def concatenate(a, b=None):
     if all('face_normals' in m._cache for m in is_mesh):
         face_normals = np.vstack(
             [m.face_normals for m in is_mesh])
+        
+    # always save vertex normals
+    vertex_normals = vstack_empty(
+        [m.vertex_normals.copy() for m in is_mesh])
 
     try:
         # concatenate visuals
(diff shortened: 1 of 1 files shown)


mikedh:
Yeah the current data model is a little confusing, the vertex normals are currently thrown away because they are auto-generated only when requested. I think for the current data model in this PR it should check to see if any mesh has them:

- probably should only be included `if any('vertex_normals' in m._cache for m in meshes)` 
  - this prevents auto-generating them just for the concatenate

I was debating changing the data model for `vertex_normals`, `face_normals`, etc to:
- if auto-generated gets put into `mesh._cache`
- is the setter is called explicitly it gets put into `mesh._data`
 
I played with this a little in https://github.com/mikedh/trimesh/commit/948cdd7be659f826ba405da96b8f21ca14782770 but the major change to the data model and inevitable subtle bugs make me a little hesitant haha. Thanks for the PR!!

dabeschte:
Ah, I did not know that they were auto-generated.  tried to include them only if they are in m._cache at first, but in contrast to face normals, the vertex normals are not in there.

I am also not sure how to best solve this. but currently, the exported fused meshes only include face normals.

PR #2017, opened 2023-08-30
retain PBR properties when fusing materials

dabeschte:
Loading a mesh with trimesh.load(..., force="mesh") quietly converts PBR materials to SimpleMaterials. This PR changes that behaviour and keeps the PBR material properties alive.

Furthermore, it adds somefixes for the fusion process:
- UV coordinates of exactly 1.0 are not wrapped (I had triangles with uv coordinates (0,0) (0,1) (1,1) that were changed to (0,0) (0,0) (0,0) due to the module operator `uv = uv % 1.0`)
- individual textures in the trimsheet are now padded by a single pixel replicating the nearest neighbor
- setting the 'rotate' parameter of `rectangles_single`

I am not done with this PR yet, and I want to add test cases. So far, I only tested it on meshes with PBR "factors", but not with PBR "textures".

But I wanted to open this PR in case you have some early comments/concerns.

TODO:
- test cases!


trimesh/visual/material.py
```diff
@@ -525,7 +525,7 @@ def normalTexture(self):
           Normal texture.
         """
         return self._data.get('normalTexture')
-
```
dabeschte, line 528: sorry for these. But I used 
```
autopep8 --recursive --verbose --in-place --aggressive trimesh
flake8 trimesh
```
as mentioned on the contributing page. But I did not check in the changes in other files. Seems like you are now using ruff instead

trimesh/visual/material.py
```diff
@@ -793,25 +852,92 @@ def material_to_img(mat):
...
     assert len(uvs) == len(materials)
 
+    use_pbr = any(isinstance(m, PBRMaterial) for m in materials)
+
     # collect the images from the materials
-    images = [material_to_img(materials[g[0]]) for g in mat_idx]
+    images = [get_base_color_texture(materials[g[0]]) for g in mat_idx]
+    unpadded_sizes = [np.array(img.size) for img in images]
+
+    if len(images) <= 1:
```
dabeschte, line 861: I am not 100% sure about the padding. trimsheet islands are always difficult to handle and the padding can fix interpolation between different islands.
But when I add 2 pixels to a 1024x1024 texture, the output will be a 2048x2048 texture, because we use the next power of two

trimesh/path/packing.py
```diff
@@ -459,7 +459,11 @@ def rectangles(extents,
         # run a single insertion order
         # don't shuffle the first run, shuffle subsequent runs
         bounds, insert = rectangles_single(
-            extents=extents, size=size, shuffle=(i != 0))
+            extents=extents, 
+            size=size, 
+            shuffle=(i != 0),
+            rotate=rotate
```
dabeschte, line 465: this parameter was not set, resulting in occasional wrong coordinates and overlaps.
image
(image was truncated to focus on interesting part)

dabeschte: btw. some other bug that I was not able to find it causing the left most part to be unused. 
From my understanding, the `rectangles_single` function allocates space for this object marked in red [here in the code](https://github.com/mikedh/trimesh/blob/a2f89a6917d69e76914b09c7864acea3a5193f47/trimesh/path/packing.py#L208).
image

But then it starts the insertion with the largest texture which does not fit - so it allocates new space and inserts the largest texture _after_ the already allocated space.
But simply allocating 0 space also did not seem to solve it for me.

dabeschte: If you want to try it out yourself: I used [this example mesh](https://sketchfab.com/3d-models/wall-unit-620a4629d52e458fbaa7b05160547ddb). It is a great sample for testing the scene fusion function!

mikedh:
This looks awesome!! Yeah it would be sweet to extend the logic to PBR materials. Agreed on the need for test cases, the packing logic works for most things but I've definitely seen cases where it did something weird, I suspect it gets off by a pixel or something.

Let me know when you're ready for a review pass, thanks for the PR!

tests/test_texture.py
```diff
@@ -221,6 +221,24 @@ def test_pbr_export(self):
...
+        # and one with specular glossiness color values
+        m = g.get_mesh("pbr_cubes_emissive_spec_gloss.zip", force='mesh')
+
+        assert isinstance(m, g.trimesh.Trimesh)
+
+        mat_m = m.visual.material
+        assert isinstance(mat_m, g.trimesh.visual.material.PBRMaterial)
+        assert mat_m.baseColorTexture is not None
+        assert mat_m.metallicRoughnessTexture is not None
+        assert mat_m.emissiveTexture is not None
```
dabeschte, line 239: texture values are checked in the packing function

trimesh/visual/color.py
```diff
@@ -939,8 +939,8 @@ def uv_to_interpolated_color(uv, image):
     img = np.asanyarray(image.convert('RGBA'))
 
     colors00 = img[y_floor, x_floor]
-    colors01 = img[y_ceil, x_floor]
-    colors10 = img[y_floor, x_ceil]
+    colors01 = img[y_floor, x_ceil]
+    colors10 = img[y_ceil, x_floor]
```
dabeschte, line 943: a00 and a01 both use (1-dy) for Y interpolation - they also have to use the same y coordinates for the texture

trimesh/visual/material.py
```diff
@@ -820,21 +1075,53 @@ def material_to_img(mat):
...
-        # note this is only true for simple colors
-        # interpolation on complicated stuff can break this
-        assert (compare == check_flat).all()
-
-    return SimpleMaterial(image=final), stacked
+            # it had to fill in with default data
+            material_textures_values = []
+            for texture_load_fn, _ in material_textures:
+                orig_img = texture_load_fn(mat)
+                current = color.uv_to_interpolated_color(image=orig_img, uv=uv)
```
dabeschte, line 1096: using the interpolated version to circumvent floating point rounding errors

dabeschte:
Sorry for taking so long, but I had to find some time to craft a good test case, improve the testing procedure and fix the remaining bugs. But now I am finally happy with this PR.

Let me know what you think. Btw. the [scifi girl](https://sketchfab.com/3d-models/scifi-girl-v01-96340701c2ed4d37851c7d9109eee9c0) proved to be quite a good test case again.
It contains high resolution textures, and many different PBR properties. The trimsheet even exceeded the maximum allowed pixel count in pillow :-D (I implemented an iterative downscaling of the largest textures to circumvent this issue).

However, alpha blending does not work anymore when fusing so many meshes, because the triangles are not rendered in correct order. I disabled alpha blending now for fused meshes.

original - converted to metallic roughness
image
with alpha blending (cropped because NSFW)
image
without alpha blending
image

mikedh:
Oops,

mikedh:
Thanks for the great work on this, tested some sample data and it looked good! I went down a little bit of a rabbit hole, if you get a chance would you mind checking to see if branch `rc1` (merged to `release-candidate` now) still behaves as you expect? 

I noticed a lot of time in the profiler was being spent in the pad image function, which I was able to eliminated by passing through the `spacing` kwarg in `packing.rectangles` through to `packing.images`, which since the function returns the resulting bounding boxes this theoretically avoids having to keep track of padded vs unpadded size and speeds it up a lot. 

I was struggling to get the old inline test to pass and it was so complex that I had a hard time figuring out what it was doing. I replaced the test with an in-line check against the final image which I *think* does most of what the original test did while being quite a bit simpler.

Scifi girl currently looks the same to me as when I loaded/showed from your branch, although the hair detail looks a tad screwy:

image

dabeschte:
Hey!
Sorry for taking so long again, but I finally got a change to try out your modified version.
Thanks for taking the time to refactor the code!
It looks like it is mostly correct, but there are some cases which cause problems:

e.g. this does not correct anymore. the original model does not contain any textures which means that I generate a bunch of 1x1 textures for every material. The padding does not produce the same results as before (debugged all textures), but I assume that the UV coordinates should be correct. I essentially repeated the last pixel 1 more time on all sides of the texture. The new padding only adds a "neutral" pixel, but this leads to artifacts as can be seen in the examples below.
I also noted that you are now using quite 'atol=10' and the non-interpolated[ uv_to_color function](https://github.com/mikedh/trimesh/blob/b2e59eef61abf4b1435146ca37bf14a73fa94bb1/trimesh/visual/material.py#L1070), which can cause some problems. I really recommend to use interpolated color sampling to avoid rounding errors.

[0a11092e4d47496a99e9b42aaeb8f91f.zip](https://github.com/mikedh/trimesh/files/12844210/0a11092e4d47496a99e9b42aaeb8f91f.zip)

image
image
image

You can also see the problem with the 5 cubes test file that I made for this function. My original inline test should have detected this error in material values.
image
image
image

Also, I think that [this part](https://github.com/mikedh/trimesh/blob/b2e59eef61abf4b1435146ca37bf14a73fa94bb1/trimesh/visual/material.py#L1065) is not working correctly, because you wrap the 'new' coordinates using the offsets calculated with the old texture size. I'd suggest to wrap the texture coordinates before applying scale and offset.

[deduplicating texture](https://github.com/mikedh/trimesh/blob/b2e59eef61abf4b1435146ca37bf14a73fa94bb1/trimesh/visual/material.py#L932) also sounds like a bad idea, because some materials might share 1 property, but not the other? this is especially likely with the 1x1 pixel textures that we generate for properties without textures.


Not 100% sure, but I think [here](https://github.com/mikedh/trimesh/blob/b2e59eef61abf4b1435146ca37bf14a73fa94bb1/trimesh/visual/material.py#L1050) you have to divide by (final_size -1), because a UV coordinate of 1.0 refers to the last pixel, not the one after.


Please let me know if you need any more feedback or if I should create a new PR! But I hope the changes should be fairly easy to implement, except for the padding maybe.

mikedh:
Awesome thanks for taking a look! I just made a commit addressing some of your suggestions and followup PR's super welcome haha. 

- *I essentially repeated the last pixel 1 more time on all sides of the texture. The new padding only adds a "neutral" pixel, but this leads to artifacts as can be seen in the examples below.*
  - Gotcha I missed that behavior, yeah in packing it just places them into a new image with nothing next to them. Maybe the fix here is "single color" textures get a `2x2` pixel image? I just changed all of the "single color" images to a 2x2 and it seems to have gotten a little more stable
   
- *I also noted that you are now using quite 'atol=10' and the non-interpolated which can cause some problems. I really recommend to use interpolated color sampling to avoid rounding errors.*
  - I was having trouble getting the interpolated results to return anything close to the expected value on those gradients haha. Just switched to interpolated and lowered it to `atol=4` although yeah it would be great to get this tighter.
- *deduplicating texture also sounds like a potential cause for bugs, because some materials might share a property, but not the other? this is especially likely with the 1x1 pixel textures that we generate for properties without textures. The resulting textures could have different offsets and therefore require different uvs.*
  - Happy to revert this but it seemed to me like if we were wrapping correctly and 2 hash-identical images were packed this should be OK. 
- *I'd suggest to wrap the texture coordinates before applying scale and offset.*
  - Ahh great catch, wrapping post-scaling makes no sense, just pushed a change.

dabeschte:
> * _I essentially repeated the last pixel 1 more time on all sides of the texture. The new padding only adds a "neutral" pixel, but this leads to artifacts as can be seen in the examples below._
>   
>   * Gotcha I missed that behavior, yeah in packing it just places them into a new image with nothing next to them. Maybe the fix here is "single color" textures get a `2x2` pixel image? I just changed all of the "single color" images to a 2x2 and it seems to have gotten a little more stable

It does help, but it does not really solve the problem, unfortunately.
The corners pointing to the camera are a bit darker, because they got interpolated into neighboring regions.
We could fix that by manually adjusting the uv coordinates, but I wouldn't want to do this.
Also, I think that replicating the corner pixels is in general a better solution, because it also works for textures.
image

> * _deduplicating texture also sounds like a potential cause for bugs, because some materials might share a property, but not the other? this is especially likely with the 1x1 pixel textures that we generate for properties without textures. The resulting textures could have different offsets and therefore require different uvs._

>   * Happy to revert this but it seemed to me like if we were wrapping correctly and 2 hash-identical images were packed this should be OK.

I did a quick test and it turns out that we can't have multiple texture atlasses. The cubes above already show the problem (saw the yesterday already, but did not understand why at first). The emissive texture and the base_color texture have different layouts which is why the 3rd cube contains both and the 4th is black.

But what do you think? Should we add replication padding to the individual textures or allow for some smaller errors?
We can always change the uv coordinates to be in the middle of a pixel for single pixel textures, but it is not a generally correct solution, because we might still need padding for other textures.
I simply don't know a much better way to process that without writing custom c++ operations. I was actually already quite proud of my numpy solution :D

Some nits:
 - the function is now using a padding of 2 pixels which is probably not necessary.
 - I would still argue that in line 1071 should be `uv_scale = (np.array(img.size) - 1) / (final_size - 1)`

Issue #1947, opened 2023-06-16
Implement boolean operations into the library


emil-peters:
When shipping code I can't force people to have blender or SCAD installed onto their machine.
Would it be possible to have boolean operation build into the library instead of having them using a different engine?

Kind regards

mikedh:
Hey, yeah I would also like to drop or reduce the scope of `trimesh.interfaces` to OpenSCAD/blender if at all possible. Until recently the state of boolean operation backends available has been quite bad (unreliable and slow). I haven't tested but this looks very promising and the OpenSCAD merge is a pretty good sign: https://github.com/elalish/manifold/issues/361

I think the road to getting that as an `trimesh[easy]` dependency is to [set up cibuildwheel](https://github.com/mikedh/embreeX/blob/main/pyproject.toml) and to release a matrix [of Python 3.6-3.latest x Mac/Win/Linux](https://github.com/mikedh/embreeX/releases/tag/2.17.7.post1) and I'm sure they'd love packaging help 😄.

pca006132:
@mikedh yes we would love some python packaging help, we are unable to get the windows build working: https://github.com/elalish/manifold/issues/362

Issue #1974, opened 2023-07-11
Are Self Intersecting Meshes Volumes?


Eric-Vin:
Is a mesh considered a volume if it has self intersections? It doesn't seem like this is considered in the description of `is_volume`, but I've seen it used in other definitions. I've attached an example of a mesh that appears to have intersecting faces but is registered as a volume by Trimesh.
[table_mesh.zip](https://github.com/mikedh/trimesh/files/12017601/table_mesh.zip)

caseymjohnson:
I took a peek at the attached .obj in meshlab, and wasn't able to detect any self intersections

and, anyone feel free to correct me if this is inaccurate, but with respect to the is_volume question:

meshes are inherently a representation of 'inside' versus 'outside', so given consistent winding (no discontinuities wrt. 'inside' vs 'outside' between adjacent faces) and a mesh surface with no holes, the mesh will possess volumetric data

will the magnitude of that volume be particularly meaningful? I believe that depends on the calculation algorithm used, but in general it may not be a particularly useful thing to ask 'how much air is in this impossible balloon'

hope this is useful! 

@mikedh do you think this could be closed?

Eric-Vin:
There are definitely intersecting faces in the `Trimesh` object, but I think the issue is that the `Trimesh` object actually contains multiple meshes, which intersect with each other (i.e. the different support beams). Since `is_volume` is called over the whole Trimesh object, I just wanted to know what the intended semantics were in this case.

As an example, here's some code that returns True on my machine using the above mesh, indicating the bodies/faces intersect:
```
mesh.split()[0].intersection(mesh.split()[8]).is_volume
```

caseymjohnson:
ah you're absolutely correct, I missed the funkiness at the corners

<img width="837" alt="Screenshot 2023-10-26 at 1 49 30 PM" src="https://github.com/mikedh/trimesh/assets/47021682/5c7b422d-bd95-471f-a28b-0341f3a9de8b">

so looking at `is_volume`

```
valid = bool(
            self.is_watertight
            and self.is_winding_consistent
            and np.isfinite(self.center_mass).all()
            and self.volume > 0.0
        )
 ```
 
 does this mean that an object of multiple mesh components is said to be a volume if all of the constituent components meet all four criteria?

Eric-Vin:
[...]

PR #2018, opened 2023-08-30
New interface to Manifold3D.

Eric-Vin:
This PR adds an interface to the [Manifold](https://github.com/elalish/manifold) library, using their Python interface, as discussed in this [issue](https://github.com/mikedh/trimesh/issues/1947).

This is a WIP, as I developed it using a locally built wheel and we'll probably want to wait at least till the wheels are pushed to PyPi (Relevant Issue: https://github.com/elalish/manifold/issues/362) for the CI to run the tests properly. At that point I'll be able to add it as an dependency, which I believe should meet the `[easy]` requirements (thanks to the excellent work in https://github.com/elalish/manifold/pull/537).

@mikedh if there's any missing functionality the interface needs to be complete, please let me know. At the moment it passes all tests on my local machine (with the manifold wheel manually installed), including `test_boolean.py` (in which I added manifold as a potential engine).


mikedh:
Awesome super excited for this!! 

- `trimesh.interfaces` were originally written for moderately janky `subprocess` interfaces to executables. since there's upcoming nice `cibuildwheel` wheel matrices, I'd actually probably build this directly in to `trimesh.boolean`
- if manifold behaves and benchmarks well I'd probably drop the interface boolean functions (especially the openSCAD one, which is now using manifold itself so it seems kinda pointless).

Eric-Vin:
Sounds like a plan! Hopefully this commit addresses everything you listed, and the tests all run on my machine with manifold installed. Please let me know if there's anything else you'd like changed! Once the wheels are released I can revisit the PR to add the dependencies and ensure all the tests are passing on GitHub as well.

Eric-Vin:
@mikedh I think this should be good to review! I was worried we might get issues with meshes like the one in https://github.com/mikedh/trimesh/issues/1974 , but I tested it on my machine and Manifold seems to handle those cases gracefully, so we're probably fine. Please let me know if you have any thoughts!

Eric-Vin:
Hmm, it looks like there are actually conflicts due to black being applied to the main repo. If you'd like I can run black on this branch and hopefully that'll fix things, or you can resolve any conflicts manually when you merge?

Eric-Vin:
Upon further discussion in [this](https://github.com/mikedh/trimesh/issues/1974) issue, it seems like we should establish that a Trimesh object is a valid manifold before calling out to manifold3d. Any thoughts @mikedh ?

PR #2019, opened 2023-09-01
New `torus` function in `creation` module

clemense:
A vectorized function to create the mesh of a torus (saw here https://github.com/mikedh/trimesh/issues/1490#issuecomment-1046328566 that PRs are welcome).
Parameterized by two radii and the number of sections.

Example:
```
import trimesh
trimesh.creation.torus(1.0, 0.2).show()
```

Looks like this:
Screenshot from 2023-08-31 19-48-33

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -326,12 +326,16 @@ def check_triangulation(v, f, true_area):
     assert g.np.isclose(area, true_area)
 
 
-def test_torus(self):
+def test_torus():
     torus = g.trimesh.creation.torus
 
-    m = torus(major_radius=1.0, minor_radius=0.2)
+    major_radius = 1.0
+    minor_radius = 0.2
+    m = torus(major_radius=major_radius, minor_radius=minor_radius)
     
-    extents = g.np.array([1.4, 1.4, 0.4])
+    extents = g.np.array([2 * major_radius + 2 * minor_radius,
+                          2 * major_radius + 2 * minor_radius,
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome looks great thanks for the PR! Could we use `trimesh.creation.revolve` for the face logic or is it way slower?

clemense:
Ah, good point!  It looks like the runtimes are pretty similar:
```
# Original implementation
In [5]: %timeit trimesh.creation.torus(1.0, 0.2, 128, 64)
2.53 ms ± 18.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# Using revolve()
In [6]: %timeit trimesh.creation.torus(1.0, 0.2, 128, 64)
8.86 ms ± 38.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
...and the implementation is much shorter and easier to maintain this way.

I also added the `'shape'`-metadata (analog to what `creation.cylinder(...)` does) although I'm not 100% sure if that's required/creates other problems.

mikedh:
Looks great thanks for the update! Yeah the revolve logic is kind of fiddly, it's nice to have only one place to debug haha. They're both probably fast enough it doesn't make much difference but I wonder why your original implementation is 3x faster, looks like `trimesh.creation.revolve` uses a lot of `np.tile` that it could probably optimize a bit or just use advanced indexing/masks. 

Thanks for the PR!

PR #2023, opened 2023-09-04
Set gmsh options before opening the file

OliverLengwinat:
I wanted to hide all terminal output when converting step files to meshes but I couldn't. 

**How it was:**
General.Terminal is set to 1, then the file is opened, and only after that the gmsh options (args in load_gmsh) are set.
This results in output such as this during opening: 
_Info    : Reading '..\test_files\my_models\my_file.stp'...
Info    :  - Label 'SomeLabel' (3D)
Info    :  - Color (0.74902, 0.74902, 0.74902) (3D & Surfaces)
Info    :  - Color (0.996078, 0.996078, 1) (Surface)
Info    :  - Color (0.996078, 0.996078, 1) (Surface)
Info    :  - Color (0.996078, 0.996078, 1) (Surface)
..._

**Now:**
gmsh options are set before opening the step file. 
This allows to hide the terminal output during opening by passing arg _("General.Terminal", 0)_. 

This works well for me. Please enlighten me if I missed any drawbacks.

Thanks for the great work on trimesh!

### DIFF
--- a/trimesh/interfaces/gmsh.py
+++ b/trimesh/interfaces/gmsh.py
@@ -67,6 +67,10 @@ def load_gmsh(file_name, gmsh_args=None):
         gmsh.initialize()
     gmsh.option.setNumber("General.Terminal", 1)
     gmsh.model.add('Surface_Mesh_Generation')
+    # loop through our numbered args which do things, stuff
+    for arg in args:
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the fix!!

Issue #2029, opened 2023-09-07
WebP extension for GLTF export


xiaoxiae:
I want to use `webp` as the texture format for my 3D models, since they will be displayed on the web and so it's vital for the texture size to be as small as possible. GLTF has the following extension for this: 

https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_texture_webp/README.md

Reading through it and following an extension-related PR, I've attempted to do this:

```python3
scene = trimesh.Scene()
scene.add_geometry(my_triangle_mesh)
scene.metadata.update({'gltf_extensions': {"EXT_texture_webp": {"source": 0}}})

scene.export("test.glb")
```

I understand that this is wrong, because it adds the metadata to the root of the JSON when it should be added to the `textures` node. I'm also not sure how to specify that the texture should be saved as `webp` instead of the default `png`.

Is there a way to do this?

mikedh:
Hey, I think you'd have to hack it into `_append_image`:
https://github.com/mikedh/trimesh/blob/475f921e3497b5c360c56ff269b29c7535147f62/trimesh/exchange/gltf.py#L1769-L1775

Have you tried compressing either the GLTF folder, or the GLB bytes? That has the advantage of applying compression to the vertex/face arrays in addition to just the texture, and can be loaded pretty easily in a web context using standard stuff (there's a few nice webassembly zstandard decompressors on NPM, and probably zip as well):

```
In [1]: import trimesh

# a model with a bunch of textures
In [2]: m = trimesh.load('bigmodel/airplane/models/model_normalized.obj')

In [3]: m
Out[3]: <trimesh.Scene(len(geometry)=29)>

In [4]: e_zip = trimesh.util.compress(m.export(file_type='gltf'))

In [5]: len(e_zip)
Out[5]: 891966

In [6]: e_glb = m.export(file_type='glb')

In [7]: len(e_glb)
Out[7]: 2247540

# compression ratio, i.e. zipped is 39.6% the size of the raw GLB export
In [8]: len(e_zip) / len(e_glb)
Out[8]: 0.3968632371392723

# a non-standard library compression algorithm from FB
In [9]: from zstandard import compress

In [10]: e_zstd = compress(e_glb)

In [11]: len(e_zstd)
Out[11]: 825268

# slightly better than zip
In [12]: len(e_zstd) / len(e_glb)
Out[12]: 0.3671872358222768
```

xiaoxiae:
[...]

PR #2031, opened 2023-09-09
Support for glTF WebP extension

xiaoxiae:
Implements #2029.

I think that `max_texture_size` mentioned in the issue is not necessary, since you can just edit the PIL image in the material itself. 

There is currently no test since I'm not sure how it should behave. Would it be sufficient to export a mesh using the extension, load it and check that the texture has an image with the WebP extension?


mikedh:
Thanks for the PR! Yeah if you just add a load/export cycle it should be checked against the GLTF validator in `test_gltf.py` (or you can call the function in there explicitly).

xiaoxiae:
Running `test_gltf.py` passes. The newly added test loads a mesh with texture and tests save/load validity.

mikedh:
Awesome, thanks for the PR and test!

PR #2035, opened 2023-09-14
fixed gltf export when using paths instead of only file name

dabeschte:
When you store a GLTF file using a path instead of only a file name, this line effectively duplicated the path.

E.g., instead of writing all files into `out` when using `scene.export("out/scene.gltf")`, it saved all files in `out` except for the `gltf` file which is stored in `out/out/mesh.gltf` - raising an exception doing so.

### DIFF
--- a/trimesh/exchange/export.py
+++ b/trimesh/exchange/export.py
@@ -90,7 +90,7 @@ def export_mesh(mesh,
     if isinstance(export, dict):
         # if we have a filename rename the default GLTF
         if file_name is not None and 'model.gltf' in export:
-            export[file_name] = export.pop('model.gltf')
+            export[os.path.basename(file_name)] = export.pop('model.gltf')
 
         # write the files if a resolver has been passed
         if resolver is not None:


mikedh:
Thanks for the fix!! What's the combination of export args that hits this? I was trying to test with variants on this but couldn't quite get it:

```
    from tempfile import TemporaryDirectory
    with TemporaryDirectory() as d:
	trimesh.creation.box().export(d, file_type='gltf')
        r = trimesh.load(os.path.join(d, 'model.gltf'))
```

dabeschte:
I can reproduce the error with this script:

```
import trimesh
m = trimesh.load("example.glb", force="mesh")
m.export("fused/example.gltf")
```

error message:
`[Errno 2] No such file or directory: C:\\Users\\dabeschte\\Desktop\\test_meshes\\fused\\fused/pbr_cubes_emissive_fused.gltf`

the same happens when I use double backslashes or `os.path.join`

PR #2036, opened 2023-09-14
Spec gloss tests

dabeschte:
I think I finally managed to get the entire function working as expected. 
The basic conversion was already correct, but I made a lot of mistakes in handling the textures.
According to GLTF spec, textures are stored in sRGB space, but the factors must be applied in linear space.
Also, I faced some issues with 2 channel textures. Not sure where the problem is coming from, but it could easily be solved by using 3 RGB textures instead.

pyramid.zip using specular glossiness
image

pyramid.zip using metallic roughness
image

[scifi girl](https://sketchfab.com/3d-models/scifi-girl-v01-96340701c2ed4d37851c7d9109eee9c0) using specular glossiness
image

scifi girl using metallic roughness. not a 100% fit, but metallic roughness is the weaker model.
image

I also added another testcase with an extreme example where all color is coming from specular, but none from diffuse. the same scene will be reused for my PBR material fusion tests.


mikedh:
Awesome thanks for the PR, those examples look great!! Could you add a docstring on the helper functions like `srgb2lin(s)` and maybe a comment or two explaining where the constants (like `mask = lin > 0.0031308`) come from?

dabeschte:
Sure. I linked the [source](https://entropymine.com/imageworsener/srgbformula/) that derives these constants.

PR #2043, opened 2023-10-01
Modified to skip processing when the vertex attribute of primitive is None

iory:
This PR avoids errors when loading collada (dae) files that do not have vertices by skipping the process if the vertex is None.


mikedh:
Thanks for the fix!

PR #2055, opened 2023-10-22
[BUGFIX] Fixing UV coordinate issue in obj loading when setting maintain_order=True

### LINKED ISSUE #2002: some unexpected color points in sample_surface
Thanks for the great project! I have noticed that trimesh.sample_surface now can sample color, but when I use this function, some unexpected color points appear.  I use this mesh [mesh.zip](https://github.com/mikedh/trimesh/files/12342377/mesh.zip) as example. The code is 
```
mesh = trimesh.load('mesh-f00001.obj', process=False, maintain_order=True)
sample_points, _, colors = trimesh.sample.sample_surface(mesh, 50120, sample_color=True)
np.savetxt('tmp.txt', np.concatenate([sample_points, colors[:,:3]],axis=1))
```

The result of the sampled points [tmp.txt](https://github.com/mikedh/trimesh/files/12342454/tmp.txt) looks like this
snapshot00

There are some white points on the hair and some black points on the elbow, while when I open the mesh with meshlab, the color is normal.
 
snapshot01

Is this a bug or something wrong with this mesh?

jinyx728:
- Fixes #2002 

In the mask computation in `unmerge_faces`, the masks are set to default as 0 and the median value of 3 options is chosen afterwards. The problem is that if only one option is set (like 100), then the median is computing [0, 0, 100] and returns vertex index 0, but the mask should return 100. This results in multiple 0 entries in the mask which corrupts the texture.

My fix is changing the default value to -1 (which doesn't conflict with true vertex index 0) and remove all -1 before median computation. This way computing [-1, -1, 100] returns 100. The way of removing -1 is changing them to nans and call np.nanmedian.

Visualizing the result using the mesh provided in #2002, this is before fix:

Screenshot 2023-10-21 at 20 25 57

and this is after fix:

Screenshot 2023-10-21 at 20 25 10

Testing code is simply:
```
import trimesh

mesh = trimesh.load('mesh/mesh-f00001.obj', process=False, maintain_order=True)
mesh.show()
```


mikedh:
Awesome thanks for the fix!! Would it be possible to add a unit test for this? No worries if not this sort of thing can be hard to test.

jinyx728:
Thanks! yeah maybe I can try render a textured obj with both `maintain_order=True` and `False` and compare their pixel values? Don't know if it is a robust thing to do in the unittest.

PR #2064, opened 2023-10-30
[collada] Apply mesh scale and resize vertices and Set default value for diffuse and shininess

iory:
In this pull request, I've made the following key changes:
## Mesh Scaling and Vertices Resizing: 

I've introduced functionality to apply the mesh scale and dynamically resize the vertices based on the specified scale values. 

## Default Material Adjustments: 

Set default values for diffuse and shininess properties.

### DIFF
--- a/trimesh/exchange/dae.py
+++ b/trimesh/exchange/dae.py
@@ -365,6 +365,8 @@ def _unparse_material(material):
     # TODO EXPORT TEXTURES
     if isinstance(material, visual.material.PBRMaterial):
         diffuse = material.baseColorFactor
+        if diffuse is None:
+            diffuse = np.array([255.0, 255.0, 255.0, 255.0])
         diffuse = diffuse / 255.0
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR!! Oh nice I didn't realize collada had units. I think just multiplying the vertices isn't enough though, won't that leave the transforms at the original scale? `scene.scaled` has [some complicated logic](https://github.com/mikedh/trimesh/blob/5a301ed9e86181e9e98f6f3164f3ffb3fa716133/trimesh/scene/scene.py#L1074-L1200) for that which may or may not be necessary.

I wonder if we could just extract a units string (`m`, `inch` etc) from the `unitmeters` value and then the user could apply `scene.scaled` via `convert_units` if they desired.

iory:
I've added 'units' to the metadata. Users can now change the scale at will. How does this look?

PR #2067, opened 2023-11-06
Release: Manifold Booleans

mikedh:
- release #2018 which switches the default boolean engine to the python bindings for [manifold(https://github.com/elalish/manifold], which seems like a massive, massive improvement over previous options. OpenSCAD also switched to it as a boolean backend so the interfaces to OpenSCAD no longer makes sense IMO. 
  - in our simple-ish tests gets a 55x speedup over calling blender with subprocess 😎: `test_boolean.py:67    {'blender': 3.3091907501220703, 'manifold': 0.0628511905670166}`
- release #2064 
  - allow and test for unit strings to have a simple scale factor. I.e `0.01 * meters` is now valid. This is relatively strict (i.e `meters * 0.01` is not valid, `2meters/3` is not valid, etc) to avoid any stuffing of math other than a fully evaluated float into the units field.

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -71,6 +71,7 @@ Trimesh has a lot of soft-required upstream packages, and we try to make sure th
 |`pytest-cov`| A plugin to calculate test coverage. | | `test`|
 |`pyinstrument`| A sampling based profiler for performance tweaking. | | `test`|
 |`vhacdx`| A binding for VHACD which provides convex decompositions | | `recommend`|
+|`manifold3d`| A binding for the Manifold mesh boolean engine | | `recommend`|
 
 ## Adding A Dependency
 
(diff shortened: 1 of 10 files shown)


elalish:
Wow, I was about to file an issue to see if Manifold might be useful to your project, only to find this! Thanks @mikedh and @Eric-Vin - I'm a huge fan of trimesh. I even built a [Colab](https://colab.research.google.com/drive/1VxrFYHPSHZgUbl9TeWzCeovlpXrPQ5J5?usp=sharing) to show how to use Manifold in Python using your library for visualization and file I/O, though now it probably makes sense to use the functions you built into trimesh. Would you be interested in commenting on https://github.com/elalish/manifold/discussions/340? I use it to keep track of projects we're a dependency of, and to give potential users some confidence. 

Also, have you seen our new official glTF extension, [EXT_mesh_manifold](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_mesh_manifold)? I designed it specifically for lossless transmission of manifold meshes with properties, so I'm guessing it would be useful for trimesh. I made a JS implementation, but it would be amazing to also have a Python implementation so that we could reliably interchange between multiple projects.

PR #2068, opened 2023-11-08
Add missing engine keyword in boolean test_multiple

musicinmybrain:
I noticed this because `test_multiple` was failing when `manifest3d` was not available.


musicinmybrain:
That’s what happens when I open a PR without actually testing it first. Force-pushing.

mikedh:
Thanks for the fix!

PR #2073, opened 2023-11-16
fix error when specular-glossiness only has one channel

chenming-wu:
Otherwise would rase error "ValueError: can only specify one unknown dimension"


mikedh:
Awesome thanks for the fix!

Issue #2093, opened 2023-12-13
Skip loading texture


xiaoxiae:
Loading a texture for certain models (e.g. generated by photogrammetry) can take quite a while when the texture is large (~hundreds of megabytes). It might be a good idea to add an argument for the loaders to not load the texture since we might only care about the mesh itself.

Would a PR that implements this for the mesh loaders be welcome?

xiaoxiae:
It seems that certain formats have options that do this (e.g. obj's skip_materials) but certain don't (e.g. PLY, which is why I created this issue). I guess I could just implement it for PLY?

mikedh:
Yeah PR's super welcome on this thanks for looking in to it!

xiaoxiae:
Closing since it was merged.

PR #2094, opened 2023-12-14
PLY and GLB parameters for skipping texture loading

xiaoxiae:
Relates to #2093.

Since `obj` already implemented this via `skip_materials`, I implemented the same parameter for `glb` and similar parameter (`skip_texture`) for `ply`.


xiaoxiae:
The tests are implemented in maybe not a great way since I didn't find a general way to test if a mesh has no visuals (since they always default to something).

mikedh:
Looks great thanks for the PR!! I might rename `skip_texture` to `skip_materials` just to match the other two functions even if it's less accurate.

Issue #2099, opened 2023-12-20
New type hints prevent passing PathLib Path to load(...)


friggog:
The `Loadable` type is `Union[str, IO]` but this gives a type error if passing `Path` from pathlib even though this is explicitly supported in the code.

Ideally the `Loadable` type would be something like `Union[str, Path, IO]` to resolve this

mikedh:
Thanks for the fix!

PR #2100, opened 2023-12-20
Support pathlib Paths in Loadable

friggog:
#2099

### DIFF
--- a/trimesh/typed.py
+++ b/trimesh/typed.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import IO, Dict, List, Optional, Sequence, Tuple, Union
 
 # our default integer and floating point types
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the fix! Just to double check it shouldn't have raised a runtime `TypeError` in this case, it's just that pylance/etc gets mad at it?

```
# pre this fix
In [4]: import trimesh

In [5]: from pathlib import Path

In [6]: trimesh.load(Path('models/featuretype.STL'))
Out[6]: <trimesh.Trimesh(vertices.shape=(1722, 3), faces.shape=(3476, 3), name=`featuretype.STL`)>


```

friggog:
Yeah exactly, not runtime error, but pylance and linters like Mypy would complain

PR #2108, opened 2024-01-06
Update obj.py to support quad faces with texture coordinates

zhou-wb:
Support quad faces for obj file with texture coordinates and/or face normals

the support for 8 and 12 columns for quads can be easily done by make tiny changes, I have done some tests and it works properly.

### DIFF
--- a/trimesh/exchange/obj.py
+++ b/trimesh/exchange/obj.py
@@ -404,7 +404,7 @@ def _parse_faces_vectorized(array, columns, sample_line):
     # TODO: probably need to support 8 and 12 columns for quads
     # or do something more general
     faces_tex, faces_norm = None, None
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! Looks like tests are failing, I think you want to change the check from `columns == 6 or 8` to `columns in (6, 8)`:

```
In [1]: columns = 10

In [2]: columns == 6 or 8
Out[2]: 8

In [3]: columns in (6, 8)
Out[3]: False

```

zhou-wb:
Thanks for the reply. Oh, I actually want to check if the columns == 6 (for triangle facet with one attribute, vn or vt) or columns == 8 (for quad facet with both attributes vn and vt)

PR #2110, opened 2024-01-08
fix: copy visual data when copying points

HiroIshida:
As the failure of unit test in 59b9403 shows, PointCloud's copy operation does not copy the visual data in the current implementation. This PR fixes this problem.


mikedh:

PR #2117, opened 2024-01-11
Fix error in transformations.angle_between_vectors

### LINKED ISSUE #2116: angle_between_vectors fails when the vectors are identical
The function `trimesh.transformations.angle_between_vectors` fails when the input vector are identical.

Trimesh version: 4.0.8
Python version: 3.8.10

Example
```
>>> import trimesh
>>> import numpy as np

>>> trimesh.transformations.angle_between_vectors(np.ones(3), np.ones(3))
lib/python3.8/site-packages/trimesh/transformations.py:1936: RuntimeWarning: invalid value encountered in arccos
  return np.arccos(dot if directed else np.fabs(dot))
nan
```

lucagrementieri:
Close #2116.

If have found no contribution guidelines, so I open this plain PR without really knowing how to contribute to Trimesh. Thank you for the amazing work!

### DIFF
--- a/trimesh/transformations.py
+++ b/trimesh/transformations.py
@@ -1933,6 +1933,7 @@ def angle_between_vectors(v0, v1, directed=True, axis=0):
     v1 = np.array(v1, dtype=np.float64, copy=False)
     dot = np.sum(v0 * v1, axis=axis)
     dot /= vector_norm(v0, axis=axis) * vector_norm(v1, axis=axis)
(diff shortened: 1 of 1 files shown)


mikedh:
Looks great, thanks for the fix! Yeah clipping before the `arccos` seems like a good idea in general, looks like this case was off by 2e-16 from floating point "stuff"

```
In [1]: dot
Out[1]: 1.0000000000000002
```

PR #2123, opened 2024-01-20
Fix for always converting ASCII STL names to lower case for issue #2114 

njpatter:
Probably not the most elegant or efficient, but it works!


mikedh:
Thanks for the PR!! Would you mind adding a unit test as well? 

I'll have to check the performance of looping through the file twice before releasing, my other thought is we might be able to keep track of position such that we could pull the name from the non-lower-case copy of the file.

PR #2124, opened 2024-01-21
Fix missing colors after concatination with empty mesh

mishasweetpie:
This PR resolves https://github.com/mikedh/trimesh/issues/1991 by changing the way coloring kind is selected in concatenation method.


mikedh:
Thanks for the fix and test!!

PR #2141, opened 2024-02-02
Uses unitize null safe function to unitize normals in gltf export

clemgaut:
Proposed fix for #2140 and an associated unit test


mikedh:
Thanks for the fix!

PR #2145, opened 2024-02-08
Adding solver_options and self_intersection to bool.intersection

Erkhembayaar:
image
Adding Solver Option and Use_Self making bool.intersection work in most cases

### DIFF
--- a/trimesh/boolean.py
+++ b/trimesh/boolean.py
@@ -57,7 +57,7 @@ def union(meshes, engine=None, **kwargs):
     return result
 
 
-def intersection(meshes, engine=None, solver_options="FAST", use_self=False, **kwargs):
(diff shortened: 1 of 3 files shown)


mikedh:
Thanks for the PR! Since `solver` has only two options, maybe we should make both arguments booleans? I.e.: `use_self: bool = True, use_exact: bool = True`

As a side note, have you tried using the `pip install manifold3d` engine?

Erkhembayaar:
Here is the new PR

Erkhembayaar:
New commit with description - condition check to cond is True and removed some trailing whitespaces

Erkhembayaar:
@mikedh Could you please review the pull request and inform me if there are any elements that need further attention or improvement?

PR #2149, opened 2024-02-13
Fix reshape in `get_specular_glossiness`.

eliphatfs:
The current code is wrong when the array has a singleton dimension as channel dimension already.

### DIFF
--- a/trimesh/visual/gloss.py
+++ b/trimesh/visual/gloss.py
@@ -180,9 +180,9 @@ def specular_to_pbr(
                 or specularGlossinessTexture.shape[-1]
             ) == 1:
                 # use the one channel as a multiplier for specular and glossiness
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! I'll have to play with it but this won't necessarily handle the case `or specularGlossinessTexture.shape[-1] == 1`? Do you have a small example model you could share which hits this case?

eliphatfs:
Revisiting the code I think the case shall not be hit for the current latest PIL library as single channel images lead to 2D arrays.
I was actually copying the function to my code and passing an array from opencv (which has the singleton dimension at the end).

mikedh:
Ok, I separated out the two cases. The specular glossiness conversion needs more test coverage haha, PR's very welcome 😄

PR #2150, opened 2024-02-13
fix binary ply export and extend unittest

moberweger:
- Fix export of binary ply format with different dtypes (float32, float64)
- Adjust unittest to cover this issue
- Resolves issue https://github.com/mikedh/trimesh/issues/2147


mikedh:
Awesome thanks for the fix and test! Were we converting the dtype when we shouldn't have been with `_numpy_type_to_ply_type`?

moberweger:
Thanks @mikedh for considering the PR.
Yes, it seems that there was one conversion too much. As evident two lines above the modification, the additional conversion is not there for the 1D case.

PR #2151, opened 2024-02-13
Fix caching with FIPS mode

MattTheCuber:
With FIPS mode the following error is thrown.
`ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
`


mikedh:
Thanks for the fix! I think I'll change the fallback to check Python version rather than try-except which should fix tests. It looks like that keyword argument was added in Python 3.9.

PR #2162, opened 2024-02-25
Fix typing for lights in Scene.__init__

maitbayev:
`lights` should be `Optional[List[lighting.Light]]` instead of `Optional[lighting.Light]`


maitbayev:
cc @mikedh

mikedh:
Thanks for the fix!! I added typeguard to check this in the future, which caught a bunch of other instances. I didn't do the whole module, but if anyone else wants to contribute typing fixes if you run tests with typeguard it catches a lot more of them:

```
pytest --typeguard-packages=trimesh tests
```

PR #2172, opened 2024-03-04
Update README.md

AndreNasci:
Fix broken links in the Quick Start section. 

During my first contact with the library, when trying to access the links in the quick start section of the README.md, I noticed that some of them were broken. More specifically those that refer to the ipython notebook version and cross section example.

This PR fix these links.


mikedh:
Thanks for the fix!!

PR #2176, opened 2024-03-07
Update docstring for transform structure used in creation methods

nmarticorena:
Hi, Thanks for such nice and usefull library!

A small PR to add the dosctring for the creation methods to specify the transform.


mikedh:
Awesome thanks for the PR!

PR #2196, opened 2024-03-28
Fix broken solid name in ASCII STL mesh export

lcavegliacurtil:
Exporting a mesh in ASCII STL format results in the placeholder `{name}` to be printed as solid name instead of the `name` metadata, due to a typo. This PR fixes the typo and introduces a simple unit test.


mikedh:
Thanks for the fix and test!! I'll also enable [RUF027](https://docs.astral.sh/ruff/rules/missing-f-string-syntax/#missing-f-string-syntax-ruf027) which appears to catch this typo for the future.

PR #2202, opened 2024-04-03
Option for closed path polygon sweeping

Raserino:
Hello,

thanks for your ongoing awesome work on Trimesh!

For a project I needed closed sweeping of a polygon, so the start and the end of the path are connected. I wanted to use `creation.sweep_polygon()` with a circular path but it always adds caps to the mesh at the end of the path.

I thought sweeping a closed path is actually quite a common use case, so I added an `closed` option to `creation.sweep_polygon()` to close the mesh at the path ends instead of addings caps. 

The results look good as long as the path ends are roughly facing each other:
  
image
image

Additional note:

One issue is a potential narrowing and twisting when adding angles while sweeping something rotational symmetric like a hexagon because the first and last polygon are rotated against each other without the rotation being visible due to the rotational symmetry (see the example image below).

But I couldn't come up with a trivial way around this, especially without calculating the symmetry angle of the polygon which I didn’t find an existing method for. Is there already a method for this in trimesh or shapely? Otherwise this is maybe something I can look into on another day...

image

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -147,6 +147,7 @@ class CreationTest(g.unittest.TestCase):
             vec = g.np.dot(rotmat, vec)
         poly = g.Polygon(perim)
 
+        # --- test open sweep
         # Create 3D path
         angles = g.np.linspace(0, 8 * g.np.pi, 1000)
         x = angles / 10.0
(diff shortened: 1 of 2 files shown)


mikedh:
Looks great, thanks for the PR! The only thing that comes to mind is it would probably be nice to detect `closed` from the input path rather than take it as a boolean argument? I.e.:

```
closed  = np.linalg.norm(path[0] - path[-1]) < tol.merge

```

PR #2204, opened 2024-04-04
remove broken gmsh.isInitialized() check

l-koehler:
fixes issue #2157, it works just fine without that check (at least not worse than before :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = ["setuptools >= 61.0", "wheel"]
 [project]
 name = "trimesh"
 requires-python = ">=3.7"
-version = "4.2.4"
+version = "4.2.5"
 authors = [{name = "Michael Dawson-Haggerty", email = "mikedh@kerfed.com"}]
 license = {file = "LICENSE.md"}
 description = "Import, export, process, analyze and view triangular meshes."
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR!! Would you be able to add a test that runs in the docker images? Last time I looked at this for some reason I ran a `pip install gmsh` and wasn't able to hit this code block for whatever reason.

l-koehler:
I am not sure why the tests don't catch this either, my (failing) code doesn't do anything significantly different. Both scripts end up calling `g.trimesh.interfaces.gmsh.load_gmsh("path/to/file.STEP")`.

mikedh:
I wonder if this is an upstream packaging thing, have you done a `pip install --upgrade gmsh` lately? Happy to merge this but I might add some more conditionals.

```
In [1]: import gmsh

In [2]: gmsh.initialize
Out[2]: <function gmsh.initialize(argv=[], readConfigFiles=True, run=False, interruptible=True)>

In [3]: gmsh.is_initialized
Out[3]: <function gmsh.isInitialized()>

In [4]: gmsh.is_initialized()
Out[4]: 0

In [5]: gmsh.isInitialized
Out[5]: <function gmsh.isInitialized()>

In [6]: gmsh.isInitialized()
Out[6]: 0
```

By the way, have you checkout out using [`cascadio`](https://github.com/trimesh/cascadio) for STEP files? I recently packaged it and it seems anecdotally a lot more robust than the GMSH bindings. `pip install trimesh cascadio` should get you pretty solid STEP support, at least within the reliability of OpenCASCADE 😄.

PR #2212, opened 2024-04-19
Fix pbr export no uv

nelianos:
Thanks for the great work!

In this PR, we introduce a fix for properly exporting pbr materials to gltf when there's no UV mapping.

* GLTF allows PBR materials without texture image, with only the `Factors` be defined.
* Within trimesh, `TextureVisual` may not have a uv map associated, and annotated as optional.
* Currently, since version 4.0.0, exporting to gltf would avoid adding a reference from the mesh to the material if `uv` does not exist. Therefore the loaded `.gltf/glb` would load with default colors.

This is the same issue raised in https://github.com/mikedh/trimesh/issues/2111

### DIFF
--- a/tests/test_texture.py
+++ b/tests/test_texture.py
@@ -206,6 +206,32 @@ class TextureTest(g.unittest.TestCase):
         c = m.copy()
         assert c.visual.uv is None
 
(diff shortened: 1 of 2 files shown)


mikedh:

PR #2213, opened 2024-04-22
Pass through registration kwargs to icp, e.g. to dissallow reflection

obooij:
In the documentation it says that kwargs given to `.register` are passed through to `icp` and further down to `procrustes`. This can, for example, be used to disallow reflection.
However, these kwargs are currently not passed through. This PR corrects that.

Note, that the `scale` flag could be removed (as it is part of kwargs), but I left it in to not change the order of the arguments for backward compatibility.


mikedh:
Thanks for the fix!

PR #2217, opened 2024-05-03
Fix load cam gltf

nelianos:
* Loading a `.gltf/.glb` with a camera, will currently not load the camera object to the scene, and instead contain an empty "dangling" node that was originally the camera node.
* This is shown with the unit test; previously the loaded camera would have default values different than the original
* In this PR, we suggest loading the camera into the `Scene` object if it exists in the gltf, populating the gltf-compatible fields (`fov`, `name`, `z_near`) and the `camera_transform`


mikedh:
Thanks for the fix looks great!

PR #2231, opened 2024-05-22
Fixes #2195

### LINKED ISSUE #2195: Likely a bug in `trimesh.visual.gloss.specular_to_pbr`
According to the GLTF 2.0 specification, the metallicRoughnessTexture is stored as "green channel contains roughness values and its blue channel contains metalness values" 
image
(https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material).

However, in the conversion of `specular_to_pbr` (shown as below), the current implementation puts the metallic value to the R channel, which seems to be a bug.
https://github.com/mikedh/trimesh/blob/563690aa2f36135e4442a3a8f3baf0f0eafc4fa9/trimesh/visual/gloss.py#L363

FrozenBurning:
Update trimesh.visual.gloss.specular_to_pbr to align with the material definition in gltf 2.0, as specified in https://github.com/mikedh/trimesh/issues/2195

### DIFF
--- a/tests/test_gltf.py
+++ b/tests/test_gltf.py
@@ -307,7 +307,8 @@ class GLTFTest(g.unittest.TestCase):
         )
         assert metallic_roughness.shape[0] == 84 and metallic_roughness.shape[1] == 71
 
-        metallic = metallic_roughness[:, :, 0]
+        # https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material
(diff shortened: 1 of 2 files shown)


mikedh:
Sorry for the delay on this! Looks good, I think the tests just need to be updated to match the change?

FrozenBurning:
@mikedh Hey! I've updated the tests accordingly with all run tests passed.

mikedh:
Thanks for the update!

PR #2233, opened 2024-05-30
Fix passing read only vertex color buffer when loading GLTF

RealDanTheMan:
Without the fix GLTF loading will fail assigning vertex color properly down the line when colors are converted to RGBA which attempts to modify the buffer contents directly, as a result raising exception.

https://github.com/mikedh/trimesh/blob/7853a5ebae3b35275f4d8d65cbc48bcd3a3c8a40/trimesh/visual/color.py#L583

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -1560,11 +1560,11 @@ def _read_buffers(
                             if len(colors) == len(kwargs["vertices"]):
                                 if visuals is None:
                                     # just pass to mesh as vertex color
-                                    kwargs["vertex_colors"] = colors
+                                    kwargs["vertex_colors"] = colors.copy()
                                 else:
                                     # we ALSO have texture so save as vertex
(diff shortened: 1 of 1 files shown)


RealDanTheMan:
Proposed solution only creates a copy when vertex colors are valid, when their length == vertex count

mikedh:
Looks good thanks for the fix! Is there a sample mesh we could add to unit tests for this (or something in `tests/corpus` maybe?)

RealDanTheMan:
Hey @mikedh, I have added the test case along with a simple subdivided cube model for it, the cube is filled with magenta vertex color.

The test case has to actually validate the values stored in the vertex color to prove success, because even if the vertex color access fails it falls back on some default or pre-calculated values, in my test cases without the fix they were reading `[102 102 102 255]`

mikedh:
Awesome test and fix looks good, thanks for the PR!

PR #2254, opened 2024-07-21
Update transformations.py

apalkk:
In the function rotation_matrix the input array direction is being normalised by the the function unit_vector anyways, so there is no need to explicitly require direction to only consist of unit_vectors in the description of rotation_matrix.


mikedh:
Looks good to me, thanks for the PR!

PR #2273, opened 2024-08-26
Flexible section amounts for bounding_cylinder

djean032:
Adds ability to specify number of sections to better approximate cylinders.

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -88,6 +88,7 @@ class Trimesh(Geometry3D):
         use_embree: bool = True,
         initial_cache: Optional[Dict[str, ndarray]] = None,
         visual: Optional[Union[ColorVisuals, TextureVisuals]] = None,
+        sections: int = 32,
         **kwargs,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! Yeah it's a good point we might want better control of all of the parameters used for the magic `@property` values like `mesh.ray` (controlled by `use_embree: bool` right now), `mesh.bounding_cylinder`, etc. I'm not totally sure where the right place to put those would be, but I think it should probably also include values for `mesh.bounding_sphere`, `mesh.bounding_cylinder`, `mesh.ray`. Maybe the right place to put these would be as a `dataclass` although it is a bit wordy:

```
@dataclass
class PropertySettings:
    # keyword arguments passed to `trimesh.primitives.Cylinder`
    bounding_cylinder: dict

    use_embree: bool = True
```

Although I guess for cylinders you could also call the methods it's currently using:
```
from trimesh import bounds, primitives

kwargs = bounds.minimum_cylinder(a_mesh)
mincyl = primitives.Cylinder(mutable=False, sections=100, **kwargs)
```

djean032:
Thanks for considering the PR. I think that's an interesting idea. I closed this for now, because my solution was hacky, but I'll keep working on it, because it's pretty useful in my current work.

PR #2275, opened 2024-08-27
Speed up GLTF loading.

eliphatfs:
In my profile, more than 80% time is spent in the loop of strided accessor construction when loading GLTF/GLB files.

I thus applied `np.lib.stride_tricks.as_strided` to speed up the process.
I tested it on multiple glb files and the results are the same as the original logic.

This implementation also works for `stride = 0`, repeating the first element for `count` times. But glTF 2.0 specification 5.11.4 says it should be positive, so I added an assertion.

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -1436,20 +1436,20 @@ def _read_buffers(
                 if "byteStride" in buffer_view:
                     # how many bytes for each chunk
                     stride = buffer_view["byteStride"]
-                    # the total block we're looking at
-                    length = count * stride
                     # we want to get the bytes for every row
                     per_row = per_count * dtype.itemsize
-                    # we have to offset the (already offset) buffer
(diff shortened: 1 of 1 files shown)


mikedh:
Looks good!! The docs for [`as_strided`](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.as_strided.html) do encourage "extreme caution" but it does look like you're checking all the things they warn about (copy the result, make sure data is long enough, etc). And you're right it's a nice speedup on `corpus` (although I'd also like to check that it loads the same number of meshes :)


```                                                       
This PR:
- 3.289 load_gltf  trimesh/exchange/gltf.py:268
- 3.054 load_glb  trimesh/exchange/gltf.py:343

Current `main`:
- 5.022 load_gltf  trimesh/exchange/gltf.py:268
- 3.616 load_glb  trimesh/exchange/gltf.py:343
```

PR #2276, opened 2024-08-27
Release: Stride Speedup + Hints

mikedh:
- add some type hints to `trimesh.inertia`
- in `trimesh.base` replace `caching.cache_decorator` with `cache_decorator`. 
- add `mesh.copy(..., include_visual: bool = True)` option to copy meshes skipping the visual data.
- release #2275 
- release #2277
- remove `mesh.as_open3d` immediately and without deprecation as unfortunately it appears to be untested and fully broken in 2024. 
  - This modifies `mesh.simplify_quadric_decimation` to use [pyvista/fast-simplification](https://github.com/pyvista/fast-simplification), which as of testing only requires `numpy` and has `cibuildwheel` set up. I added it to the `recommend` pip extra which most people won't get, and if it works out we might move it up to `easy`.
  - Applies March 2024 deprecation by removing `Trimesh.simplify_quadratic_decimation`
- add FCL back to docker images for tests with a workaround for currently broken PyPi wheels.
- in docker we should no longer have to `force-reinstall` the `[all]` extra since release of https://github.com/imageio/imageio/pull/1098
- updates github actions versions for checkout/cache/setup-python to silence warnings.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,9 @@ jobs:
     name: Check Code Formatting
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: "3.12"
     - name: Install Formatting
@@ -38,9 +38,9 @@ jobs:
           - os: macos-13
             python-version: 3.9
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Test a minimal install
@@ -57,9 +57,9 @@ jobs:
     needs: [tests, containers]
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: '3.11'
     - name: Install publishing dependencies
@@ -81,7 +81,7 @@ jobs:
     - name: Log In to Docker Hub
       run: echo ${{ secrets.DH_PASS }} | docker login --username mikedh --password-stdin
     - name: Checkout trimesh
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
     - name: Build Images And Docs
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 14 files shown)


Jakub-Ramlab:
Dear Michael,

First of all, thank a lot for putting together and maintaining this great library.

In our software stack, we were relying in a number of places on `mesh.as_open3d` to interface between the two libraries. It always seemed to work without problems. So I cannot agree that it's fully broken. Partially? Maybe, but we experienced nothing that would indicate that.

Removing this function suddenly, without deprecation, and with no further explanation why is it "fully broken" caught us by surprise. It gives us extra work to, essentially, implement the conversion ourselves and replace it in every single instance in our code base. While not a critical issue, it creates overhead and erodes our trust in Trimesh being a reliable part of our stack. I am sure we are not the ones in this situation.

I would much appreciate if, in the future, such changes were phased in more slowly and clearly communicated to end users. Thank you very much in advance.

Jakub Sikorski,
RAMLAB BV

Scheepsbouwweg 8 - F6, 
3089 JW Rotterdam 
The Netherlands

mikedh:
Sorry for the hassle. This one is a bit of a special case as `mesh.as_open3d` was on the base object and I though the downsides of new users trying to use it over the next year were worse than the downsides of yanking it. Usually there's at least a year of deprecation warnings, and for what it's worth I'll try to avoid this in the future if at all possible. 

I'd recommend pinning `trimesh==4.4.7` if you need that method.

"Fully broken" probably depends on the environment, but when I checked it (my original plan was to add a unit test), a user on Python 3.11 who tried to use it would hit 400mb+ `open3d` wheels, followed by "libgomp not found," "libx11 not found", and when it imports `mesh.as_open3d` immediately segfaults:

```
mikedh@orion:trimesh$ docker run -it python:3.11-slim-bookworm /bin/bash

root@12e029342d42:/# pip install open3d trimesh==4.4.7

root@12e029342d42:/# python
Python 3.11.10 (main, Sep 27 2024, 06:09:21) [GCC 12.2.0] on linux
>>> import trimesh
>>> trimesh.__version__
'4.4.7'

>>> import open3d
  File "/usr/local/lib/python3.11/site-packages/open3d/__init__.py", line 61, in <module>
    load_cdll(str(next((Path(__file__).parent / "cpu").glob("pybind*"))))
  File "/usr/local/lib/python3.11/site-packages/open3d/__init__.py", line 38, in load_cdll
    return CDLL(str(path))
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/ctypes/__init__.py", line 376, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libgomp.so.1: cannot open shared object file: No such file or directory


root@12e029342d42:/# apt update && apt install -qq libgomp1
After this operation, 319 kB of additional disk space will be used.


root@12e029342d42:/# python
Python 3.11.10 (main, Sep 27 2024, 06:09:21) [GCC 12.2.0] on linux
>>> import open3d
  File "/usr/local/lib/python3.11/site-packages/open3d/__init__.py", line 61, in <module>
    load_cdll(str(next((Path(__file__).parent / "cpu").glob("pybind*"))))
  File "/usr/local/lib/python3.11/site-packages/open3d/__init__.py", line 38, in load_cdll
    return CDLL(str(path))
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/ctypes/__init__.py", line 376, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libX11.so.6: cannot open shared object file: No such file or directory


root@12e029342d42:/# apt install -qq xorg
After this operation, 389 MB of additional disk space will be used.

root@12e029342d42:/# python
Python 3.11.10 (main, Sep 27 2024, 06:09:21) [GCC 12.2.0] on linux
>>> import open3d
>>> import trimesh

>>> trimesh.__version__
'4.4.7'

>>> open3d.__version__
'0.18.0'

>>> m = trimesh.creation.box()
>>> m.as_open3d
Segmentation fault (core dumped)

```

PR #2277, opened 2024-08-30
fix documentation for repair.stitch

dcetin:
This PR fixes the docstring for trimesh.repair.stitch function, which expects a mesh as input and not an array of vertices.


mikedh:
Thanks for the fix!

PR #2280, opened 2024-09-02
Update gmsh.py

sadiwali:
Added optional parameter to load_gmsh to allow running outside the main thread.

### DIFF
--- a/trimesh/interfaces/gmsh.py
+++ b/trimesh/interfaces/gmsh.py
@@ -23,7 +23,7 @@ _warning = " ".join(
 )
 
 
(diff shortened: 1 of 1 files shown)


mikedh:
Looks good thanks for the option! If you're using `trimesh.interfaces.gmsh` you may want to pull the file into your project, as it's scheduled to be removed [January 2025](https://github.com/mikedh/trimesh/blob/f7df971d8196fa226eee057fd69f2e76a3704f1c/trimesh/interfaces/gmsh.py#L15-L24).

PR #2281, opened 2024-09-04
Fix typing bug in merge_vertices function

ayersb:
Small typing bug that got flagged by pyright when my team updated trimesh recently. 

The python typing for a few arguments is `Optional[bool]` when it should be `Optional[int]` based on the type comments and downstream code.

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -1112,9 +1112,9 @@ class Trimesh(Geometry3D):
         self,
         merge_tex: Optional[bool] = None,
         merge_norm: Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the fix!! I also added the corrected hints to `trimesh.grouping.merge_vertices` and updated the type to `trimesh.typed.Integer` which includes `numpy.int64` in addition to Python `int` etc in https://github.com/mikedh/trimesh/pull/2278/commits/c5b9a6b7618c262e340de6d1b0d8da49751b459f

PR #2282, opened 2024-09-06
add support for path3d export to ply format

moberweger:
This PR adds support to export `Path3D` object into the `PLY` format. This can be used to visualize the `Path3D` objects in common 3D mesh viewers, such as MeshLab. These are shown as line stripes, as shown below.
This PR contains:
- add export and import capabilities for `PLY` format that uses the `edge` property of the `PLY` format
- add basic unittest that saves and loads a simple `Path3D`

Screenshot from 2024-09-06 16-01-43

### DIFF
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -248,19 +248,71 @@ class ExportTest(g.unittest.TestCase):
         # the scene should be identical after export-> import cycle
         assert g.np.allclose(loaded.extents / source.extents, 1.0)
 
-    def test_ply_path(self):
+    def test_ply_path_empty(self):
         """
-        Should be able to load a path and export it as a PLY
+        Test empty path export does not fail
         """
-        path3D = g.trimesh.load_path([(0, 0, 0), (1, 0, 0), (1, 1, 0)])
+        path3D = g.trimesh.path.Path3D()
+
+        # export to ply
+        ply = path3D.export(file_type="ply")
+        assert len(ply) > 0
+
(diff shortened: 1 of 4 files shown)


mikedh:
Awesome thanks for the PR!! It seems like we could probably merge this in with `trimesh.exchange.ply` and have the load function return either a `Trimesh` or `Path3D`? Do we need the separate `trimesh.path.exchange.ply` module? I guess since PLY is a container format the choice is between confusing return types or confusing function calls :smile:. 
 

My only other note is we could probably support any entity type pretty easily by using the `entity.discrete(path.vertices) -> (n, 2) float64` which for a `Line` will do the same thing as the current code:

```
if isinstance(e, Line):
    for pp in range(len(e.points) - 1):
        entities.append((e.points[pp], e.points[pp + 1]))
else:
    raise NotImplementedError(f"type {type(e)} not supported as entities")
```

mikedh:
Thanks for the PR and fixes!

mikedh:
Sorry for the slow release, I'll try to get this out next week. I was testing this and I was wondering are there any existing packages that output or load PLY paths in this format? 3JS maybe?

moberweger:
Thanks for looking into this so thoroughly. The almighty `Open3D` library can save the `ply` format, it is called [LineSet](https://www.open3d.org/docs/release/python_api/open3d.geometry.LineSet.html) over there. I used it before porting this stuff to `trimesh`, but Open3D was too heavy for my usecase.

PR #2283, opened 2024-09-09
fix for #2269 with new test case for incomplete creation.revolve()

bguan:
Added a cap: bool parameter to creation.revolve(...) and implement additional logic to add end cap faces to an incomplete revolution so that the resultant mesh is a valid volume.

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -312,6 +312,22 @@ class CreationTest(g.unittest.TestCase):
         assert all(s.volume > 0 for s in split)
 
 
+    def test_revolve(self):
+        # create a cross section and revolve it to form some volumes
+        cross_section = [[0, 0], [10, 0], [10, 10], [0, 10]]
+
+        # high sections needed so volume is close to theoretical value for perfect revoulution
+        mesh360 = g.trimesh.creation.revolve(cross_section, 2 * g.np.pi, sections=360)
+        mesh360_volume = g.np.pi * 10**2 * 10
+        assert g.np.isclose(mesh360.volume, mesh360_volume, rtol=.1)
+        assert mesh360.is_volume, "mesh360 should be a valid volume"
+
+        mesh180 = g.trimesh.creation.revolve(cross_section, g.np.pi, sections=180, cap=True)
+        assert g.np.isclose(mesh180.volume, mesh360.volume/2, rtol=.1), \
+            "mesh180 should be half of mesh360 volume"
+        assert mesh180.is_volume, "mesh180 should be a valid volume"
+
+
 def check_triangulation(v, f, true_area):
     assert g.trimesh.util.is_shape(v, (-1, 2))
     assert v.dtype.kind == "f"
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome thanks for the PR looks good!

Yeah with the `cap` arguments the repairs might be needed because `triangulate_polygon` can sometimes depending on arguments insert vertices, which totally messes up the index. We've seen that in `slice_mesh` and `sweep` too, maybe we should add a `triangulate_polygon(..., force_same_vertices: bool = False)` argument which modifies the arguments and raises if the triangulation engine inserted anything.

I think we should also probably be raising in `tol.strict` mode here just to make sure in simple cases we've indexed correctly:
```
    # HACK: minor repairs if needed
    if not closed and cap and not mesh.is_volume:
      mesh.update_faces(mesh.nondegenerate_faces())
      mesh.update_faces(mesh.unique_faces())
      mesh.remove_infinite_values()
      mesh.remove_unreferenced_vertices()
      mesh.fix_normals()
```

bguan:
Thanks @mikedh.  Sorry for not checking the formatting and leaving a bunch of unnecessary white spaces around triggering those test failures. I will amend the next PR to take out those pesky trailing whitespaces.  

As you've confirmed my fear that ***triangulate_polygon(...)*** can sometimes add new vertices, I can change my capping logic to just use and add the vertices returned by it instead of reusing existing vertices, use trigonometry to transform them to find the end cap vertices and also add them, use ***cap_0_faces*** indices to create ***cap_end_faces***... this will ensure correct behavior even if ***triangulate_polygon(...)*** introduces new vertices.  BTW is there a mesh repair to consolidate duplicate or close enough vertices that takes care of changing face indices? As for your idea of adding a ***force_same_vertices*** param to ***triangulate_polygon(...)*** I will leave you or others to implement it :smile:

Also to clarify about your comment on my HACK checks... did you mean I should amend my repair conditions to be like this:
```
    #  vvvvvvvvvv
    if tol.strict and not closed and cap and not mesh.is_volume:
        ...
```
I can certainly do that in an updated PR.

Also wondering if you think my repairs are excessive.  For example:
```
      ...
      mesh.remove_infinite_values()
      mesh.remove_unreferenced_vertices()
      ...
```

Thanks again, and all suggestions welcome.

mikedh:
> As you've confirmed my fear that triangulate_polygon(...) can sometimes add new vertices, I can change my capping logic to just use and add the vertices returned by it instead of reusing existing vertices, use trigonometry to transform them to find the end cap vertices and also add them, use cap_0_faces indices to create cap_end_faces... this will ensure correct behavior****

Yeah one of the challenges of supporting [multiple triangulation backends](https://github.com/mikedh/triangulator_benchmarks/) is sometimes they insert vertices. This has come up elsewhere too, I'll add the `force_same_vertices` boolean to `triangulate_polygon` and raise a ValueError if it inserts so from the point of view of this function we can just rely on it not inserting. 

> Also wondering if you think my repairs are excessive.

Heh yeah it probably shouldn't need the finite/unreferenced check. In general I think we probably shouldn't need to run any repairs if we've done the indexing carefully and there have been no inserted vertices. 

We can also probably reconstruct the indexes from a subset with a `radius, index, = scipy.spatial.cKDTree(mesh_vertices).query(new_vertices)` and then `assert (radius < 1e-12).all()`.

mikedh:
Thanks for the PR!! I'll add the `force_same_vertices: bool` to `triangulate_polygon` so we don't have to handle that in this cap logic.

PR #2286, opened 2024-09-10
Convert to numpy array if need

michaelbratsch:
Due to commit https://github.com/mikedh/trimesh/commit/7f2aa95f6732fe32b91597394675ec6cc0197725 related to issue https://github.com/mikedh/trimesh/pull/2257, the function `normals` does not accept containers such as lists for the parameter `triangles` anymore. This was possible before due to the call of `cross` which itself could handle that.


mikedh:
Thanks for the fix!!

PR #2305, opened 2024-10-18
Fix missing visual copy for primitives

clemense:
This fixes the following problem:
```
import trimesh
b = trimesh.primitives.Box()
b.visual.face_colors[:] = [0, 0, 255, 255]

assert b.visual.defined == b.copy().visual.defined
```
This in turn affects `dump` and `export`.

I added `include_visual` as a parameter just to be consistent with `base.py::Trimesh::copy`.


mikedh:

Issue #2306, opened 2024-10-23
Boolean unit tests fail if manifold3d is not installed


onitake:
Some of the unit tests have been set up so they execute with all available boolean mesh operation engines, but some of them require the default manifold3d engine.

The Debian build system requires that all dependencies must be available in the Debian ecosystem, but we haven't packaged manifold3d yet and need to run all unit tests against the blender engine for now.

Example error:
```
______________________________________________________________________________________________ test_multiple_difference _______________________________________________________________________________________________

    def test_multiple_difference():
        """
        Check that `a - b - c - d - e` does what we expect on both
        the base class method and the function call.
        """
    
        # make a bunch of spheres that overlap
        center = (
            np.array(
                [
                    [np.cos(theta), np.sin(theta), 0.0]
                    for theta in np.linspace(0.0, np.pi * 2, 5)
                ]
            )
            * 1.5
        )
        # first sphere is centered
        spheres = [g.trimesh.creation.icosphere()]
        spheres.extend(g.trimesh.creation.icosphere().apply_translation(c) for c in center)
    
        # compute using meshes method
>       diff_base = spheres[0].difference(spheres[1:])

test_boolean.py:223: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../trimesh/base.py:2926: in difference
    return boolean.difference(
../trimesh/boolean.py:47: in difference
    return _engines[engine](meshes, operation="difference", **kwargs)
../trimesh/boolean.py:137: in boolean_manifold
    mesh=Mesh(
../trimesh/exceptions.py:34: in __call__
    raise super().__getattribute__("exception")
[...]

PR #2309, opened 2024-10-26
Make boolean engine selection more flexible

onitake:
Closes: #2306 

As announced, here's a patch that refactors the boolean engine autodetection.

I had to refactor a few things to make the two engines behave the same way and avoid circular imports.

`reduce_cascade` is now in `util.py`. I assumed that this is needed in other places, so I didn't move it into the new interfaces.manifold module.

`interfaces.blender` and `interfaces.manifold` are now dynamic imports. This avoids a circular import loop and allows autodetection at load time.

### DIFF
--- a/tests/test_boolean.py
+++ b/tests/test_boolean.py
@@ -5,16 +5,12 @@ except BaseException:
 
 import numpy as np
 
-try:
-    import manifold3d
-except BaseException:
-    manifold3d = None
 
-
-engines = [
-    ("blender", g.trimesh.interfaces.blender.exists),
-    ("manifold", manifold3d is not None),
-]
+# test only available engines by default
+engines = g.trimesh.boolean.available_engines
+# if we have all_dep set it means we should test all engines
+if g.all_dependencies:
+    engines = g.trimesh.boolean.all_engines
 
 
 def test_boolean():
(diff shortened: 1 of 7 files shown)


mikedh:
Awesome thanks for the PR!! Few quick thoughts:

- is there any reason we couldn't use `trimesh.util.has_module` for the autodetection? It seems pretty much the same idea as here.
- I've been meaning to move `trimesh.boolean.reduce_cascade` and `trimesh.util.chain` into their own `trimesh.iteration` submodule for the "materially similar to functions in `itertools`" stuff. I'll do that on merge which should reduce circular import problems.

Is there a side effect of `try: import manifold3d`? Because we could also pick the default by  using the first one that isn't an `ExceptionWrapper`:
```
# which backend boolean engines                                                            
_engines = {
    "manifold": boolean_manifold,
    "blender": interfaces.blender.boolean,
    "junk": ExceptionWrapper(ValueError("nah")),
}
_engines[None] = next(
    (v for v in _engines.values() if not isinstance(v, ExceptionWrapper)), None
)
```

onitake:
Looks like it didn't quite work as well as I had hoped.

Thanks for the pointers, I'll take a look at `has_module`.

PR #2314, opened 2024-10-30
Update blender_boolean.py.tmpl

Hellseher:
Fix build.

### DIFF
--- a/trimesh/resources/templates/blender_boolean.py.tmpl
+++ b/trimesh/resources/templates/blender_boolean.py.tmpl
@@ -36,7 +36,7 @@ if __name__ == '__main__':
     bpy.ops.object.delete(use_global=True)
 
     # get temporary files from templated locations
-    mesh_pre = $MESH_PRE
+    mesh_pre = '$MESH_PRE'
     mesh_post = os.path.abspath(r'$MESH_POST')
 
     for filename in mesh_pre:  # use data.objects instead of context.scene.objects


mikedh:
Thanks for the PR! What was the crash you were seeing?

https://github.com/mikedh/trimesh/blob/9dbb10cf76c5fea6ed217b0d18e9c64b6cb16a2e/trimesh/interfaces/generic.py#L45

So this should be evaluated in the template to:
`mesh_pre = ["file_name_a", "file_name_b"]`

Where wouldn't this change emit:
`mesh_pre = '["file_name_a", "file_name_b"]'`

Which then the loop would be through characters of the string, rather than file names?

Hellseher:
Hi,

I've checked the history of updates which Guix index has, it looks like it was
valid on 3.10.7, which I can't reproduce it after upgrade to the 4.5.2. It might
be "self healed" between versions :-)

The link to a commit where the package was added https://git.savannah.gnu.org/cgit/guix.git/commit/?id=b4b93035c1c4dda16b8399fc77b6ef150831ef1d

Thanks,
Oleg

mikedh:
I tried this in a branch and it fails `test_boolean` for the reasons above. I think the `guix` package can probably update to the latest trimesh, and remove that patch. I recently added `blender` back to the docker tests so if you pull in main to this PR it should also be caught in CI:

```
usr/lib/python3.12/subprocess.py:571: CalledProcessError
=========================== short test summary info ============================
FAILED test_boolean.py::test_boolean - subprocess.CalledProcessError: Command '['/home/mikedh/.local/bin/blender',...
FAILED test_boolean.py::test_multiple - subprocess.CalledProcessError: Command '['/home/mikedh/.local/bin/blender',...
FAILED test_boolean.py::test_empty - subprocess.CalledProcessError: Command '['/home/mikedh/.local/bin/blender',...
FAILED test_boolean.py::test_multiple_difference - subprocess.CalledProcessError: Command '['/home/mikedh/.local/bin/blender',...
========================= 4 
```

Thanks for the followup from GUIX!

PR #2322, opened 2024-11-07
Fix vertex_attributes subdivisions and corresponding test added.

dbukenberger:
Hey Mike,

I think in the simple subdivision routine, the _vertex attribute_ computation somehow got a bit jumbled.
There is some linear triangle-edge interpolation going on - but the new attributes do not match the values that I would have expected.

Thus, I have simplified the per-edge attribute interpolation to be analog to the actual subdiv-vertex position computation and added a testcase to verify.

Cheers
Dennis


mikedh:
Looks good, thanks for the fix and test! Yeah I agree simpler is probably better here.

PR #2332, opened 2024-12-17
Also add map_Kd to the OBJ/MTL material kwargs

ChuangTseu:
While all the other key/values for the MTL material are provided directly on top of Trimesh's interpretation for SimpleMaterial, then accessible through material.kwargs which is useful for custom handling of the materials, map_Kd isn't. I've had a need for getting the map_Kd file path directly instead of the already loaded PIL.Image.

### DIFF
--- a/trimesh/exchange/obj.py
+++ b/trimesh/exchange/obj.py
@@ -338,6 +338,8 @@ def parse_mtl(mtl, resolver=None):
                 # load the bytes into a PIL image
                 # an image file name
                 material["image"] = Image.open(util.wrap_as_stream(file_data))
+                # also store the original map_kd file name
+                material[key] = file_name
             except BaseException:
                 log.debug("failed to load image", exc_info=True)
 


mikedh:
Thanks for the PR! I guess this would be accessed through: `mesh.visual.material.kwargs['file_name']`? That seems like it could be a little ambiguous, what if just added it on as a runtime patch value which isn't the greatest but is probably a little clearer:

```
img = Image.open(util.wrap_as_stream(file_data))
# add the file name on as a runtime attribute
img.file_name = file_name
material["image"] = img
```

I was also thinking about making a `class LazyImage(PIL.Image)` which didn't actually load the image until it was accessed, maybe at some point in the future.

mikedh:
Actually `PIL.Image` has a `.info` dict, I put it there and added a test:
`assert m.visual.material.image.info["file_path"].endswith("fuze uv.obj")`

Thanks for the PR!

ChuangTseu:
Was away for a bit, just saw you merged my PR, thank you! To answer your previous question (though not needed anymore) yeah I think it's better to keep the raw "map_kd" key value like it's already done for the other MTL keys, rather than change it to "image". 

And thank you for the second change too, I may actually use that one :)

PR #2339, opened 2025-01-14
fixed not loading point cloud colors from glb format files

henryzhengr:
Updates the GLTF/GLB reader to include point cloud colors if available.


henryzhengr:
@mikedh Seems like there's version issues with the test checks

mikedh:
Thanks for the PR!! Yeah don't worry about the test, there's a pretty large outstanding release that's coming up. Change looks good! Would it be possible to add this on export as well? That would let us to make a quick unit test which round-trips a point cloud with color and checks that the colors survive.

mikedh:
Ah I had forgotten we supported loading pointcloud with color from GLB haha. I added a roundtrip test in https://github.com/mikedh/trimesh/commit/2a632593be311d5a25af227caff19ee1242fa994 and it looks good! Thanks for the PR!!

PR #2343, opened 2025-01-22
Add x86 condition for openctm

nepfaff:
OpenCTM doesn't support Mac ARM:
```
ERROR: Could not find a version that satisfies the requirement openctm==0.0.6 (from versions: none)
ERROR: No matching distribution found for openctm==0.0.6
```

This follows the style from the `embreex` dependency that has similar issues.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,7 +96,7 @@ recommend = [
     "scikit-image",
     "fast-simplification",
     # "python-fcl", # do collision checks # TODO : broken on numpy 2
-    "openctm", # load `CTM` compressed models
+    "openctm; platform_machine=='x86_64'", # load `CTM` compressed models
     "cascadio", # load `STEP` files
 ]
 


mikedh:
Thanks for the PR! Just out of curiosity, how are you managing your trimesh dependency? There are multiple "helper" extras, none of which are required, and depending on anything other than `trimesh[easy]` is unlikely to work on all platforms.

For upstream projects, it's a lot easier to manage the poetry/UV/requirements lock files with exactly the soft dependencies that a project actually uses, i.e. `trimesh, scipy, xatlas` rather than using an extra, since just trimesh will only pull in numpy.

nepfaff:
We currently depend on `trimesh[recommend]` as we need some of the dependencies there for parts of trimesh that we are using. However, only depending on trimesh and the exact soft dependencies that we actually need is a good suggestion and seems like the most robust way to handle this.
Thanks!

PR #2346, opened 2025-01-29
Fix/forward kwargs

michaelbratsch:
With the pull-request https://github.com/mikedh/trimesh/pull/2241, `load_mesh` does not respect the parameter `process` anymore.

With this pull-request, the `kwargs` are being forwarded in `load_stl`.

### DIFF
--- a/tests/test_loaded.py
+++ b/tests/test_loaded.py
@@ -31,6 +31,18 @@ class LoaderTest(g.unittest.TestCase):
         assert len(m.faces) > 0
         assert m.area > 1e-5
 
+    def test_load_mesh(self):
+        # test the influence of the parameter `process` for `load_mesh`
+        with open(g.os.path.join(g.dir_models, "featuretype.STL"), "rb") as file_obj:
+            mesh = g.trimesh.load_mesh(file_obj=file_obj, file_type="stl", process=False)
+            # check that number of vertices is not being reduced
+            assert len(mesh.vertices) == 10428
+
+        with open(g.os.path.join(g.dir_models, "featuretype.STL"), "rb") as file_obj:
+            mesh = g.trimesh.load_mesh(file_obj=file_obj, file_type="stl", process=True)
+            # check that number of vertices is being reduced
+            assert len(mesh.vertices) == 1722
+
     def test_fileobj(self):
         # make sure we don't close file objects that were passed
         # check load_mesh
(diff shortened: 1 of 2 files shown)


michaelbratsch:
The fix of this pull-request seems to cause a regression in the existing unit tests: https://github.com/mikedh/trimesh/blob/aed7eee9d4351c18869c05c030fc399738441c5d/tests/test_stl.py#L18

@mikedh What is the supposed way in the current design to forward the `kwargs`?

mikedh:
Thanks for the fix and great test! I think we have to do this in `load_scene` to replicate the old behavior, this needs some work but passes your test: https://github.com/mikedh/trimesh/pull/2347/commits/907d7dd15ba3808c00c7a597f30974b93db43b5f

michaelbratsch:
Thank you for your quick reply and immediately taking care of it ❤️

I am also wondering about a related issue which is not being fixed with the current pull-request. I put the questions along with this example that way it is probably easiest to see. If wanted, I can open a proper issue for it.
```python
import trimesh

file_name = "models/featuretype.stl"

with open(file_name, "rb") as fh:
    mesh = trimesh.load_mesh(file_obj=fh, file_type="stl", process=False)

    assert len(mesh.faces) == 3476
    assert len(mesh.vertices) == 10428

with open(file_name, "rb") as fh:
    mesh = trimesh.load_mesh(file_obj=fh, file_type="stl", process=True)
    
    # Why is only the number of vertices being reduced but not the number of faces?
    assert len(mesh.faces) == 3476
    assert len(mesh.vertices) == 1722
    
    mesh.export('./mesh_test.stl')

with open('./mesh_test.stl', 'rb') as fh:
    mesh = trimesh.load_mesh(file_obj=fh, file_type="stl", process=False)
    
    assert len(mesh.faces) == 3476
    # Why is the number of vertices again 10428 even though it is from the exported reduced model?
    assert len(mesh.vertices) == 10428
```

mikedh:
No worries! I'm going to check whether we have the same bug in the other load entrypoints before merging.

> Why is the number of vertices again 10428 even though it is from the exported reduced model?

That's actually because STL just outputs a "bundle of triangles" with no indexes (hence the need for the vertex merging on load, since trimesh started as just an STL loader haha). If you use a format with indexes (OFF, OBJ*, GLB, PLY, etc) it should pass as you expect:

```
    mesh.export('./mesh_test.off')

    with open('./mesh_test.off', 'rb') as fh:
        mesh = trimesh.load_mesh(file_obj=fh, file_type="off", process=False)
        assert len(mesh.faces) == 3476                                                       
        assert len(mesh.vertices) == 1722
```

michaelbratsch:
All right, got it. Thanks for the explanation.

PR #2355, opened 2025-02-11
chardet -> charset-normalizer

mjd3:
This PR removes `chardet` as an `easy` dep and replaces it with `charset_normalizer` which is faster, has a smaller wheel, is more robust, and has an MIT license to match this project.

More info here: https://github.com/jawah/charset_normalizer


mikedh:
Looks like a nice project and solid improvement over chardet, thanks for the update!

PR #2356, opened 2025-02-12
Add vertex attributes in path

StylianosVitalis-TomTom:
This PR adds support for `vertex_attributes` in paths, to match the functionality of `Trimesh` when importing/exporting custom attributes for lines from/to glTF.


mikedh:
Looks great! Thanks for the feature and tests!

PR #2362, opened 2025-02-26
Efficient Laplacian calculations

nickeopti:
The current implementation of `trimesh.smoothing.laplacian_calculation` relies on list comprehensions. The case of `equal_weight=True`, at least, can be computed directly using sparse matrix operations. This PR does that, which turns out to be nearly 20 times faster.

I have been unable to figure out a way of computing the `equal_weight=False` case using sparse matrix operations, so I have let the current implementation of that be. This implies quite a performance difference between the two flag values. This is noted in the docstring, which is now also more descriptive.

Correctness tests are added, which include a verbatim copy of the original implementation to test against.

Performance characteristics:
```python
In [1]: import trimesh

In [2]: from tests import test_smoothing

In [3]: m = trimesh.creation.icosphere(8, 10)

In [4]: m.vertices.shape, m.faces.shape
Out[4]: ((655362, 3), (1310720, 3))

In [5]: %timeit trimesh.smoothing.laplacian_calculation(m, equal_weight=True)
44.4 ms ± 783 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit test_smoothing.explicit_laplacian_calculation(m, equal_weight=True)
790 ms ± 9.89 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit trimesh.smoothing.laplacian_calculation(m, equal_weight=False)
2.87 s ± 31.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit test_smoothing.explicit_laplacian_calculation(m, equal_weight=False)
2.91 s ± 34.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
where `test_smoothing.explicit_laplacian_calculation` is the original implementation.

### DIFF
--- a/tests/test_smoothing.py
+++ b/tests/test_smoothing.py
@@ -4,7 +4,87 @@ except BaseException:
     import generic as g
 
 
+
+def explicit_laplacian_calculation(mesh, equal_weight=True, pinned_vertices=None):
+    """
+    Exact copy of non-sparse calculation of laplacian, for correctness testing.
+
+    Calculate a sparse matrix for laplacian operations.
+    Parameters
+    -------------
+    mesh : trimesh.Trimesh
+      Input geometry
+    equal_weight : bool
+      If True, all neighbors will be considered equally
+      If False, all neighbors will be weighted by inverse distance
+    Returns
+    ----------
+    laplacian : scipy.sparse.coo.coo_matrix
+      Laplacian operator
+    """
+    # get the vertex neighbors from the cache
+    neighbors = mesh.vertex_neighbors
+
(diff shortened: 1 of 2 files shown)


mikedh:
Looks like a huge improvement, thanks for the PR and great test! It seems like since `equal_weight=False` is not called by the internal uses of the function it's pretty reasonable for the flag to have a large performance penalty, since people should know what they're asking for. 

Don't worry about the CI failing, it's just some minor housekeeping.

PR #2365, opened 2025-02-27
[WIP] Pyglet 2

mikedh:
A rough draft towards supporting Pyglet 2. Keeps the Pyglet 1 viewer which may be deprecated at some point but not for quite a while. Pyglet 2 requires basically a viewer re-write, and this is currently showing a single mesh but needs quite a bit more work (instanced meshes, etc). 

This applies the following moves and deprecations: 
- `trimesh.viewer.widget` -> `examples/widget`
  - Applies the deprecation warning for the move of the widget into the examples folder. 
- `trimesh.viewer.windowed` -> `trimesh.viewer.pyglet1`
  - Keeps the old viewer as the new one probably needs some time, and to avoid breaking people's Python installs.
- `trimesh.rendering` -> `trimesh.viewer.pyglet1.conversion`
  - This was positioned as something more general than it probably was, which is a Pyglet specific indexed vertex list constructor. Still thinking about the right way to deprecate this. 
- `trimesh.viewer.pyglet2`
  - New viewer. 

Other changes:
- Viewer uses the matrices provided by the `Camera` object which should help with people who want to render things in the same frames as the scene. This should also make it relatively easy to support orthographic cameras.
- Viewer needs shaders for points, lines, textured meshes, etc. Shaders should be stored in the library for easy editing.
- This should probably either wait to release until `pyglet.Models` stabilizes or use a lower level construct.


YoniChechik:
have you considered integration with www.rerun.io ?  They already did all the visualization heavy lifting

caffeinepills:
Hello! I contribute and help maintain pyglet. I see with this PR that you had a goal to add support for newer pyglets, but dropped it. I wouldn't mind working on this to help add support for the modern paths in trimesh, but would like to discuss what is the blocking issue with this? Thanks!

mikedh:
Hey thanks for looking at this! This got auto-closed on some branch shenanigans, I do think we should merge this at some point. Some quick notes on my own ancient PR haha:

- One of the long-term asks from many issues is having the `Scene` camera matrix match the debug viewer exactly. It looks like this branch does that and maybe even tests it. 
- Since I did this reorg `pyglet3` has released, maybe instead of the branch's current `viewer.{pyglet1, pyglet2}` we should put pyglet1 in `viewer.pyglet.legacy` and then have the rest of the stuff be Pyglet2 and Pyglet3.
- I don't remember what I was doing but I think I ran into something with the Model API which has almost certainly stabilized by now? Honestly just a quick run-through from someone who knows what they're doing in Pyglet and making some notes would be SUPER helpful! At a high level I was on the fence if we should try to use "lowest common denominator" Pyglet API's and do more with our own shaders, specifically pulling in [Matt's PBR shaders](https://github.com/mmatl/pyrender/tree/master/pyrender/shaders) to the core viewer.
- The basic test matrix for both Pyglet1 and Pyglet2+ just checking that nothing looks insane (this may pass now): 
  - meshes with vertex color and face colors
  - meshes with texture (SimpleMaterial, PBRMaterial: fuze.obj, cesiummilktruck)
  - Point clouds
  - Path2D, Path3D
  - Voxels 
  - Scenes of all the above

caffeinepills:
> Hey thanks for looking at this! This got auto-closed on some branch shenanigans, I do think we should merge this at some point. Some quick notes on my own ancient PR haha:
> 
No worries! 
> * Since I did this reorg `pyglet3` has released, maybe instead of the branch's current `viewer.{pyglet1, pyglet2}` we should put pyglet1 in `viewer.pyglet.legacy` and then have the rest of the stuff be Pyglet2 and Pyglet3.

To be clear 3.0 isn't released officially, but it is close. We do have some development versions out. Mostly have been using it as a good time to improve and change some APIs over the version changes and get input from users. Overall, for most users, the core behavior should be the same, just a lot of backend changes. So, targeting both shouldn't be too tough. 

> * I don't remember what I was doing but I think I ran into something with the Model API which has almost certainly stabilized by now? Honestly just a quick run-through from someone who knows what they're doing in Pyglet and making some notes would be SUPER helpful! At a high level I was on the fence if we should try to use "lowest common denominator" Pyglet API's and do more with our own shaders, specifically pulling in [Matt's PBR shaders](https://github.com/mmatl/pyrender/tree/master/pyrender/shaders) to the core viewer.

There have been some model changes, and should stable. I think a lot of changes for model in pyglet 2.x were held back due to no API for instanced rendering, which is now supported with pyglet 3.0.

> * The basic test matrix for both Pyglet1 and Pyglet2+ just checking that nothing looks insane (this may pass now):
>   
>   * meshes with vertex color and face colors
>   * meshes with texture (SimpleMaterial, PBRMaterial: fuze.obj, cesiummilktruck)
>   * Point clouds
>   * Path2D, Path3D
>   * Voxels
>   * Scenes of all the above

I will take a look. I'm not too familiar with trimesh, so I will try to limit myself to where the pyglet code intersects. However, looking at your feature list and they all seem to pass on my end.  Same with the camera tests, so I assume the matrix stuff is good? 

If you have any examples that don't work properly, behavior not implemented or is not working as expected, let me know.  I also tried a basic example of rendering a box and sphere and it ran correctly. I can take a closer look at the pyglet2/ folder as well if there is anything specific you aren't sure on. EDIT: I will also start a pyglet3/ folder and get that working as well.

Also I will mention, if you have time and use Discord, you can always visit the Pyglet discord, and we can discuss things more directly there as well.

PR #2367, opened 2025-03-03
Fix complex path export

StylianosVitalis-TomTom:
This PR will fix exporting of complex paths (i.e., having multiple `Line` entities).

This will still fail for other entities than `Line` that need proper discretization, but there seems to be no trivial way to solve this problem for curves.

### DIFF
--- a/tests/generic.py
+++ b/tests/generic.py
@@ -26,6 +26,7 @@ import contextlib
 import subprocess
 import collections
 import numpy as np
+import pytest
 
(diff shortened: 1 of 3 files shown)


mikedh:
Thanks for the PR and sorry for the delay! Looks good, I made minor changes, the largest of which was warning for attributes on paths with non-Line entities and dropping the attributes in the export rather than raising. Should be released now!

PR #2385, opened 2025-04-02
Make merge_vertices consistent regardless of caching normals

michaelbratsch:
`merge_vertices` acts differently depending on whether vertex normals have been cached or not, see https://github.com/mikedh/trimesh/blob/cc0d3231814d3afe5c0a693d77f4f840668da1a7/trimesh/grouping.py#L94

To have consistent behaviour the vertex normals are now computed if they are not in cache and added to the stack if `merge_norm` is `False`. That way the added unit test is passing.

This fix can't be merged as is because it seems to have some side effects and unit tests are failing. I still opened this pull request because it seems to me that `merge_vertices` should not depend on whether something is in the cache or not. 

@mikedh  Please feel free to make any modifications or close this pull request.

### DIFF
--- a/tests/test_merge.py
+++ b/tests/test_merge.py
@@ -47,6 +47,26 @@ class MargeTest(g.unittest.TestCase):
         assert copied.euler_number == 2
         assert copied.referenced_vertices.sum() == 8
 
+    def test_caching(self):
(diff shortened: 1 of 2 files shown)


michaelbratsch:
@mikedh I am a bit puzzled why so many unit tests are failing. Are they all relying on the fact that `vertex_normals` are not in the cache? This feels kind of arbitrary, right?

PR #2389, opened 2025-04-16
Publish package using official GitHub action

EpicWink:
Trusted publishing (with attestations) means I can know for certain that what I download from PyPI is the same artefact which was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing (rather than having to manually review all of the installed files on each release).

See [the Python packaging documentation](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing) and [the official pypi-publish GitHub action documentation](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) on trusted publishing - you'll need to configure an environment in PyPI and GitHub. You will be able to remove the `PYPI_TOKEN` project secret.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,6 +56,11 @@ jobs:
     name: Release To PyPi
     needs: [tests, containers, corpus]
     runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/trimesh
+    permissions:
+      id-token: write
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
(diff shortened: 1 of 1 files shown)


mikedh:
Looks good, thanks for the update! I added trimesh as a trusted publisher and removed the token, hopefully this releases smoothly.

mikedh:
I don't think we need the `environment` values right, just the `id-token: write`? I don't think [it's required](https://github.com/trimesh/vhacdx/blob/main/.github/workflows/release.yml#L52-L69).

EpicWink:
@mikedh Setting [`environment`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idenvironment) does a couple of things:

* protects the job from running in the wrong conditions, eg by another PR author, or in the wrong branch/tag (if you configure it in the GitHub project settings)
* show the environment (under "Deployments") in the GitHub project's main page
* ensures the upload to PyPI comes only from the specified environment (when [configured](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) in PyPI management)

See [the documentation](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment).

Adding an environment is optional, but recommended to prevent accidental uploads to PyPI. Up to you, IMO.

PR #2395, opened 2025-04-22
Full resolution is maintained when saving an image of the scene

TaqiHamoda:
This pull request addresses an issue where the `render_scene` function, when used in the `save_image` method, could produce images with unexpected resolutions. This occurred because the rendering window was being maximized by default, which, in cases where the specified resolution exceeded the display's native resolution, would result in the image being rendered at the display's resolution. Additionally, the window was resizable, allowing users to accidentally alter the intended resolution.

**Key Changes:**

* **Disabled automatic maximization:** The `fullscreen` parameter in `render_scene` is now set to `False` by default within the `save_image` context. This ensures that the rendering window adheres to the provided `resolution` parameter, even if it's higher than the display resolution.
* **Disabled window resizing:** The `resizable` parameter in `render_scene` is now set to `False` by default within the `save_image` context. This prevents users from manually resizing the rendering window and potentially distorting the dimensions of the captured image.

**Rationale:**

These changes ensure that the `save_image` method produces images with the exact resolution specified by the user, regardless of the display's native resolution. This provides greater control and predictability over the output image dimensions.

**Important Note:**

These changes exclusively affect the behavior of `render_scene` when called by the `save_image` method. The `show` method retains its original behavior, allowing for window maximization and resizing.

### DIFF
--- a/trimesh/scene/scene.py
+++ b/trimesh/scene/scene.py
@@ -1053,7 +1053,7 @@ class Scene(Geometry3D):
         """
         from ..viewer.windowed import render_scene
 
-        return render_scene(scene=self, resolution=resolution, **kwargs)
+        return render_scene(scene=self, resolution=resolution, fullscreen=False, resizable=False, **kwargs)
 
     @property
     def units(self) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR! Setting those from the constructor arguments makes sense to me.

PR #2401, opened 2025-05-02
fix check_volume variable propogation during mesh boolean operations

kaaviyaalt:
This pull request provides a patch fix for the bug reported in issue #2253 by including two changes:

- Adding the check_volume variable to the `kwargs` of the boolean operations for propogation.
- Introducing a volume validity check for the Blender engine, similar to the one used prior to calling the Manifold engine.


kaaviyaalt:
Hi Mike, Thanks for your prompt response :) I have added a test case, do let me know if anything else needed.

Issue #2165, opened 2024-02-27
Voxel to Mesh Watertightness Question


Eric-Vin:
I'm trying to convert a voxel grid exactly to a mesh, i.e. the resulting mesh should contain the exact volume of the original voxel grid. I'm having a problem with the resulting watertightness though, as it looks like some of the edges are connected to 4 faces, which conflicts with Trimesh's definition of watertightness (exactly 2 faces to every edge). Is it possible to do such a conversion from voxels to meshes exactly in Trimesh?

I've attached an image illustrating the problem, with `fix_broken_faces` having been used to color the problematic faces red. 

![image](https://github.com/mikedh/trimesh/assets/8935814/886046ad-8ec8-464a-a36a-86a2f8c25dfa)

mikedh:
Hey, yeah does it work if you disable vertex merging? Is the "voxel" input a mesh? You probably need to pass `Trimesh(..., process=False)` wherever the mesh is being created?

Eric-Vin:
Thanks for the response! I suspect it would work but I'm using the triangles `to_kwargs` constructor so I think I'll have to refactor to use faces and vertices to try it. I'm working from a Trimesh VoxelGrid and creating triangles for each of the faces on the surface, so I get a mesh that represents the exact same space. My main concern is if I turn off processing upon creation and that mesh gets used in a boolean operation down the line, it might not be a volume on the way out if processing is applied internally. Is there a way to represent this sort of thing as a proper watertight mesh in Trimesh with the vertices merged/processing applied?

Here's my code for reference if you want to see what I'm doing. It's a rough prototype but the goal is to more rapidly construct what you would get from unioning all the boxes from the `multibox` VoxelGrid method together:
```
    def mesh(self):
        # Extract values for original voxel grid and the surface of the voxel grid.
        dense_encoding = self.voxelGrid.encoding.dense
        hpitch = self.voxelGrid.pitch[0] / 2
        hollow_vr = trimesh.voxel.VoxelGrid(
            trimesh.voxel.morphology.surface(self.voxelGrid.encoding),
            transform=self.voxelGrid.transform,
        )

        surface_indices = numpy.argwhere(hollow_vr.encoding.dense == True)
        surface_centers = hollow_vr.indices_to_points(hollow_vr.sparse_indices)

        # Determine which faces should be added for each voxel in our extracted surface.
        point_face_mask_list = []

        def index_in_bounds(index):
[...]

PR #2404, opened 2025-05-08
Added remove_internal_faces parameter to multibox function

srasoulzadeh:
By default, the `trimesh.voxel.ops.multibox` function returns a Trimesh object with a box at every voxel tenter. However, boxes corresponding to adjacent voxel centers share redundant internal faces. This pull request modifies the this function  by adding the optional boolean `remove_internal_faces` parameter that removes internal faces between adjacent voxels (boxes), resulting in a more lightweight geometry useful for exporting and visualization.

**Notes:**
- Added `remove_internal_faces` param to `trimesh.voxel.ops.multibox`
    - When set to `False` (default), existing behavior is preserved (all faces included).
    - When set to `True`, shared faces between neighboring boxes are excluded from the output mesh.
- Added a corresponding unit test (`test_multibox.py`) for a 2 **3 binary voxel grid.
- All operations on face or vertex arrays implemented with vectorized numpy operations.
- This pull request maintains full backward compatibility.

### DIFF
--- a/tests/test_multibox.py
+++ b/tests/test_multibox.py
@@ -1,0 +1,47 @@
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+
+class MultiboxTest(g.unittest.TestCase):
+    def test_multibox_internal_face_removal(self):
+        np = g.np
+
+        # Create a small 2 * 2 * 2 binary voxelgrid
+        voxelgrid = np.ones((2, 2, 2), dtype=bool)
+        indices = np.argwhere(voxelgrid)
+        points = g.trimesh.voxel.ops.indices_to_points(indices)
+
+        # With internal faces (remove_internal_faces=False)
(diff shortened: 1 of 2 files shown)


srasoulzadeh:
This could also be a workaround for the issue raised here: https://github.com/mikedh/trimesh/issues/2165

mikedh:
Awesome thanks for the PR! The results look great, I was also able to add an assert for `is_volume` and `body_count == 1` which is nice. 

My one thought is, should `multibox` accept the boolean matrix (in addition to, or as a replacement for the voxel centers) to do the neighbor calculation? The list comprehension over `neighbor_coords` doing set checks did appear to get a little slower at larger voxelgrid sizes (13s at 100**3). I couldn't quite frame it correctly but it might be possible to generate `keep_mask` with advanced indexing on the matrix, i.e. something like `matrix[:,:,1:] != matrix[:,:,:-1]`. 

No worries if not, just thought it could be a nice speedup.

PR #2416, opened 2025-06-10
Updated integral_mean_curvature function

yajushikhurana:
The integral_mean_curvature (IMC) function did not agree with the analytical value for non-convex objects such as a torus and an annulus. The angle in the function needs to have an appropriate sign to detect concave angles. This code corrects that error and also adds a test function for IMC to test for non-convex objects.


mikedh:
Awesome thanks for the PR and test! Quick question, do we need another function and attribute for `vector_angle_with_sign`? Couldn't one use the existing `mesh.face_adjacency_convex` value?

```
In [12]: signs = np.array([-1.0, 1.0])[m.face_adjacency_convex.astype(np.int64)]

In [13]: angles = m.face_adjacency_angles * signs
```

mikedh:
Looks great thanks for the update!

PR #2420, opened 2025-06-27
Added kwargs passthrough to creation.capsule

squarra:
Very small change. All other creation methods have this already, only capsule was missing it.


mikedh:

PR #2421, opened 2025-06-30
Fix base frame naming bug

qiuqiweicn:
Bug
`if base_frame in names`: never triggered because names is keyed by index, not by node name, so a node named "world" could silently clash.

Fix
Reserve "world" in name_index before the node loop. The existing unique_name() logic now resolves any collision deterministically; the random fallback is removed.

### DIFF
--- a/trimesh/exchange/gltf.py
+++ b/trimesh/exchange/gltf.py
@@ -1692,17 +1692,16 @@ def _read_buffers(
     # the mutated dict in every loop
     name_index = {}
     name_counts = {}
+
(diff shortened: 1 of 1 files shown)


mikedh:
Good catch! Sometimes getting unique names can be quite tricky, looks like the tests aren't happy. Would just using `unique_name` on the base frame name work here:

```
base_frame = unique_name("world", contains=names.values())
```

alessiocancian-emax:
Hi @mikedh this is still an issue, why it was reverted and never released?

Issue #1117, opened 2021-01-12
uv value lost


iamnotstone:
there are some files when I export from blender there are UV values. But when I load them from trimesh, the uvs had gone.

please see the file below:

tmp.tmp.obj:
```
# Blender v2.82 (sub 7) OBJ File: ''
# www.blender.org
mtllib tmp.tmp.mtl
o Cube
v 1.000000 1.000000 -1.000000
v 1.000000 -1.000000 -1.000000
v 1.000000 1.000000 1.000000
v 1.000000 -1.000000 1.000000
v -1.000000 1.000000 -1.000000
v -1.000000 -1.000000 -1.000000
v -1.000000 1.000000 1.000000
v -1.000000 -1.000000 1.000000
vt 0.625000 0.500000
vt 0.875000 0.500000
vt 0.875000 0.750000
vt 0.625000 0.750000
vt 0.375000 0.750000
vt 0.625000 1.000000
vt 0.375000 1.000000
vt 0.375000 0.000000
vt 0.625000 0.000000
vt 0.625000 0.250000
vt 0.375000 0.250000
vt 0.125000 0.500000
vt 0.375000 0.500000
vt 0.125000 0.750000
vn 0.0000 1.0000 0.0000
vn 0.0000 0.0000 1.0000
vn -1.0000 0.0000 0.0000
vn 0.0000 -1.0000 0.0000
vn 1.0000 0.0000 0.0000
vn 0.0000 0.0000 -1.0000
usemtl Material
s off
f 1/1/1 5/2/1 7/3/1 3/4/1
f 4/5/2 3/4/2 7/6/2 8/7/2
f 8/8/3 7/9/3 5/10/3 6/11/3
f 6/12/4 2/13/4 4/5/4 8/14/4
f 2/13/5 1/1/5 3/4/5 4/5/5
f 6/11/6 5/10/6 1/1/6 2/13/6
```

there are many **vt** values in the obj file. but when I load it like this:
```
m = trimesh.load('./tmp.tmp.obj')
print(m.visual.uv)  ---> None
```

mikedh:
Hey, it's most likely discarding them as there are a different number of vt values than v values.

iamnotstone:
But this obj file works well in blender. Do you think we should not discard the **vt** values?

zachzeyuwang:
[...]

PR #2423, opened 2025-07-03
Simplify _parse_faces_vectorized to handle arbitrary faces

JulianKnodt:
Previously, this function was special cased for quads and tris, but was subtly wrong for quads with normals but not textures.
Instead, handle the general case, which should work for arbitrary polygons.

I think that technically #1117 should be closed for the original problem of loading a quad mesh, not by this PR, but it was fixed sometime after `4.6.10`, which is the version I had locally, but the problem was fixed on `main`.


mikedh:
Awesome thanks for the fix! Yeah it's definitely much nicer to not have a quad-only special case here.

PR #2424, opened 2025-07-07
Robustify DAE shininess parsing

nepfaff:
I found a lot of online .dae files with invalid `shininess` values. Parsers like `MeshLab` successfully ignore these values and I think `trimesh` should do the same.

As an example, the files from [ArtVIP](https://huggingface.co/datasets/x-humanoid-robomind/ArtVIP) contain these invalid shininess values:
```xml
<shininess>
  <color>0.000000</color>
</shininess>
```
The original code failed with a runtime error: `TypeError: unsupported operand type(s) for +: 'float' and 'tuple'`.
The updated code succeeds and prints the following: `Invalid shininess value: (0.0, 0.0, 0.0, 1.0), using default roughness`

### DIFF
--- a/models/cube_invalid_shininess.dae
+++ b/models/cube_invalid_shininess.dae
@@ -1,0 +1,88 @@
+<?xml version="1.0" encoding="utf-8"?>
+<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">
+  <library_effects>
+    <effect id="Material-effect">
+      <profile_COMMON>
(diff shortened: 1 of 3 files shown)


mikedh:
Sounds good to me, thanks for the PR! Is there any chance we could add a small model to tests? Also don't worry about the CI, that's been fixed in the open release PR.

mikedh:
Awesome thanks for the fix and test! I should probably add huggingface to the corpus tests at some point, although having to pull them every time might be a little much.

Issue #2431, opened 2025-07-30
Bug: `trimesh.creation.torus` generates incorrect mesh when angle is not 2π


birgerbr:
The mesh generated by `trimesh.creation.torus` does not work as expected for sections, i.e. when angle is not set to 2π.

Example code
```python
import numpy as np
import trimesh
trimesh.creation.torus(major_radius=2, minor_radius=1, angle=np.pi/2).show()
```

<img width="1034" height="727" alt="Image" src="https://github.com/user-attachments/assets/462cc186-3fc8-4e76-92a7-5b19a35be2cc" />

PR #2432, opened 2025-07-30
Fix torus section creation

birgerbr:
Fixes https://github.com/mikedh/trimesh/issues/2431

image


mikedh:
Awesome thanks for the fix!

PR #2441, opened 2025-08-15
utils.concatenate now also concats vert and face attributes

Nahuel-M:
I hope this makes sense for trimesh as a whole. It would greatly improve ergonomics for me :)

This addition will only concatenate vertex and face attributes if all keys are found in all meshes. An alternative is discarding attributes that are not in all meshes and concatenating the rest.


mikedh:
Thanks for the PR! I changed it to also check that every attribute has a correct length for every mesh, looks like that fixed the tests.

PR #2447, opened 2025-08-26
Fall back to linear_color_map if matplotlib is not available.

### LINKED ISSUE #2446: Missing `matplotlib` dependency
[`trimesh.visual.color.interpolate`](https://github.com/mikedh/trimesh/blob/d8c75b576bf0c6c5c8e263956332b2e50bc2775e/trimesh/visual/color.py#L914) imports `matplotlib`, but `matplotlib is not listed as a dependency in [`pyproject.toml`](https://github.com/mikedh/trimesh/blob/d8c75b576bf0c6c5c8e263956332b2e50bc2775e/pyproject.toml). This causes a `ModuleNotfoundError` to be raised when calling the function unless the user has manually installed `matplotlib`.

bicarlsen:
Resolve #2446 by falling back to [`linear_color_map`](https://github.com/mikedh/trimesh/blob/d8c75b576bf0c6c5c8e263956332b2e50bc2775e/trimesh/visual/color.py#L841) if `matplotlib` is not available.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ easy = [
     "vhacdx; python_version>='3.9'",
     # old versions of mapbox_earcut produce incorrect values on numpy 2.0
     "mapbox_earcut >= 1.0.2; python_version>='3.9'",
+    "matplotlib",
 ]
 
 recommend = [
--- a/trimesh/visual/color.py
+++ b/trimesh/visual/color.py
@@ -911,9 +911,12 @@ def interpolate(
     if color_map is None:
         cmap = linear_color_map
     else:
-        from matplotlib.pyplot import get_cmap
-
-        cmap = get_cmap(color_map)
+        try:
+            from matplotlib.pyplot import get_cmap
+            cmap = get_cmap(color_map)
+        except BaseException:
+            util.log.warning("matplotlib not available, falling back to linear_color_map")
+            cmap = linear_color_map
 
     # make input always float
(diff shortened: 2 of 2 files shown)


mikedh:
Thanks for the PR! Yeah agreed we should fall back on `linear_color_map` regardless. I'm hesitant to add `matplotlib` as a dependency since it's only used here for this tiny function and checking on a venv it increases the `trimesh[easy]` install from 276 mb -> 327mb, and I've been hoping to go in the other direction these days haha. 

Playing with the matplotlib colormaps right now, they are an 8 bit lookup table (256 values). I was playing with spline fitting of matplotlib's colormaps to avoid a dependency in https://github.com/mikedh/trimesh/commit/e73279d9ba60e17623acc48a22563e5edd60d202 It feels a little janky but gets pretty good results for sparsity, accuracy, and smoothness:


```
Color Map `viridis`
Channel R Spline parameters: (18,)
Channel R Control Error: Max 0.00113, Mean 0.00015+/-0.00012
Channel G Spline parameters: (15,)
Channel G Control Error: Max 0.00165, Mean 0.00014+/-0.00014
Channel B Spline parameters: (21,)
Channel B Control Error: Max 0.00157, Mean 0.00014+/-0.00014
```
The spline fit is a dashed line almost completely covered by the original line as they're very close: 

spline_viridis

Still debating whether it's worth it to include fitted colormaps, but I'll merge this for the fallback and remove the `matplotlib` dependency. Thanks for the fix!

PR #2450, opened 2025-09-02
Don't capture the whole exception in `ExceptionWrapper`

SquidDev:
`ExceptionWrapper` currently preserves the original `BaseException` instance. Unfortunately, this also includes the exception's traceback, including all stack frames and their locals.

As several trimesh modules are lazily loaded (e.g. `trimesh.path.polygons`), this can permanently leak any locals that were on the stack when a trimesh function is first called.

We now store the exception type and its message/arguments, and recreate the exception when accessing attributes instead. This does mean some information from the original exception is lost, but sadly I don't think that's avoidable.

### DIFF
--- a/trimesh/exceptions.py
+++ b/trimesh/exceptions.py
@@ -18,7 +18,7 @@ class ExceptionWrapper:
     """
 
     def __init__(self, exception):
-        self.exception = exception
+        self.exception = (type(exception), exception.args)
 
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the fix, makes sense to me! I was playing with it and the errors still seem quite easy to use. Explicitly storing just what we need to recreate the "you don't have scipy" message rather than the whole object seems a lot cleaner.

PR #2457, opened 2025-09-12
Material fixes for `trimesh.util.concatenate()`

brentyi:
Hi @mikedh!

We've been having some trouble with materials changing after I call `trimesh.util.concatenate()`.

Here's a code snippet to demonstrate. We create a red mesh, a green mesh, and a concatenated mesh:
```python
import numpy as np
import trimesh.util
from trimesh.creation import box
from trimesh.visual import TextureVisuals
from trimesh.visual.material import PBRMaterial


def make_boxes():
    red_box = box(extents=[1, 1, 1])
    red_box.visual = TextureVisuals(
        material=PBRMaterial(
            baseColorFactor=[1.0, 0.2, 0.2, 1.0],
            metallicFactor=0.0,
            roughnessFactor=1.0,
            emissiveFactor=[0.0, 0.0, 0.0],
        )
    )

    green_box = box(extents=[1, 1, 1])
    green_box.visual = TextureVisuals(
        material=PBRMaterial(
            baseColorFactor=np.array([0.2, 1.0, 0.2, 1.0], dtype=np.float32),
            metallicFactor=0.0,
            roughnessFactor=1.0,
            emissiveFactor=[0.0, 0.0, 0.0],
        )
    )
    green_box.apply_translation([1.5, 0, 0])

    both_boxes = trimesh.util.concatenate([red_box, green_box])

    return red_box, green_box, both_boxes
```

If we run this using a source install of trimesh, export, and visualize, here's what we get:

image

I spent some time investigating this. This PR contains some fixes:
- **Metallic-roughness texture channel order.** `metallic` is allocated to the blue channel of metallic-roughness textures [in the glTF2.0 spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material). It was instead being written to the red.
- **Gamma correction for `baseColorFactor`.** `baseColorFactor` was being copied directly into `baseColorTexture`. The units are different, though: the former is a linear scaling term, but `baseColorTexture` should be encoded with an sRGB transfer function. I copy-and-pasted a conversion function from `gloss.py`; these are local in both files.
- **UV indexing errors caused by deduplication.** Textures were being independently deduplicated during concatenation. It looks like this was causing alignment issues when some textures were unique (red vs green color in the example) but others were duplicated (roughness/metallic in the example are same for both meshes).
    - This part has a lot of moving parts so I added some tests, which fail on `main` but pass with the deduplication fix.

Here are the fixed meshes after these changes:

image

---

For completeness, the screenshots above were taken using trimesh's glTF export, loaded using GLTFLoader in `three.js` and rendered also via `three.js`. This is the code I used:
```python
import viser  # three.js-based 3D visualizer we're working on!

server = viser.ViserServer()
red_box, green_box, both_boxes = make_boxes()
server.scene.add_mesh_trimesh("/both", both_boxes, position=(0, 2, 0))
server.scene.add_mesh_trimesh("/box1", red_box)
server.scene.add_mesh_trimesh("/box2", green_box)
server.sleep_forever()
```

### DIFF
--- a/tests/test_pbr_concatenate.py
+++ b/tests/test_pbr_concatenate.py
@@ -1,0 +1,149 @@
+"""
+Test for PBR material concatenation bug fix.
+
+Tests that concatenating meshes with PBR materials preserves
+metallicFactor and roughnessFactor values and creates correct
+UV mappings for textures.
+"""
+
+try:
+    from . import generic as g
+except BaseException:
+    import generic as g
+
+
+class PBRConcatenateTest(g.unittest.TestCase):
+    def test_pbr_factors_preserved(self):
+        """
+        Test that metallicFactor and roughnessFactor are preserved
+        when concatenating meshes with PBR materials.
+
+        This tests the fix for the bug where these factors became None
+        after concatenation, causing meshes to appear shiny/metallic.
+        """
+        # Create two boxes with PBR materials
+        box1 = g.trimesh.creation.box(extents=[1, 1, 1])
+        mat1 = g.trimesh.visual.material.PBRMaterial(
+            baseColorFactor=[255, 51, 51, 255],  # Red
+            metallicFactor=0.0,
+            roughnessFactor=1.0,
+        )
+        box1.visual = g.trimesh.visual.TextureVisuals(material=mat1)
+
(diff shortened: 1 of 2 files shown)


mikedh:

PR #2472, opened 2025-10-18
Improve typing for Trimesh objects

MattTheCuber:
This PR adds a lot of missing typing to methods and properties in the `Trimesh` class. Additionally, this PR updates a bunch of docstrings that were missing information.

### DIFF
--- a/trimesh/base.py
+++ b/trimesh/base.py
@@ -1728,7 +1728,7 @@ class Trimesh(Geometry3D):
         """
         return cKDTree(self.vertices.view(np.ndarray))
 
-    def remove_degenerate_faces(self, height: float = tol.merge) -> None:
+    def remove_degenerate_faces(self, height: Floating = tol.merge) -> None:
         """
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome thanks for the fixes! 

One quick note, I'd recommend changing the `float, int` type hints to one of `trimesh.typing.(Number, Floating, Integer)` as they handle the "numpy int and float are sometimes but not always python int and float" issue:

```
In [1]: import numpy as np

In [2]: isinstance(np.zeros(10, dtype=np.int64)[0], int)
Out[2]: False

In [3]: isinstance(np.zeros(10, dtype=np.float64)[0], float)
Out[3]: True

In [4]: isinstance(np.zeros(10, dtype=np.float32)[0], float)
Out[4]: False
```

I thought using the `ExceptionModule` in type hints broke something (and we needed to do a forward-hint like `"Image"` or something), but it looks like it passes here so I guess it doesn't haha. With regards to the `BufferedRandom` does it pass if that's explicitly added to `Loadable`?

PR #2475, opened 2025-10-23
Add optional name parameter to min_distance_internal

markstor:
Hi again @mikedh! ^^

Here a PR that extends `CollisionManager.min_distance_internal()` with an optional `name` parameter to compute distances from a specific object to all others.

### Changes
- Added `name=None` parameter to `min_distance_internal()`
- When `name` provided: computes distance from specified object to others
- When `name=None`: existing behavior (all pairs) unchanged
- Temporarily removes target object from manager to avoid self-distance calculation
- Added test check for new functionality

### Usage
```python
# Existing (unchanged)
distance = manager.min_distance_internal()

# New functionality  
distance = manager.min_distance_internal(name="my_object")
```

### DIFF
--- a/tests/test_collision.py
+++ b/tests/test_collision.py
@@ -181,6 +181,10 @@ class CollisionTest(g.unittest.TestCase):
         assert g.np.isclose(dist, 2.0)
         assert names == ("cube2", "cube3")
 
+        dist, names = m.min_distance_internal(name="cube1", return_names=True)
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome thanks for the PR!! By any chance did you check the performance of `unregisterObject`/etc? Before I release I was going to check if that was "free-ish," I could imagine there's some world where that has to rebuild the scene.

PR #2481, opened 2025-11-08
[OBJ] Allow splitting scenes by 'o' and 'g' tags

### LINKED ISSUE #2413: Question: Trimesh OBJ loading, split on objects/groups instead of usemtl?
Currently it seems like when loading a mesh into trimesh, if I do not specify any usemtl statements it will load the entire OBJ file into a single mesh and deduplicate exactly overlapped vertices independently of the `o <object_name>` or `g <group_name>`. I was wondering if I had just missed a way to split an OBJ file on these statements instead of on usemtl, since I would like to split each object to be passed downstream to xatlas, but there are currently no materials assigned since there is no parameterization yet.

`trimesh.__version__ == 4.6.10`

Just for my own sanity, I'll list the steps I checked so far here:
I have read thru the help documentation on `trimesh.load_scene`.
I have read thru the documentation on `trimesh.exchange.obj.load_obj` and have tried setting `group_materials = False` and `maintain_order = True`. Neither fixed the issue
I have checked Github Issues, and saw 1 related issue to where a mesh was being returned as a scene (with separate meshes) when previously it was returned as 1 mesh. This is kind of the opposite issue, where I want a scene and am getting a mesh.

How I am currently handling this bug:
I am currently injecting unused `usemtl <incremented counter>` statements before each mesh. This works fine, but I think it would be more clear to provide an option to split on groups/objects. I also imagine for someone who is just using the library and is not as familiar with the OBJ format, it may be less clear why this is occurring.

Precedence for this option:
Blender provides an option for loading objs which split on object or group.

JulianKnodt:
Allow splitting obj files on 'o' and 'g' tags, where previously meshes could only be split by material before. This allows for an obj file with multiple objects and groups to be converted into a scene with multiple meshes.

Fixes #2413

### DIFF
--- a/tests/test_obj.py
+++ b/tests/test_obj.py
@@ -65,6 +65,18 @@ def test_obj_groups():
     # check to make sure there is signal not just zeros
     # assert g.np.ptp(mesh.metadata['face_groups']) > 0
 
+def test_obj_groups_split():
+    # a wavefront file with groups defined
+    mesh = g.get_mesh("groups.obj", split_groups=True)
+
+    # make sure the right number of meshes are split
+    assert len(mesh.geometry) == 14
+
+    mesh = g.get_mesh("groups.obj", split_objects=True, split_groups=True)
+
+    # make sure the right number of meshes are split
+    assert len(mesh.geometry) == 14
+
 
 def test_obj_negative_indices():
     # a wavefront file with negative indices
@@ -89,7 +101,7 @@ def test_obj_quad():
 
 def test_obj_multiobj():
     # test a wavefront file with multiple objects in the same file
-    scene = g.get_mesh("two_objects.obj", split_object=True, group_material=False)
+    scene = g.get_mesh("two_objects.obj", split_objects=True, group_material=False)
     assert len(scene.geometry) == 2
 
     for mesh in scene.geometry.values():
@@ -108,7 +120,7 @@ def test_obj_split_attributes():
     scene = g.get_mesh(
         "joined_tetrahedra.obj",
         process=False,
-        split_object=True,
+        split_objects=True,
         group_material=False,
     )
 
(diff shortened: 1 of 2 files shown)


JulianKnodt:
I was also hoping to add a unit test, but I have no idea how to run them. Are there docs on how to run them/add new ones?

mikedh:
Awesome thanks for the PR!! You should be able to run tests with just `pytest test_obj.py`, or  `python test_obj.py` and yeah adding a test before I merge would be great.

We should probably check performance with these options off as we are adding an additional `re.finditer` regardless of whether we're splitting on a group which can be quite slow in Python. Seems like we could lazily do the regex for both object and group which is probably worth it:

```
    idx_obj = np.array([m.start(0) for m in re.finditer("\no ", f_chunk)], dtype=int)
    idx_group = np.array([m.start(0) for m in re.finditer("\ng ", f_chunk)], dtype=int)

    # find all the indexes where we want to split
    splits = np.unique(np.concatenate(([0, len(f_chunk)], idx_mtl, idx_obj)))
    splits = np.unique(np.concatenate(([0, len(f_chunk)], idx_mtl, idx_obj, idx_group)))
```
Seems like it could lazily append those before the unique:
```
splits = [0, len(f_chunk)], idx_mtl]
if split_object:
   splits.append([m.start(0) for m in re.finditer("\no ", f_chunk)])
if split_group:
   splits.append([m.start(0) for m in re.finditer("\ng ", f_chunk)])

splits = np.unique(np.concatenate(splits, dtype=np.int64))
```

Also the tuple `key` is nice but we can probably in-line that as it's only called at the top of the function?

JulianKnodt:
I've changed what you suggested, tested it locally, but I still need to add a test before it's ready for merging

mikedh:
Yeah let's update the tests to explicitly pass `split_object=True` for now. I think the new behavior is a lot more explicit and for the majority of people who don't care about objects, faster. But yeah it does change the default behavior... I'll probably chicken out before release and change the default value to `split_object=True`. 

I think the last thing we need is a test, `models/groups.obj` has a bunch of groups (but no objects). 

Thanks for the follow up on this!

PR #2492, opened 2025-12-12
center of mass for flat meshes

lucagrementieri:
The current calculation of the center of mass returns always the origin for flat meshes. This PR replaces the origin with the centroid of the mesh that is the best estimate of the center of mass for an infinitely thin homogenous flat mesh.

For example, running the script
```python
import numpy as np
import trimesh

# Square [100,101] x [100,101] in the XY plane
vertices = np.array([[100, 100, 0], [101, 100, 0], [101, 101, 0], [100, 101, 0]])
faces = np.array([[0, 1, 2], [0, 2, 3]])
square_mesh = trimesh.Trimesh(vertices=vertices, faces=faces)
print("The current center of mass is", square_mesh.center_mass)
```
the result with the current code is
```>>> The current center of mass is [0. 0. 0.]```

while the update code gives
```>>> The current center of mass is [100.5 100.5   0. ]```

### DIFF
--- a/trimesh/triangles.py
+++ b/trimesh/triangles.py
@@ -292,8 +292,8 @@ def mass_properties(
 
     if center_mass is None:
         if np.abs(volume) < tol.zero:
-            # if there is no volume set center of mass to the origin
-            center_mass = np.zeros(3, dtype=np.float64)
+            # if there is no volume set center of mass to the centroid
+            center_mass = np.average(f1 / 3.0, weights=area(triangles, crosses), axis=0)
         else:
             # otherwise get it from the integration
             center_mass = integrated[1:4] / volume


mikedh:
Awesome thanks for the PR! Yeah I agree the current "just zero if it's broken" isn't great. I think in retrospect `mesh.center_mass` should probably be None when `not mesh.is_volume`, but I don't want to break existing code. 

The face-weighted average is already available as `mesh.centroid`, and I guess the original idea was that a user would run `mesh.center_mass if mesh.is_volume else mesh.centroid` but realistically that seems unlikely. 

Maybe, since every computed value in `triangles.mass_properties` is going to be wrong anyway if the mesh isn't watertight, we should just always return the integrated value and do a `log.debug` for the negative check (which actually isn't enough information to fully determine if it's valid since it doesn't have edge information). Not sure what the right answer here is.

PR #2500, opened 2026-01-25
Modify emissive factor validation range

WeakKnight:
Update emissive factor validation to allow values that are greater than one.


mikedh:
Thanks for the PR! Yeah I agree we should be less opinionated for values like this.

Issue #2504, opened 2026-02-23
intersects_id can return duplicate hits when mesh.ray._scale is very small


hoyeong-heo:
# Bug: Duplicate intersections due to too-small ray offset (`self._scale`)
I encountered an issue where `trimesh.ray.intersects_id` returns many duplicate intersection points when `multiple_hits=True`.

## Symptoms

- A single ray produces up to twenty duplicate hit locations.
- Increasing `max_hits` results in proportionally more identical points.
- All returned intersection points are effectively the same (within floating-point precision).

## Root cause (suspected)

The problem appears to be caused by an extremely small `self._scale` value, which makes the computed ray offset:

```python
distance = np.clip(_ray_offset_factor * self._scale, _ray_offset_floor, np.inf)
```

too small relative to the scene scale.

As a result:

- The updated ray origin (`new_origins + ray_offsets`) does not move far enough.
- The next Embree query hits the same triangle again.
- This repeats until `max_hits` is reached, producing duplicate intersections.

In short: ray origins are not effectively advancing between iterations.

## Expected behavior

Ray origins should advance enough to avoid immediately re-hitting the same triangle (or duplicates should be filtered).

## Environment

- `trimesh` version: `4.11.2`
- Backend: Embree
- OS: linux

## Reproduction

I will provide:

- Minimal test script
```py
import trimesh


mesh = trimesh.load_mesh("combined.stl")

x = -164.70464393956098
y = 275.5539451575655
z = 10000

origin = [[x, y, z]]
direction = [[0, 0, -1]]

# mesh.ray._scale_to_box = False

index_tri, index_ray, locations = mesh.ray.intersects_id(
    origin, direction, multiple_hits=True, return_locations=True
)

print(f"{index_tri=}")
print(f"{index_ray=}")
print(f"{locations=}")
```
- `combined.stl` mesh that reproduces the issue

[combined.zip](https://github.com/user-attachments/files/25479223/combined.zip)

## Notes

This seems related to scale-dependent robustness in the multi-hit ray loop.

Potential fixes might include:

- A configurable option for minimum offset
- Triangle ID re-hit prevention

AmenLemiesa:
Hey everyone! I’m a student working on an open-source project for a class, and since I use Trimesh in my research, I’d love to try fixing this.
I’m reproducing it with combined.stl and thinking of either tightening the minimum ray distance or blocking repeated triangle hits.
I’ll open a draft PR soon; let me know if there’s a preferred approach.

PR #2524, opened 2026-04-04
Fix duplicate ray hits in pyembree intersects_id (#2504)

AmenLemiesa:
When querying rays with multiple_hits=True at large coordinate scales the float32 precision of embree combined with a very small _scale caused the distance offset to be swallowed. Because the ray origin didn't actually advance, the loop repeatedly hit the exact same triangle until it reached max_hits.

I added a tracker array last_hit_triangle inside the max_hits loop in ray_pyembree.py. If a ray hits the exact same triangle two iterations in a row, the logic now correctly filters it out as a duplicate while allowing the "ray" to continue advancing.

### DIFF
--- a/tests/test_ray.py
+++ b/tests/test_ray.py
@@ -203,6 +203,30 @@ class RayTests(g.unittest.TestCase):
                 # top or bottom face
                 assert g.np.isclose(p[2], mesh.bounds[:, 2]).any()
 
+    def test_embree_duplicate_hits(self):
+        """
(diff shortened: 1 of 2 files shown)


mikedh:
Awesome thanks for taking a look at this!! Does this case come up reliably in any of the unit tests? Would you be able to add a test for it? We might also be able to combine the self-hit check with the other mask conditional?

AmenLemiesa:
This did not come up in the unit tests for ray! I added a test and combined the self-hit check with the other mask.

PR #2525, opened 2026-04-08
Drop support for Python 3.8 and 3.9 (EOL)

jorenham:
Python 3.9 is EOL since 2025-10-31, so there's no need to support it anymore, and can even be considered as bit of a security risk (especially Python 3.8). See https://devguide.python.org/versions/ for more info.

This drops support for Python 3.8 and 3.9. I also cleaned up the dead branches, and fixed all new ruff errors (pyupgrade, mostly).

---

For some reason I had to add an additional `tests.generic` import attempt in `tests/test_voxel.py` to get the tests to run locally. It'm not sure what's going on with that, and downgrading pytest or using a different `--import-mode` didn't seem to help. The change is backwards-compatible though, so I hope it's not an issue.

---

For context: my goal was actually to increase the type annotation coverage of trimesh (it's currently at 22.9%, see  https://jorenham.github.io/typestats/dashboard/report/#trimesh for details). But doing so would be a lot easier when we don't have to worry about Python<3.10 legacy typing stuff, so I figured I might as well drop those EOL versions first.

But assuming this gets merged, I'd actually first follow-up on this by cleaning up the legacy typing re-exports in `trimesh.typed`, i.e. replace `Union` and `Optional` with `|`, and use `list[_]` instead of `List[_]`, `dict[_, _]` instead of `Dict[_, _]`, etc.

---

Oh and in case it matters here; I didn't use any AI for this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,16 +30,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         os: [ubuntu-latest, macos-latest, macos-15-intel, windows-latest]
-        exclude:
-            # windows runners have gotten flaky
-          - os: windows-latest
-            python-version: 3.8
-          - os: windows-latest
-            python-version: 3.9
-          - os: macos-13
-            python-version: 3.9
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
     - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.13"]
+        python-version: ["3.10", "3.13"]
         os: [ubuntu-latest, windows-latest, macos-latest, macos-15-intel]
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Github Actions](https://github.com/mikedh/trimesh/workflows/Release%20Trimesh/badge.svg)](https://github.com/mikedh/trimesh/actions) [![codecov](https://codecov.io/gh/mikedh/trimesh/branch/main/graph/badge.svg?token=4PVRQXyl2h)](https://codecov.io/gh/mikedh/trimesh)  [![Docker Image Version (latest by date)](https://img.shields.io/docker/v/trimesh/trimesh?label=docker&sort=semver)](https://hub.docker.com/r/trimesh/trimesh/tags) [![PyPI version](https://badge.fury.io/py/trimesh.svg)](https://badge.fury.io/py/trimesh)
 
 
-Trimesh is a pure Python 3.8+ library for loading and using [triangular meshes](https://en.wikipedia.org/wiki/Triangle_mesh) with an emphasis on watertight surfaces. The goal of the library is to provide a full featured and well tested Trimesh object which allows for easy manipulation and analysis, in the style of the Polygon object in the [Shapely library](https://github.com/Toblerity/Shapely).
+Trimesh is a pure Python 3.10+ library for loading and using [triangular meshes](https://en.wikipedia.org/wiki/Triangle_mesh) with an emphasis on watertight surfaces. The goal of the library is to provide a full featured and well tested Trimesh object which allows for easy manipulation and analysis, in the style of the Polygon object in the [Shapely library](https://github.com/Toblerity/Shapely).
 
 The API is mostly stable, but this should not be relied on and is not guaranteed: install a specific version if you plan on deploying something using `trimesh`.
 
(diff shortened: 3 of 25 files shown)


mikedh:
Thanks for the PR and taking a look at this!! 

Changes look good to me, and yeah I'm open to dropping EOL Python, typically I've only done it when it becomes a blocker for something but union type hints, skipping the `typing.List` special case imports etc seem worth it. Checking download statistics, it looks like 2% of people are on 3.9 and 1% of downloads are on 3.8. I suppose setting the `requires-python` will mean they continue to get the as-of-now current version which seems graceful enough. 

> But assuming this gets merged, I'd actually first follow-up on this by cleaning up the legacy typing re-exports in trimesh.typed, i.e. replace Union and Optional with |, and use list[_] instead of List[_], dict[_, _] instead of Dict[_, _], etc.

That makes sense, and the `ruff` `UP` rules do this automatically iirc once we bump the `requires-python`?

> Oh and in case it matters here; I didn't use any AI for this.

Nice! Yeah it seems like one of the valuable things for libraries going forward is that humans have looked at every line, even the wrong ones haha. I had Claude on a branch do the type hints as aggressively as possible, and it did OK, the main issue it had was the soft dependencies (see: https://github.com/astral-sh/ruff/issues/7175). But yeah I think in general it's a bad idea to do large Claude PR's (maybe with the exception of finishing old zombie plumbing work like #2365). With regards to "soft" dependencies it really would be nice to use forward references (i.e. `def foo(graph: "networkx.DiGraph") -> float`) but I guess that's not on the table, and any ideas for that issue would be welcome!

I might release this as an `rc` series for a few months, and maybe finally bump the library from `alpha` to `beta`.

jorenham:
> That makes sense, and the `ruff` `UP` rules do this automatically iirc once we bump the `requires-python`?

Yes, exactly

> I had Claude on a branch do the type hints as aggressively as possible

In my experience, even the current state-of-the art LLM models are pretty bad when it comes to annotating Python code. My guess is that there's simply not a lot of training data out there, and that a large portion of that data is polluted by stubs that are mass produced by tools such as mypy's stubgen (which is only meant as a starting point, yet most people use it as an end product, grumble...).

> the main issue it had was the soft dependencies (see: [astral-sh/ruff#7175](https://github.com/astral-sh/ruff/issues/7175)).

Well, in case of `scipy`, I might have a solution to that. Because in the meantime *someone* (😏) has written stubs for SciPy, which goes by the creative name of `scipy-stubs` (https://github.com/scipy/scipy-stubs/). And a little known fact about `scipy-stubs` is that it doesn't require `scipy` to be installed ('tis but an optional dependency). So all you have to do is install `scipy-stubs` in the same env as you're running the type-checker in, and the type-checker will stop complaining about scipy, because it can just ask `scipy-stubs` what it *looks* like.

> But yeah I think in general it's a bad idea to do large Claude PR's

Agreed; It's way to much fun to write type annotations by hand :stuck_out_tongue_winking_eye:. 

> With regards to "soft" dependencies it really would be nice to use forward references (i.e. `def foo(graph: "networkx.DiGraph") -> float`) but I guess that's not on the table, and any ideas for that issue would be welcome!

Something like should work:

```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import networkx  # assumes types-networkx is installed

def foo(graph: "networkx.DiGraph") -> float: ...  # type-checkers: ☑️
```

mikedh:
Awesome sounds good, thanks again for looking at this! Let me know when you're ready to merge into a `beta` branch where we can experiment with which typing approach works without disrupting users, and do prerelease `rc` versions at some point.

Last time I looked in to expanding type coverage the soft dependency and `ExceptionWrapper` strategy was a bit of a challenge. One vague thought which may or may not be a good idea, was moving all wrapped soft-dependencies to `trimesh.typed` as a single-import location. This could keep the wrapping in one place and potentially let us put `pyi`'s next to them. And then module code would be something like `from ..typed import scipy` rather than the multiple `try` blocks. 

It also would be neat if polishing the soft-dependency import method let us try [`importlib.LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) from the standard library to decrease our cold-start time (~2s on my machine). Although their example code is somewhat suspicious and frightening haha. 

Anyway open to experimenting with different options!

jorenham:
> Awesome sounds good, thanks again for looking at this! Let me know when you're ready to merge into a `beta` branch where we can experiment with which typing approach works without disrupting users, and do prerelease `rc` versions at some point.

I guess that could also work. I'm guessing that the trimesh API is pretty stable at this point, so hopefully it won't be too much work to keep those branches in sync so we can avoid merge conflicts and stuff.

FWIW, in my time as NumPy maintainer I've seen quite a lot of experiments being pushed to the main branch. Sometimes a downstream library complains that we broke something in nightly build, and then we simply revert it. And in cases where it gets into a release, then we revert it in a patch release. Or in case someone really messes up (like I recently did with numpy 2.4.0 haha), then we yank that release, and even then I haven't seen anyone complain about it. 
So even in worst case for one of the most used Python libs, it's not as bad as you might imagine it would be. 
Long story short; personally I think it's fine in this case to just go directly to main. And if someone complains, then we can always revert. But if you're not comfortable with that, then I'm of course also fine with doing this in a beta branch, it'll just require a bit more work (e.g. resolving merge conflicts and keeping, keeping the branches in sync, maybe some CI adjustments, etc.). 

Also, in this case, I wouldn't even call it an experiment. Typing improvements like these are pretty common, and I have done this for quite a lot of similar open source projects (it's pretty much my job description).
But I guess that also means that I'm probably biased. So I can understand how from your standpoint this could look pretty experimental.

Anyway, I'll rebase this onto main, and then this should be ready. If you want to create a `beta` or `dev` branch or something, then we can simply switch this PR's target branch from `main` to that one.

> Last time I looked in to expanding type coverage the soft dependency and `ExceptionWrapper` strategy was a bit of a challenge. One vague thought which may or may not be a good idea, was moving all wrapped soft-dependencies to `trimesh.typed` as a single-import location. This could keep the wrapping in one place and potentially let us put `pyi`'s next to them. And then module code would be something like `from ..typed import scipy` rather than the multiple `try` blocks.

For untyped libraries like that we could write some local stubs (in a `typings/` directory typically). Those don't have to be bundled in the sdists or wheels or anything, as long as they're available when type-checking it'll work just fine. Or better yet, we could try and directly contribute to those packages [^1].

[^1]: I've been doing this quite a lot in the last 2 years, and in most cases those PRs were even accepted (not to pat myself on the back or anything, just saying that it's something that could work).

mikedh:
Awesome sorry for the delay on this. Yeah I agree this doesn't need to be in a branch for very long, I'm just going to use the EOL Python as a good break-point to apply some deprecation that are past their stated date haha. And maybe finish the 4 year old Pyglet 2 migration PR. 

> For untyped libraries like that we could write some local stubs (in a typings/ directory typically). Those don't have to be bundled in the sdists or wheels or anything, as long as they're available when type-checking it'll work just fine. Or better yet, we could try and directly contribute to those packages

Cool yeah the `typing.py` -> `typing/__init__.py` move is extremely low risk and has worked for other stuff (`gltf`) and it seems like it adds some flexibility here.

Thanks for the follow up on this and all the typing work!!

PR #2540, opened 2026-05-06
Modernized type annotations

jorenham:
As promised in #2525, here's the PR that modernizes the type annotations. Specifically, this replaces

- `typing.List` with `builtins.list`
- `typing.Dict` with `builtins.dict`
- `typing.Set` with `builtins.set`
- `typing.Tuple` with `builtins.tuple`
- `typing.Optional[T]` with `T  | None`
- `typing.Union[T1, T2]` with `T1 | T2`

This also explicitly marks type aliases with `typing.TypeAlias`, and removes the trivial and unused `typing` re-exports from `trimesh.typed`.
The `trimesh.typed.Stream` type alias was also simplified by exploiting the fact that the `typing.*IO` and `io.IO` types all have `typing.IO` as base class.


jorenham:
The failed docker test CI job seems unrelated (502 error)

PR #2541, opened 2026-05-12
`trimesh.util`: fix typing errors and add missing annotations

jorenham:
This fixes most typing errors in `trimesh.util` (checked with pyrefly and basedpyright), and increases type coverage of this module to 100%, which in turn increases the total type coverage of `trimesh` by +3.38% (according to [typestats](https://jorenham.github.io/typestats/dashboard/report/#trimesh).

In one or two locations I made minor changes to the implementation to help type-checkers understand what's going on. These should not lead to different behavior, or at least, not when the functions in questiona re used correctly, that is. 

This also removes a `numpy.linalg.multidot`  implementation that was no longer needed, because the of the `numpy>=1.21`  requirement (see the relevant [numpy 1.21 docs](https://numpy.org/doc/1.21/reference/generated/numpy.linalg.multi_dot.html)).

### DIFF
--- a/trimesh/util.py
+++ b/trimesh/util.py
@@ -13,19 +13,82 @@ import time
 import uuid
 import warnings
 import zipfile
-from collections.abc import Mapping
+from collections.abc import (
+    Callable,
+    Collection,
+    Hashable,
+    Iterator,
+    Mapping,
+    MutableSet,
+    Sequence,
+)
 from copy import deepcopy
 from io import BytesIO, StringIO
+from typing import (
+    IO,
+    TYPE_CHECKING,
+    Any,
+    Final,
+    Literal,
+    TypeAlias,
+    TypedDict,
+    TypeGuard,
+    TypeVar,
+    cast,
+    overload,
+)
+
+if TYPE_CHECKING:
+    import rtree
+    from _typeshed import SizedBuffer, SupportsLenAndGetItem, SupportsRead, SupportsWrite
+    from typing_extensions import TypeIs
+
+    from trimesh.base import Trimesh
+    from trimesh.path import Path2D, Path3D
+
 
 import numpy as np
 
 from .iteration import chain
 
 # use our wrapped types for wider version compatibility
-from .typed import ArrayLike, Integer, Iterable, NDArray, float64
+from .typed import ArrayLike, Integer, Iterable, NDArray
+
+# type variables (not to be used outside of this module)
+_T = TypeVar("_T")
+_KT = TypeVar("_KT", bound=Hashable)
+_NonSequenceT = TypeVar("_NonSequenceT", bound=str | bytes | complex)
+_StrOrBytes = TypeVar("_StrOrBytes", bound=str | bytes)
+_ArrayT = TypeVar("_ArrayT", bound=NDArray[Any])
+_ShapeT = TypeVar("_ShapeT", bound=tuple[int, ...])
+_ScalarT = TypeVar("_ScalarT", bound=np.generic)
+_DTypeT = TypeVar("_DTypeT", bound=np.dtype[Any])
+_IntDTypeT = TypeVar("_IntDTypeT", bound=np.dtype[np.integer[Any]])
+
+# shape-typed ndarray aliases
+_VecI64: TypeAlias = np.ndarray[tuple[int], np.dtype[np.int64]]
+_VecF64: TypeAlias = np.ndarray[tuple[int], np.dtype[np.float64]]
+_MatI64: TypeAlias = np.ndarray[tuple[int, int], np.dtype[np.int64]]
+_MatF64: TypeAlias = np.ndarray[tuple[int, int], np.dtype[np.float64]]
+
+# typed mappings for `encode_array`
+
+
+class _EncodedArray(TypedDict):
+    dtype: str
+    shape: tuple[int, ...]
+
+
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR!! 

I know it's going to substantially reduce the amount of possible type hint coverage we can add in the library, but I'd like to at least for now, avoid introducing a few of the more involved patterns like `if TYPE_CHECKING`, `overload`, and not introduce any stub or phantom (`_typeshed`) imports, and see how far we can get it with the `from .typed import Hint` and basic annotations in functions. 

It's pretty arbitrary but outside of that scope it's kind of a pain to edit in emacs for me, particularly the overload stuff. I'll play with it in a branch though, few quick questions:

- ` def is_file(obj: object) -> TypeGuard["SupportsRead[Any] | SupportsWrite[Any]"]:`
  - is there any case `-> bool` doesn't cover here?
  - why add an `object` type vs an `Any`, or no hint?
- the numpy types like `def vector_to_spherical(cartesian: ArrayLike) -> _MatF64:` why not just make the return type an `NDArray[float64]`?
- why not define the numpy types we need in `typed.py` and then re-use them in util?

jorenham:
> Thanks for the PR!!
> 
> I know it's going to substantially reduce the amount of possible type hint coverage we can add in the library, but I'd like to at least for now, avoid introducing a few of the more involved patterns like `if TYPE_CHECKING`, `overload`, and not introduce any stub or phantom (`_typeshed`) imports, and see how far we can get it with the `from .typed import Hint` and basic annotations in functions.

Not being able to use `if TYPE_CHECKING` means that we won't be able to use any of the typing features that have been introduced in later Python version using `typing_extensions` (unless we make `typing_extensions` a required dependency, that is). 

Not being able to use `overload` means that for functions with params such as `return_spam=False`, where the return type directly depends on a typable input parameter, are forced to return unions. This can be very annoying for downstream users to have to deal with, because type-checkers will require that you'll then handle each possible return type manually, even though it's already obvious (to the user) which type will be returned. That pain can be avoided by using some tactical `@overload`s here and there. 
But I fully agree that overloads are verbose and can be annoying to maintain. Using them sparingly sounds very reasonable to me. But I don't think that completely banning them is what's best for the users.

Not being able to use `_typeshed` should be no problem. Note, however, that this means that **we** will like have to redefine protocols that are already available in `_typeshed`. So I think that banning `_typeshed` will make it achieve the opposite of what you're going for; keeping the codebase nice and maintainable. 

> It's pretty arbitrary but outside of that scope it's kind of a pain to edit in emacs for me, particularly the overload stuff.

I had the feeling that the trimesh API is pretty stable at this point. So hopefully there's no *need* for you to have to touch them at all. And in case they actually have to be updated, just ping me and I'll do my best to help out (no hard promises though).

>  I'll play with it in a branch though, few quick questions:
> 
> * ` def is_file(obj: object) -> TypeGuard["SupportsRead[Any] | SupportsWrite[Any]"]:`
>   
>   * is there any case `-> bool` doesn't cover here?

The `TypeGuard` ensures that within a `if is_file(obj): ` branch type-checkers know that `obj` is of the specified type. In this case, that means that `obj` either has a `.read` or `.write` method. So without the `TypeGuard`, there will be no *type narrowing*. This often causes type checkers to complain if, within the `if is_file(obj):` pass `obj` to a function such as `def f(obj: SupportsRead[Any] | SupportsWrite[Any])`. This will often lead to situations where downstream users are forced to "state the obvious" by having to manually `typing.cast` the `obj` to the narrowed type themselves. Put differently: `TypeGuard` automatically `cast`s the value to the specified type if it evaluates to `True`.

>   * why add an `object` type vs an `Any`, or no hint?

`Any` is intrinsically type-unsafe, because `Any` be assigned to anything and everything is assignable to `Any`. The `object` type is not type-unsafe: `object` can **only** be assigned to `object` (and `Any`), not to e.g. `int`. 
So the general rule of thumb is to avoid `Any` if possible. So in case of input parameters that can accept anything, prefer `object` over `Any`, so that it's less likely to introduce bugs.

> * the numpy types like `def vector_to_spherical(cartesian: ArrayLike) -> _MatF64:` why not just make the return type an `NDArray[float64]`?

The shape of `NDArray[float64]` arrays is unknown, and behaves similar to `Any` (meaning it's type-unsafe). The `_MatF64` type includes the shape-type, `tuple[int, int]`, representing 2-d arrays. By being explicit about the type of the shapes, you give type-checkers additional information that they can (but don't have to) use to prevent users from making mistakes, such passing a 2-d array to a function that requires 1-d input.
Shapes of array can matter quite a bit at runtime, so it's almost always a good idea to explicitly specify these shape-types when you can. 
Last year I was working on shape-typing pretty much full time (through the NumPy fellowship program), so I could talk for hours about this haha. I guess that probably makes me look biased here, but the demand for shape-typing is very high (see e.g. https://github.com/numpy/numpy/issues/16544), which is why I was working on it in the first place.
Anyway, not trying to push this onto you or anything. I wouldn't mind going with plain old `NDArray`s if you really want me to; it's your library after all.

> * why not define the numpy types we need in `typed.py` and then re-use them in util?

That's not a bad idea. But won't that make it part of the public API? That's at least why I didn't do it before; since I didn't want to increase your API surface with my esoteric typing stuff. Especially if your not sure about if you actually want to keep using them.

But if you're indeed open to it, then I wouldn't mind deduping those type definitions by moving them into `typed.py`.

PR #2546, opened 2026-05-27
Fix spurious twist in sweep_polygon along curved 3D paths

### LINKED ISSUE #2448: sweep_polygon creates twisted/self-intersecting mesh in certain cases
**Describe the bug**
When using `trimesh.creation.sweep_polygon` to extrude a polygon along a path, the resulting mesh sometimes contains twisted or self-intersecting faces. This leads to invalid or distorted geometry.

---

**Code to Reproduce**

```python
import numpy as np
import trimesh
from shapely.geometry import Polygon

def _regular_ngon_2d(n_sides: int, radius: float, start_angle: float = 0.0) -> np.ndarray:
    angles = np.linspace(0.0 + start_angle, 2*np.pi + start_angle, n_sides, endpoint=False)
    x = radius * np.cos(angles)
    y = radius * np.sin(angles)
    return np.column_stack([x, y])  # (N,2)

polygon_2d = _regular_ngon_2d(n_sides=5, radius=0.1)

polygon_2d = Polygon(polygon_2d)  # Convert numpy array to shapely Polygon

polyline = np.array([
    [-0.4844,  0.541 ,  0.0815],
    [-0.6675,  0.811 ,  0.4648],
    [-0.5557,  0.9077,  0.627 ],
    [-0.3574,  0.977 ,  0.723 ],
    [-0.1116,  1.001 ,  0.7   ],
    [ 0.1598,  0.973 ,  0.5425],
    [ 0.2084,  0.676 ,  0.1472]]
)

mesh = trimesh.creation.sweep_polygon(polygon_2d, path=polyline, cap=True, connect=False)

mesh.export('sweep_test.obj')
```

---

**Expected behavior**
The mesh should follow the sweep path smoothly, with faces aligned, and without twisted or self-intersecting geometry.

---

**Actual behavior**

* The generated mesh is visibly twisted and contains distorted faces.
* See attached screenshot for visualization:

Image

---

**Environment (please complete the following):**

* OS: Ubuntu 22.04
* Python version: 3.10
* trimesh version: 4.7.4

---

**Additional context**
It looks like the frame/orientation computation along the sweep path might not be consistent, which causes polygons to flip or rotate unexpectedly.

gaoflow:
## Summary

Fixes #2448: `trimesh.creation.sweep_polygon` produced a mesh whose cross section spuriously **twists** about the path when sweeping along a curved 3D path, even with no per-vertex `angles` requested — leading to distorted / self-intersecting geometry.

## Root cause

Each slice's coordinate frame was built **independently** from its plane normal via spherical coordinates:

```python
theta, phi = util.vector_to_spherical(normal).T
# ... frame assembled from (theta, phi, roll) ...
```

This ties the in-plane roll of the frame to the normal's azimuth `theta`. As the path curves out of a plane (so the normal's polar angle `phi` also changes), the spherical parametrization rotates the cross section about the path — a classic non-rotation-minimizing frame. (Note: for a *planar* path `phi` is constant, so the twist doesn't appear there, which is why simple in-plane test paths looked fine.)

## Fix

Replace the per-slice spherical frame with a **rotation-minimizing (parallel-transport) frame**: each slice's in-plane axes are propagated from the previous slice by the *minimal* rotation that carries the previous normal onto the current one (Rodrigues' rotation about `normal[i-1] × normal[i]`). This is the standard way to sweep a profile without introducing twist.

- The starting in-plane axis is initialized to match the previous frame's **slice 0**, so a straight / untwisted sweep is unchanged.
- The optional per-vertex `angles` roll is applied on top of the propagated frame, with the same sign convention as before.
- The frame still maps local `+Z` onto each slice normal, so the existing `tol.strict` assertions and face winding are preserved.

## Verification

A direct, implementation-independent twist metric on the `#2448` repro path (rotate each slice's boundary ring by the minimal rotation to the next slice's normal, then measure the residual rotation about that normal):

| | total inter-slice twist |
|---|---|
| before | **138.6°** |
| after | **0.0°** |

The output mesh remains watertight, winding-consistent, and a valid volume; existing open and closed sweeps (`tests/test_creation.py::test_path_sweep`) still pass, including under `trimesh.tol.strict = True`.

Adds `tests/test_creation.py::test_sweep_no_twist`, which asserts the residual twist along a curving 3D path is `< 1°` (fails on `main` at ~139°, passes with this change). Full `test_creation.py`: 18 passed. `ruff check` / `ruff format` clean on both changed files.

Closes #2448.

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -175,6 +175,72 @@ def test_path_sweep():
         assert mesh.is_volume
 
 
+def test_sweep_no_twist():
+    # A polygon swept along a curving 3D path should not spuriously twist about
+    # the path when no per-vertex `angles` are requested.
+    # https://github.com/mikedh/trimesh/issues/2448
+    if len(existing_engines) == 0:
+        return
+
+    # a regular pentagon centered on the origin, so each slice's centroid is the
+    # corresponding path vertex (used by the twist metric below)
+    a = g.np.linspace(0, 2 * g.np.pi, 5, endpoint=False)
+    poly = g.Polygon(g.np.column_stack([0.1 * g.np.cos(a), 0.1 * g.np.sin(a)]))
+
+    # a path that curves in all three dimensions
+    path = g.np.array(
+        [
+            [-0.4844, 0.5410, 0.0815],
+            [-0.6675, 0.8110, 0.4648],
+            [-0.5557, 0.9077, 0.6270],
+            [-0.3574, 0.9770, 0.7230],
+            [-0.1116, 1.0010, 0.7000],
+            [0.1598, 0.9730, 0.5425],
+            [0.2084, 0.6760, 0.1472],
+        ]
+    )
+
+    mesh = g.trimesh.creation.sweep_polygon(poly, path, cap=True, connect=False)
+    assert mesh.is_volume
+
+    # plane normal of each slice, reconstructed exactly as `sweep_polygon` does
+    util = g.trimesh.util
+    tf = g.trimesh.transformations
+    vector = util.unitize(path[1:] - path[:-1])
+    vector_mean = util.unitize(vector[1:] + vector[:-1])
+    normal = g.np.concatenate([[vector[0]], vector_mean, [vector[-1]]], axis=0)
+
+    # vertices come out as one `stride`-length boundary ring per path vertex
+    stride = len(mesh.vertices) // len(path)
+    rings = mesh.vertices.reshape((len(path), stride, 3)) - path[:, None, :]
+
+    # For each consecutive pair of slices, rotate the earlier ring by the minimal
+    # rotation between their plane normals, then measure the residual rotation about
+    # the later normal. That residual is the spurious twist, which should be ~0 for
+    # a rotation-minimizing frame. Before the fix this exceeded 130 degrees.
+    total_twist = 0.0
+    for i in range(len(path) - 1):
+        axis = g.np.cross(normal[i], normal[i + 1])
+        sin_a = g.np.linalg.norm(axis)
+        cos_a = g.np.dot(normal[i], normal[i + 1])
+        if sin_a < 1e-12:
+            rotated = rings[i]
+        else:
+            matrix = tf.rotation_matrix(g.np.arctan2(sin_a, cos_a), axis / sin_a)
+            rotated = rings[i] @ matrix[:3, :3].T
+        a0 = rotated[0] / g.np.linalg.norm(rotated[0])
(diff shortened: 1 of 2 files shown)


mikedh:
Hey, yeah I'd rather not fully rewrite the sweep logic as the original is fully vectorized and heavily benchmarked and tested. If you can re-state this issue as a very simple 10 line test, we might be able to make a smaller one-line-ish tweak to the existing sweep logic.

binggao1230:
Thanks for taking a look — and that's fair, keeping the vectorized/benchmarked sweep intact makes sense.

Here's a minimal repro that quantifies the issue (a measurement, not a rewrite). It sweeps a thin rectangle along a helix and adds up how much the cross-section rotates *about the path tangent* from slice to slice. A rotation-minimizing frame would give ~0°; on `main` it accumulates ~102°:

```python
import numpy as np, trimesh
from shapely.geometry import Polygon

rect = Polygon([(-.5, -.02), (.5, -.02), (.5, .02), (-.5, .02)])
t = np.linspace(0, 2 * np.pi, 40)
path = np.c_[np.cos(t), np.sin(t), 0.3 * t]            # helix (curves out of any plane)
V = trimesh.creation.sweep_polygon(rect, path).vertices.reshape(len(t), 4, 3)
axis = V[:, 1] - V[:, 0]                               # the cross-section's long edge
tan = np.gradient(path, axis=0)
twist = 0.0
for i in range(1, len(t)):
    n = tan[i]                                         # compare both in the SAME normal plane
    a0 = axis[i - 1] - n * (axis[i - 1] @ n) / (n @ n)
    a1 = axis[i] - n * (axis[i] @ n) / (n @ n)
    a0 /= np.linalg.norm(a0); a1 /= np.linalg.norm(a1)
    twist += np.degrees(np.arccos(np.clip(a0 @ a1, -1, 1)))
print(round(twist, 1))   # main: ~102.0  (should be ~0)
```

(`m.show()` on that mesh shows the same thing — the flat ribbon corkscrews.) A purely planar path has constant azimuth so the twist is ~0, which is why the existing `test_path_sweep` doesn't catch it.

Root cause is narrow: each slice's in-plane frame is built from the path normal's spherical coordinates, which ties the cross-section's roll to the normal's azimuth — so as soon as the path curves out of a single plane, the section spins.

If you're open to something much smaller than my PR: the minimal in-place tweak is to keep everything as-is but derive each slice's in-plane axes from the previous slice by the minimal rotation between consecutive tangents (parallel transport), instead of recomputing them from the spherical angles each step. It leaves the vectorized structure and the `angles`/roll handling untouched and only changes how the per-slice basis is seeded. Happy to put that up as a focused diff if you'd like.

PR #2547, opened 2026-05-27
Make subdivide_to_size crack-free (conformal refinement), fixes #1731

### LINKED ISSUE #1731: subdivide_to_size() does not split shared edges
**Summary**

subdivide_to_size() does not appear to check if an edge being split is shared by another triangle; meaning you can end up with a vertex that, if subsequently moved, will result in tears/holes in a mesh.

This may be related to the "Fix trimesh.remesh.subdivide subdivide_to_size so that if you specify a subset of faces it splits their neighbors in a way to maintain watertightness." comment in https://github.com/mikedh/trimesh/issues/1557

**Details**

Using  Python 3.9.2 and Trimesh 3.15.8 I'm loading a simple STL into a mesh, then subdividing it to make no edge longer than 5 units:

```
mesh = trimesh.load('part.stl')
new_v, new_f = trimesh.remesh.subdivide_to_size(mesh.vertices, mesh.faces, 5.0)
mesh = trimesh.Trimesh(vertices=new_v, faces=new_f)
```

Using the viewer I can check the original mesh (without subdivision) in both shaded and wireframe (the first and second images in the attached jpg). The third image shows the same mesh after the subdivide_to_size() call above, and the fourth is a close up of that mesh.

The circled areas show where triangles have been split, but the formerly shared edge with another triangle hasn't. If these vertices are then moved they will leave a hole in the mesh.

Is there any way to get trimesh to subdivide a mesh and ensure that edges shared with other triangles are also split?

trimesh

gaoflow:
## Overview

Fixes #1731: `subdivide_to_size` does not split shared edges, so it turns a
watertight mesh into a non-watertight "triangle soup" (T-junctions / cracks).

## Reproduce

```python
import trimesh
m = trimesh.creation.box(extents=[10.0, 1.0, 1.0])   # watertight, 12 faces
v, f = trimesh.remesh.subdivide_to_size(m.vertices, m.faces, max_edge=1.0)
s = trimesh.Trimesh(vertices=v, faces=f, process=False)
print(s.is_watertight)   # main: False (144 boundary/T-junction edges, 2064 faces)
```

## Root cause

`subdivide_to_size` only subdivided the faces that had a too-long edge, by calling
`subdivide(face_index=too_long)` which does a 1→4 split of just those faces:

```python
too_long = (edge_length > max_edge).any(axis=1)
...
current_vertices, current_faces = subdivide(current_vertices, current_faces[too_long])
```

When a refined face shares an edge with a neighbor that is *not* refined, only the
refined side gets the edge's midpoint vertex. The neighbor keeps the original long
edge, so the two faces no longer line up along that edge — a T-junction. The old
docstring acknowledged this ("Will return a triangle soup, not a nicely structured
mesh"). As @mikedh noted in the issue, the only crack-free option previously was
`subdivide` on *every* face, which over-refines.

## Fix — edge-based conformal refinement

Refine on **edges**, not faces. Each pass:

1. Find the unique edges longer than `max_edge` and create **one shared midpoint
   vertex per such edge**, so the two faces adjacent to an edge always agree on its
   midpoint.
2. Split every face with a template chosen by how many of its three edges are being
   bisected:
   - **0** → keep the face unchanged (already small),
   - **1** → bisect from the midpoint to the opposite vertex (2 triangles),
   - **2** → a corner triangle plus a quad split along its **shorter diagonal** (3 triangles),
   - **3** → the usual 1→4 split.
3. Iterate until no edge exceeds `max_edge`.

Because each shared edge is bisected identically from both sides, no T-junctions are
introduced: a watertight input stays watertight. Faces that are already small enough
are left untouched, so it stays adaptive (no global over-refinement). For non-uniform
meshes it also produces **fewer** triangles than the old soup — the 10×1×1 box above
becomes **656 watertight faces** instead of 2064 cracked ones.

The public signature is unchanged: `subdivide_to_size(vertices, faces, max_edge,
max_iter=10, return_index=False)` still returns `(vertices, faces)` or
`(vertices, faces, index)`, with the same `max_iter` cap (`ValueError` on overflow).

## Verification

Added `tests/test_remesh.py::SubDivideTest::test_subdivide_to_size_crack_free`, which
on meshes with non-uniform edge lengths (long box, cylinder, icosphere) asserts:

- `is_watertight` and `is_winding_consistent`,
- no T-junctions (`len(edges) == len(edges_unique) * 2`),
- the `max_edge` target is met,
- area and volume are exactly preserved (subdivision is linear, so new vertices lie
  on the original surface),
- `return_index` gives one parent per new face and each new face lies on its parent.

This test **fails on the previous implementation** (the box becomes non-watertight)
and passes with this change. The full `test_remesh.py` suite passes, as do
`test_creation.py`, `test_smoothing.py`, `test_voxel.py`, and `test_ray_upstream.py`
(the `voxel.creation` / `Trimesh.subdivide_to_size` callers). `ruff check` and
`ruff format` are clean.

Fixes #1731.

### DIFF
--- a/tests/test_remesh.py
+++ b/tests/test_remesh.py
@@ -241,6 +241,50 @@ class SubDivideTest(g.unittest.TestCase):
         r = m.subdivide_to_size(0.01, max_iter=10)
         assert r.is_watertight
 
+    def test_subdivide_to_size_crack_free(self):
+        # subdivide_to_size must split edges shared between a refined and an
+        # unrefined face on both sides, so a watertight mesh stays watertight
+        # (no T-junctions). This only shows up when faces have non-uniform edge
+        # lengths so that some faces are refined and their neighbors are not:
+        # an axis-aligned box that is long in one dimension is the simplest case.
+        # See https://github.com/mikedh/trimesh/issues/1731
+        meshes = [
+            g.trimesh.creation.box(extents=[10.0, 1.0, 1.0]),
+            g.trimesh.creation.cylinder(radius=1.0, height=8.0),
+            g.trimesh.creation.icosphere(subdivisions=1),
+        ]
+        for m in meshes:
+            assert m.is_watertight
+            max_edge = 0.5
+            v, f, idx = g.trimesh.remesh.subdivide_to_size(
+                m.vertices, m.faces, max_edge=max_edge, return_index=True
+            )
+            s = g.trimesh.Trimesh(vertices=v, faces=f, process=False)
+
+            # the whole point of the fix: no cracks are introduced
+            assert s.is_watertight
+            assert s.is_winding_consistent
+            # in a watertight mesh every edge is shared by exactly two faces;
+            # a T-junction would leave an edge used by only one face
+            assert len(s.edges) == (len(s.edges_unique) * 2)
+
+            # the size goal is still met and geometry is unchanged
+            edge_len = (
+                g.np.diff(s.vertices[s.edges_unique], axis=1).reshape((-1, 3)) ** 2
+            ).sum(axis=1) ** 0.5
+            assert (edge_len <= max_edge + 1e-8).all()
+            assert g.np.isclose(m.area, s.area)
+            assert g.np.isclose(m.volume, s.volume)
+
+            # return_index: one parent per new face, lying on that original face
+            assert len(idx) == len(f)
+            assert idx.max() < len(m.faces)
+            bary = g.trimesh.triangles.points_to_barycentric(
+                m.triangles[idx], s.triangles.mean(axis=1)
+            )
+            assert bary.max() < 1 + 1e-3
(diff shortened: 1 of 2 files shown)


mikedh:
Hey, I don't think I can merge this algorithm rewrite without going through it at least semi-manually and maxing out on test coverage. I had Claude try an alternative fix in [this branch](https://github.com/mikedh/trimesh/compare/main...subdivide/fix) which does benchmark better than this PR. But I'm also not gonna merge that without going through carefully either haha. The bare minimum would be timings on every file in `models` with `faces=None`, and a a deterministic-but-random-subset of faces, asserts for `is_volume` against the solution, and a table of the runtime `timeit`x3 for each model and face subsets. If you want to branch off my vectorized thing and start on the proof framework that would be great!

PR #2551, opened 2026-05-28
base: drop faces touching non-finite vertices in remove_infinite_values (fixes #2445)

### LINKED ISSUE #2445: `numpy.nan` values prevent `show`ing mesh.
I'm new to `trimesh` so let me know if I'm missing something (also, thanks so much for the great library).

If I create a [`trimesh.Trimesh`](https://trimesh.org/trimesh.html#trimesh.Trimesh) where a vertex's `z` component is [`numpy.nan`](https://numpy.org/doc/stable/reference/constants.html#numpy.nan) the mesh is created, but `mesh.show()` errors with an `IndexError` at the vertex with `numpy.nan`. If I set `validate=True` `mesh.show()` displays, but only shows a black screen.

I would expect with `process=True` and/or `validate=True` (I still don't fully understand the difference) that `trimesh.Trimesh` would handle vertices with `numpy.nan` values by removing them and the associated faces from the mesh.

gaoflow:
Fixes #2445 — meshes with NaN/Inf vertices now have their dependent faces dropped during processing, instead of being silently rewritten to degenerate triangles that crash the renderer.

## The bug

Reported in #2445:

> If I create a `trimesh.Trimesh` where a vertex's `z` component is `numpy.nan` the mesh is created, but `mesh.show()` errors with an `IndexError` at the vertex with `numpy.nan`.

```python
import numpy as np, trimesh
vertices = np.array([
    [0.0, 0.0, 0.0],   # 0
    [1.0, 0.0, 0.0],   # 1
    [0.5, 0.5, np.nan],# 2 -- bad
])
faces = np.array([[0, 1, 2]])
m = trimesh.Trimesh(vertices=vertices, faces=faces, process=True)
print(m.vertices)  # [[0,0,0],[1,0,0]]  -- NaN removed, OK
print(m.faces)     # [[0,1,0]]          -- ⚠ now degenerate
```

The NaN vertex is removed, but the face that referenced it is rewritten to `[0, 1, 0]` — a degenerate triangle reusing vertex 0. Downstream that crashes `mesh.show()` and corrupts every face-adjacency / volume / `is_watertight` calculation.

## Root cause

`Trimesh.remove_infinite_values` (in `trimesh/base.py`) had two lines:

```python
if util.is_shape(self.faces, (-1, 3)):
    face_mask = np.isfinite(self.faces).all(axis=1)   # ← always all-True
    self.update_faces(face_mask)
if util.is_shape(self.vertices, (-1, 3)):
    vertex_mask = np.isfinite(self.vertices).all(axis=1)
    self.update_vertices(vertex_mask)
```

`self.faces` is `int64`, so `np.isfinite(self.faces)` is always all-True and `face_mask` removes nothing. The faces that referenced a non-finite vertex then fall through to `update_vertices`, whose inverse-index map (`base.py:1224`) is `np.zeros(len(vertices), dtype=int64)` — the entries for removed vertices stay zero, so every face index pointing at a NaN/Inf vertex gets remapped to `0`.

Net effect: `[0,1,2]` with vertex 2 non-finite becomes `[0,1,0]`, not a removal.

## The fix

Compute the vertex-finite mask first, drop faces touching a non-finite vertex BEFORE removing the vertex, then run the usual vertex update:

```python
def remove_infinite_values(self) -> None:
    if util.is_shape(self.vertices, (-1, 3)):
        vertex_mask = np.isfinite(self.vertices).all(axis=1)
        if (
            util.is_shape(self.faces, (-1, 3))
            and not vertex_mask.all()
            and len(self.faces) > 0
        ):
            # face is kept only if all three vertex indices are valid
            face_mask = vertex_mask[self.faces].all(axis=1)
            if not face_mask.all():
                self.update_faces(face_mask)
        self.update_vertices(vertex_mask)
```

Why this is safe:

- The previous "face" mask was a no-op, so removing that block changes nothing on its own.
- The new face mask fires only when at least one vertex is non-finite — fully finite meshes hit `vertex_mask.all()` and skip the new branch entirely, preserving the existing fast path.
- Faces touching a non-finite vertex are dropped via `update_faces` (which carries face attributes / face normals correctly), so the subsequent `update_vertices` only re-indexes faces that point at survivors.
- The `update_vertices` inverse-array behavior is left alone — other callers in trimesh (post-merge re-indexing, etc.) rely on it.

## Verification

Three regression tests added to `tests/test_mesh.py`:

- `test_remove_infinite_values_drops_dependent_faces`: 5 verts (1 NaN) / 5 faces (3 touching NaN). After process: 4 finite vertices, **2** non-degenerate faces with valid indices (was 5 faces with degeneracies on main), and `mesh.scene()` builds without crashing.
- `test_remove_infinite_values_inf_vertex_drops_face`: same behavior for `np.inf`.
- `test_remove_infinite_values_no_op_on_finite_mesh`: random fully-finite mesh — vertex and face counts unchanged after `remove_infinite_values`.

```
$ pytest tests/test_mesh.py
5 passed
$ pytest tests/ --ignore=tests/test_gltf.py -k "not segmentation"
656 passed
$ pytest tests/test_gltf.py
50 passed
$ ruff check trimesh/base.py tests/test_mesh.py
All checks passed!
```

No regressions across the full unit suite.

### DIFF
--- a/tests/test_mesh.py
+++ b/tests/test_mesh.py
@@ -180,6 +180,100 @@ def test_mesh_2D():
     assert len(rend) > 1
 
 
+def test_remove_infinite_values_drops_dependent_faces():
+    # Regression test for https://github.com/mikedh/trimesh/issues/2445
+    #
+    # `remove_infinite_values` was checking `np.isfinite(self.faces)` which
+    # is always all-True (face indices are int64), so faces referencing a
+    # NaN/Inf vertex were never explicitly dropped. They then sneaked
+    # through `update_vertices`, whose inverse-index map defaults to 0
+    # for removed vertices — turning each face that touched a NaN vertex
+    # into a degenerate triangle (e.g. `[0,1,2]` -> `[0,1,0]`). Those
+    # degenerate faces crashed the renderer with `IndexError` on `mesh.show`.
+    vertices = g.np.array(
+        [
+            [0.0, 0.0, 0.0],
+            [1.0, 0.0, 0.0],
+            [0.0, 1.0, 0.0],
+            [1.0, 1.0, 0.0],
+            [0.5, 0.5, g.np.nan],
+        ],
+        dtype=g.np.float64,
+    )
+    faces = g.np.array(
+        [
+            [0, 1, 2],
+            [1, 3, 2],
+            [0, 4, 1],
+            [1, 4, 3],
+            [3, 4, 2],
+        ],
+        dtype=g.np.int64,
+    )
+    mesh = g.trimesh.Trimesh(
+        vertices=vertices.copy(), faces=faces.copy(), process=True
+    )
+
+    # the non-finite vertex is gone
+    assert g.np.isfinite(mesh.vertices).all()
+    assert len(mesh.vertices) == 4
+
+    # the three faces that touched the NaN vertex are removed; the two
+    # entirely-finite faces survive without being remapped to degenerates
+    assert len(mesh.faces) == 2
+    # every surviving face references a valid vertex
+    assert (mesh.faces < len(mesh.vertices)).all()
+    # no surviving face is degenerate (no repeated vertex index)
+    assert (mesh.faces[:, 0] != mesh.faces[:, 1]).all()
+    assert (mesh.faces[:, 1] != mesh.faces[:, 2]).all()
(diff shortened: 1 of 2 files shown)


mikedh:
Makes sense and asserts look good, thanks for the PR!

PR #2556, opened 2026-05-28
Resolve OBJ negative face indices against running vertex count (fixes #2364)

### LINKED ISSUE #2364: Bug: Loading OBJ with multiple objects with negative face indices and repeat materials
I have this wonky OBJ file with multiple repeat materials, negative face indices and multiple object groups, which trimesh loads incorrectly. With `trimesh.load_mesh("bug_report.obj")`, it reports the number of vertices as 24, but meshlab correctly gives 40 vertices.

This was tested on `trimesh.__version__ == 4.6.3`, but I also observed it on `4.5.1` and `3.2`.

This may be because vertices are mixed in with faces (example below), and if trimesh currently parses all vertices then all faces without maintaining current position then it will be incorrect:
```
v
v
v
f -1 -2 -3 (refers to 3 vertices above, not the 3 below)
v
v
v
f -1 -2 -3 (refers to 3 vertices above)
```

I think it's unlikely this bug was previously hit because the way I'm writing indices is incredibly weird (nearly all OBJ exporters batch all vertices), but most DCC tools handle the case fine. I'm creating strange OBJs like this because they don't require that I maintain any state about previous vertices or faces 😅 😅 and can output individual components

Image

[bug_report.obj.zip](https://github.com/user-attachments/files/18995845/bug_report.obj.zip)

I didn't include the MTL file for the obj, since the material is unrelated, but if you need it for some reason I can provide it.

gaoflow:
Fixes #2364.

The OBJ spec allows interleaving `v` / `vt` / `vn` and `f`
declarations.  A negative face index `-k` refers to the k-th most
recently declared element BEFORE that face line — each of
`v`, `vt`, `vn` counted in its own list.

trimesh's loader bulk-collects every `v` / `vt` / `vn` up-front via
`_parse_vertices` before processing faces, so the existing
`array[array > 0] -= 1` line in `load_obj` left negatives alone
and they were silently resolved by numpy's negative-indexing
against the file's *final* element list.  For the common "all v
then all f" layout the running count at each face line equals the
total count, so it worked by accident — but the converter that
produced the OBJ in #2364 emits each primitive as a
`v…v…f -1..-N` block, and the loader returned a 24-vertex tangle
where MeshLab returns 40 vertices and three clean solid primitives.

## Reproducer

The file from #2364 (`bug_report.obj`, 40 `v`, 26 `f`, all faces
use negative indices) loads as:

| | trimesh main | this PR | MeshLab |
|---|---|---|---|
| vertices | 24 | **40** | 40 |
| `is_volume` | False | **True** | (n/a) |
| bbox (z) | `[-0.31, 0.31]` (wrong) | **`[-0.37, 0.31]`** (correct) | — |

## Fix

Add `_resolve_negative_face_indices` (new module-private helper in
`trimesh/exchange/obj.py`) that walks the OBJ text once, tracks the
running count of `v` / `vt` / `vn` declarations, and rewrites every
negative face index into its absolute 1-based positive form on each
face line.

- Cheap regex pre-check (`(?m)^f[^\n]*-\d`) short-circuits the pass
  for files without negative face indices — there is no cost for
  the common case.
- On files that DO use negatives in the "normal" layout (all v then
  all f), the rewrite is a semantic no-op because running_count ==
  total_count at every face line.  The pre-existing
  `test_obj_negative_indices` test (which uses such a file) keeps
  passing unchanged.

## Tests

Three new tests in `tests/test_obj.py`:

- `test_obj_negative_indices_interleaved` — synthesizes two
  axis-aligned unit cubes declared in two interleaved
  `v…v…f -8..-1` blocks (same shape as the bug report).  Before
  the fix the loader produced a single 8-vertex tangle; after the
  fix the scene contains two distinct 8-vertex cubes that
  concatenate to a 16-vertex, volume-2 mesh.
- `test_obj_negative_indices_normalizer_noop_on_positive_file` —
  asserts that files whose face lines use only positive indices
  are returned bit-identical, even when vertex *coordinates* are
  negative.
- `test_obj_negative_indices_normalizer_mixed_slash_layouts` —
  covers all three token layouts: `v/vt/vn`, `v//vn`, `v/vt`.

Local `pytest tests/test_obj.py` 41/42 pass (the 1 failure
`test_multi_nodupe` requires `pycollada` which isn't installed in
this venv — unrelated).  Ruff clean.

### DIFF
--- a/tests/test_obj.py
+++ b/tests/test_obj.py
@@ -93,6 +93,139 @@ def test_obj_negative_indices():
     assert mesh.vertices.shape == (8, 3)
 
 
+def test_obj_negative_indices_interleaved():
+    # regression for issue #2364: OBJ files that interleave `v` and `f`
+    # declarations together with negative face indices were silently
+    # mis-loaded because the loader bulk-collected every `v` up-front
+    # so negative indices were resolved against the file's *final*
+    # vertex list rather than the running count at each face line.
+    #
+    # synthesize two axis-aligned unit cubes declared in two passes;
+    # each block uses `f -8 .. -1` referencing only the eight vertices
+    # that immediately precede it, like the OBJ in the bug report.
+
+    def cube_block(origin):
+        ox, oy, oz = origin
+        # 8 unit cube vertices in canonical order
+        verts = [
+            (ox, oy, oz),
+            (ox + 1, oy, oz),
+            (ox + 1, oy + 1, oz),
+            (ox, oy + 1, oz),
+            (ox, oy, oz + 1),
+            (ox + 1, oy, oz + 1),
+            (ox + 1, oy + 1, oz + 1),
+            (ox, oy + 1, oz + 1),
+        ]
+        lines = [f"v {x} {y} {z}" for x, y, z in verts]
+        # six quad faces, all using the negative offset -1..-8
+        # which under the spec refers to the 8 vertices just above.
+        # winding chosen for consistent outward normals.  the verts
+        # are indexed -8..-1 == v0..v7 with:
+        #   v0=(0,0,0), v1=(1,0,0), v2=(1,1,0), v3=(0,1,0)
+        #   v4=(0,0,1), v5=(1,0,1), v6=(1,1,1), v7=(0,1,1)
+        # (with `origin` added to each).
+        lines.extend([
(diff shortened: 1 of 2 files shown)


mikedh:
Hey, yeah I think we should probably just warn on negative indices and drop the faces, re-doing the hot parse loop is probably not worth this niche feature.

PR #2557, opened 2026-05-28
Round-trip MTL Ni/d/Tf/illum through SimpleMaterial (fixes #2349)

### LINKED ISSUE #2349: Incomplete material property export for wavefront (.obj) files
When loading wavefront (`.obj`) files with associated `.mtl` materials using `group_material = False` and `skip_materials = False` and exporting it afterwards, not all material properties from the `.mtl` file are exported.

Currently, only ambient (`Ka`), diffuse (`Kd`), specular (`Ks`), and glossiness (`Ns`) properties are supported and exported by `SimpleMaterial`. Other properties, such as the index of refraction (`Ni`), are parsed and stored in [`SimpleMaterial.kwargs`](https://github.com/mikedh/trimesh/blob/4.6.0/trimesh/visual/material.py#L87) but not exported by `SimpleMaterial.to_obj()`. Furthermore, the hash generation for unique material names only considers the exported properties, potentially leading to naming conflicts for materials differing by unexported properties.

The index of refraction is officially supported and even used in some of the models [/models/box.mtl](/mikedh/trimesh/blob/4.6.0/models/box.mtl). Loading the corresponding `.obj` file and exporting it would result in lost information.

### Proposed Solutions:

1. **Enhanced material support:** Add support for the (optional) index of refraction (`Ni`) to SimpleMaterial similar to the optional `SimpleMaterial.image`.
2. **Flexible export option:** Introduce a flag `export_all: bool = False` to `SimpleMaterial.to_obj()`, allowing for the export of all parsed material properties from kwargs. Note: This requires careful handling of hash generation to accommodate non-hashable types.

### Setting the material names during import
I was also wondering why the material names are parsed but not used in the [generation of the `SimpleMaterial` instances]((https://github.com/mikedh/trimesh/blob/4.6.0/trimesh/exchange/obj.py#L86))? All materials are called `material_0`. Extending it as follows would make the names available before export.

This would store the original material names (instead of `material_0`) in the `.obj` file and allow re-using the original `.mtl` file.
```python
materials = {k: SimpleMaterial(**v) for k, v in material_kwargs.items()}
for k, v in materials.items():
    materials[k].name = k
```

- Python version: 3.11.9
- trimesh version: 4.6.0

binggao1230:
Fixes #2349.

When loading a Wavefront `.obj` with an associated `.mtl`, the optional
MTL keys `Ni` (index of refraction), `d` (dissolve), `Tf` (transmission
filter) and `illum` (illumination model) were parsed by `parse_mtl` but
only landed in `SimpleMaterial.kwargs` as raw lowercase string tokens.
They were never written back by `SimpleMaterial.to_obj`, and the
material hash didn't consider them, so two materials that differ ONLY
in (e.g.) `Ni` collided and the OBJ exporter folded them into a single
output material.

Even the trimesh-bundled `models/box.mtl` defines `Ni`, so any
round-trip through `load(box.obj).export(...)` silently dropped data.

## Reproducer

```python
import trimesh

# MTL with all optional fields
mtl = (
  "newmtl m\nKa 0.2 0.3 0.4\nKd 0.5 0.6 0.7\nKs 0.1 0.1 0.1\n"
  "Ns 250.0\nNi 1.50\nd 0.85\nTf 1.0 0.9 0.8\nillum 2\n"
)
obj = "mtllib m.mtl\nusemtl m\nv 0 0 0\nv 1 0 0\nv 0 1 0\nf 1 2 3\n"

# load, export, inspect MTL
import io
r = trimesh.resolvers.ZipResolver()
r["m.mtl"] = mtl.encode()
mesh = trimesh.load(io.StringIO(obj), file_type="obj", resolver=r)
out = trimesh.resolvers.ZipResolver()
out["m.obj"] = mesh.export(file_type="obj", mtl_name="m.mtl", resolver=out)
print(out["m.mtl"].decode())
```

| | trimesh main | this PR |
|---|---|---|
| `Ni 1.50` round-trip | ❌ dropped | ✅ |
| `d 0.85` round-trip  | ❌ dropped | ✅ |
| `Tf 1.0 0.9 0.8` round-trip | ❌ dropped | ✅ |
| `illum 2` round-trip | ❌ dropped | ✅ |
| `hash(mat_with_Ni_1.5) != hash(mat_with_Ni_2.0)` | ❌ collide | ✅ distinct |

## Fix

- Add typed constructor arguments to `SimpleMaterial`:
  `index_of_refraction` (float), `dissolve` (float),
  `transmission_filter` (length-3 float tuple) and
  `illumination_model` (int).  They default to `None` so existing
  callers that don't pass them are unaffected, and unknown extension
  keys still land in `self.kwargs`.
- Extend `SimpleMaterial.to_obj` to emit a corresponding MTL line only
  when the field is set, so we don't tag every exported material with
  bogus `Ni 1.0` / `d 1.0` defaults.
- Extend `SimpleMaterial.__hash__` to mix the four new fields in, so
  materials that differ only in these properties don't collide during
  deduplication in `export_obj`.
- Extend `parse_mtl`'s key-to-kwarg map so the loader actually populates
  the new typed fields, and add a small `Tr` handler that converts
  `Tr <transparency>` into `dissolve = 1 - Tr` (`d` wins if both are
  present, via `setdefault`).

## Tests

Three new tests in `tests/test_obj.py` (all pass, all fail on main):

- `test_mtl_optional_properties_roundtrip` — writes an MTL with every
  optional field, loads it through the resolver flow, asserts the typed
  fields populate, exports back out, asserts the written MTL contains
  the expected `Ni`/`d`/`Tf`/`illum` lines, and re-loads to confirm a
  clean roundtrip.
- `test_mtl_Tr_maps_to_dissolve` — covers the `Tr` -> `dissolve`
  inversion.
- `test_mtl_optional_properties_distinct_hash` — builds six
  SimpleMaterials sharing identical color/glossiness but differing only
  in optional fields and asserts all six hash distinctly while two
  materials with identical args still hash equal.

Local `pytest tests/test_obj.py tests/test_texture.py tests/test_visual.py
tests/test_gltf.py`: 105 pass, 1 unrelated fail (pycollada missing).
Ruff clean.

### DIFF
--- a/tests/test_obj.py
+++ b/tests/test_obj.py
@@ -392,6 +392,124 @@ def test_mtl_color_roundtrip():
     assert g.np.isclose(m.visual.material.glossiness, r.visual.material.glossiness)
 
 
+def test_mtl_optional_properties_roundtrip():
+    # regression for issue #2349: SimpleMaterial used to silently
+    # drop `Ni`, `d`, `Tf` and `illum` on export because they were
+    # parsed only into `SimpleMaterial.kwargs` and never written
+    # back out by `to_obj`.  exercise the full
+    # parse_mtl -> SimpleMaterial -> to_obj roundtrip through the
+    # OBJ loader/exporter.
+    mtl_text = "\n".join([
+        "newmtl PropTest",
+        "Ka 0.20 0.30 0.40",
+        "Kd 0.50 0.60 0.70",
+        "Ks 0.10 0.10 0.10",
+        "Ns 250.0",
+        "Ni 1.50",
+        "d 0.85",
+        "Tf 1.0 0.9 0.8",
+        "illum 2",
+        "",
+    ])
+    obj_text = "\n".join([
+        "mtllib mat.mtl",
+        "usemtl PropTest",
+        "v 0 0 0",
+        "v 1 0 0",
+        "v 0 1 0",
+        "vt 0 0",
+        "vt 1 0",
+        "vt 0 1",
+        "f 1/1 2/2 3/3",
+        "",
+    ])
+
+    r = g.trimesh.resolvers.ZipResolver()
+    r["mat.mtl"] = mtl_text.encode("utf-8")
+    mesh = g.trimesh.load(
+        g.io.StringIO(obj_text), file_type="obj", resolver=r
+    )
+
+    mat = mesh.visual.material
+    assert g.np.isclose(mat.index_of_refraction, 1.50)
+    assert g.np.isclose(mat.dissolve, 0.85)
+    assert g.np.allclose(mat.transmission_filter, (1.0, 0.9, 0.8))
+    assert mat.illumination_model == 2
+
+    # round-trip through OBJ export
+    out = g.trimesh.resolvers.ZipResolver()
+    out["m.obj"] = mesh.export(file_type="obj", mtl_name="mat.mtl", resolver=out)
+    mtl_out = out["mat.mtl"].decode("utf-8")
+    assert "Ni 1.50000000" in mtl_out
+    assert "d 0.85000000" in mtl_out
+    assert "Tf 1.00000000 0.90000000 0.80000000" in mtl_out
+    assert "illum 2" in mtl_out
+
+    # parse the exported MTL back and confirm values survived
+    obj_bytes = out["m.obj"]
+    if isinstance(obj_bytes, str):
+        obj_bytes = obj_bytes.encode("utf-8")
+    re_loaded = g.trimesh.load(
+        g.io.BytesIO(obj_bytes), file_type="obj", resolver=out
+    )
+    rmat = re_loaded.visual.material
+    assert g.np.isclose(rmat.index_of_refraction, 1.50)
+    assert g.np.isclose(rmat.dissolve, 0.85)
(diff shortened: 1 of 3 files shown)


binggao1230:
Follow-up pushed in `db265e085080b5d78a386da891ef8b0ded3ac9d4`.

What changed:
- The round-trip test now encodes the exported OBJ string before storing it in `ZipResolver`, so the test data matches `ZipResolver.get()`'s existing bytes return contract under `pytest-beartype`.

Local verification:
- `uv run --with pytest-beartype pytest --beartype-packages=trimesh tests/test_obj.py::test_mtl_optional_properties_roundtrip -q`
- `python -m pytest tests/test_obj.py::test_mtl_optional_properties_roundtrip tests/test_obj.py::test_mtl_Tr_maps_to_dissolve tests/test_obj.py::test_mtl_optional_properties_distinct_hash -q`
- `uv run --with pytest-beartype --with pillow --with scipy --with networkx --with pycollada --with shapely --with lxml pytest --beartype-packages=trimesh tests/test_obj.py -q`
- `python -m ruff check tests/test_obj.py && git diff --check`

Remote status: all GitHub checks are passing now, including `Run Tests In Docker`; merge state is `CLEAN`.

PR #2558, opened 2026-05-28
Validate face_normals setter against every face, sign-agnostic (fixes #2394)

### LINKED ISSUE #2394: Strange behaviour when setting face_normals in Trimesh object
The setter property `face_normals` of `triemesh.Trimesh` is very strange. It essentialy checks if the normals you passed to the function are the same to the ones it computes with its own algorithm, but it checks only the first 20 vertices and if they are not the same it skips the assignment (https://github.com/mikedh/trimesh/blob/bcbc8f14739a2cd60f2130d20617cf0bfd54ef62/trimesh/base.py#L440C1-L446C19).

This is very clear with this example:
```python
import trimesh
import numpy as np

mesh = trimesh.creation.icosphere()
face_normals = mesh.face_normals
new_face_normals = face_normals.copy()
new_face_normals *= -1
mesh.face_normals = new_face_normals
print(np.allclose(mesh.face_normals, face_normals))  # True
print(np.allclose(mesh.face_normals, new_face_normals))  # False

mesh = trimesh.creation.icosphere()
face_normals = mesh.face_normals
new_face_normals = face_normals.copy()
new_face_normals[20:] *= -1
mesh.face_normals = new_face_normals
print(np.allclose(mesh.face_normals, face_normals))  # False
print(np.allclose(mesh.face_normals, new_face_normals))  # True
```

So my question is: what's the point of assigning a value only if it is equal to a fixed precomputed value? Isn't it like not assigning anything at all?

gaoflow:
Fixes #2394.

The `Trimesh.face_normals` setter recomputed normals for only the
**first 20 faces** and demanded the supplied array equal those
*outward* normals exactly. This produced two surprising behaviors
(both reported in the issue):

1. **Only the first 20 faces were checked**, so an array that matched
   there but was garbage afterwards was accepted wholesale.
2. It required equality with the canonical **outward** normals, so a
   legitimately sign-flipped (or otherwise valid) normals array was
   silently dropped — even though it's a perfectly good set of face
   normals. (@mikedh in the thread: *"they probably shouldn't be
   settable… doing a check on the whole array isn't that much faster
   than recomputing, hence the slightly janky 'only check the first
   20' behavior."*)

## Reproducer (from the issue)

```python
import trimesh, numpy as np

mesh = trimesh.creation.icosphere()
fn = mesh.face_normals.copy()
mesh.face_normals = -fn          # fully flipped -> still valid normals
print(np.allclose(mesh.face_normals, -fn))   # main: False  -> this PR: True

mesh = trimesh.creation.icosphere()
fn = mesh.face_normals.copy()
bad = fn.copy(); bad[20:] = [0, 0, 1]         # garbage past face 20
mesh.face_normals = bad
print(np.allclose(mesh.face_normals, bad))    # main: True (!) -> this PR: False
```

## Fix

Replace the first-20 equality check with a full, **sign-agnostic
perpendicularity** check: a supplied normal is accepted iff it is
parallel or anti-parallel to the cross product of the triangle's edges
(the unnormalized true-normal direction, already available via the
cached `triangles_cross`). The residual is the sine of the angle
between the supplied normal and the face plane, computed scale-
invariantly; zero-area faces have an undefined normal and are treated
as always-valid to match the getter's padding behavior.

Result:
- validates **every** face instead of a 20-face sample,
- accepts user-supplied flipped/valid normals (the explicit request in
  the issue),
- still rejects normals that don't correspond to the face geometry,
  falling back to regenerated cross-product normals.

This is essentially the perpendicularity check @mikedh sketched in the
thread, but reusing the already-cached `triangles_cross` so it doesn't
recompute and handles both edges at once.

## Tests

Adds `test_face_normals_setter_validation` (full sign-flip acceptance;
the >20-face partial-garbage rejection the old check missed; rejection
of normals rotated out of the face plane). The existing
`test_face_normals` (garbage + all-zero rejection) passes unchanged.

Local: `tests/test_normals.py tests/test_stl.py tests/test_gltf.py
tests/test_mesh.py tests/test_ply.py` all pass; the only failures in my
run are `test_creation.py` cases that need `shapely` (not installed in
this venv). Ruff clean.

### DIFF
--- a/tests/test_normals.py
+++ b/tests/test_normals.py
@@ -130,6 +130,49 @@ class NormalsTest(g.unittest.TestCase):
         mesh.face_normals = g.np.zeros_like(mesh.faces)
         assert g.np.allclose(g.np.linalg.norm(mesh.face_normals, axis=1), 1.0)
 
+    def test_face_normals_setter_validation(self):
+        """
+        Regression for issue #2394: the face_normals setter used to
+        only validate the first 20 faces against the recomputed
+        normals, so it could silently accept wrong normals (if the
+        first 20 happened to match) and silently drop legitimately
+        sign-flipped normals. The setter should validate every face
+        and accept any normals that are perpendicular to the face
+        plane regardless of sign.
+        """
+        # use enough faces that a "first 20" check would miss errors
+        # (use a fresh mesh per case to avoid face_normals cache carryover)
+        def fresh():
+            m = g.trimesh.creation.icosphere(subdivisions=2)
+            assert len(m.faces) > 20
+            return m, m.face_normals.copy()
+
+        # fully sign-flipped normals are still perpendicular to every
+        # face and must now be accepted as-is (the user's explicit
+        # request in #2394)
+        mesh, computed = fresh()
+        flipped = -computed
+        mesh.face_normals = flipped
+        assert g.np.allclose(mesh.face_normals, flipped)
+
+        # normals that are correct for the first 20 faces but garbage
+        # afterwards must be rejected (the old check missed this)
+        mesh, computed = fresh()
+        partial_garbage = computed.copy()
+        partial_garbage[20:] = [0.0, 0.0, 1.0]
+        mesh.face_normals = partial_garbage
+        assert not g.np.allclose(mesh.face_normals, partial_garbage)
+        # rejected -> falls back to the true computed normals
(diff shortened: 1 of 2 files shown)


mikedh:
Hey, this was discussed in #2394

PR #2560, opened 2026-05-28
Triangulate non-convex OBJ polygon faces with earcut (fixes #2519)

### LINKED ISSUE #2519: Earclipping issue
there's an issue where trimesh returns a version of the obj without Earclipping, image 1 is the input and image 2 is the output
the only thing that i did here is change the normals and that's all

the code used:
```py
async def FlatNormals(model: discord.Attachment):
    async with manager.session.get(model.url) as response:
        mesh: trimesh.Trimesh = await asyncio.to_thread(
            trimesh.load_mesh, io.BytesIO(await response.read()), force="mesh", file_type="obj"
        )

    def job():
        mesh.vertex_normals = np.tile([0, 0, 1], (len(mesh.vertices), 1))
        mesh.face_normals = np.tile([0, 0, 1], (len(mesh.faces), 1))

        obj_bytes = mesh.export(file_type="obj")
        return io.BytesIO(obj_bytes)

    return await asyncio.to_thread(job) # ignore how i'm using to_thread alot, the bot gets blocked if i don't do this despite being put in an asyncio task
```

input:
Image1

output:

Image2

gaoflow:
Fixes #2519.

OBJ faces with more than three corners were triangulated with a naive
triangle fan rooted at the first corner. That is only valid for
**convex** polygons — for a non-convex n-gon the fan emits triangles
that spill outside the polygon, inflating the surface area and visibly
corrupting the geometry on load/export (the artifact in the issue).

## Reproducer

`models/wallhole.obj` (the model the reporter narrowed it down to)
contains two non-convex 11-gons:

```python
import trimesh
m = trimesh.load_mesh('models/wallhole.obj', process=False)
print(m.area)   # main: 761.42   this PR: 721.42  (== exact sum of polygon areas)
```

On `main` each of the two non-convex faces covers ~20 extra units of
area **outside** its own boundary. After the fix the loaded surface
area equals the exact planar-area sum of the source polygons.

## Fix

When a triangulation engine (`mapbox_earcut`) is available and the
vertex positions are known, polygon faces are triangulated by
projecting them onto their best-fit plane and running earcut. Unlike
the `triangle` engine, earcut inserts **no new vertices**, so the
existing face tokens can be reused by index — preserving the per-corner
`vt`/`vn` references. When no engine is installed or the projection is
degenerate, the loader falls back to the previous fan, so default
installs are unchanged.

Two code paths fanned polygons, both addressed:

- the slow per-line fallback (mixed tri/quad/ngon files, e.g.
  `wallhole.obj`) now calls the new `_ngon_triangulate_tokens` helper,
  falling back to `_fan_triangulate_tokens`;
- the vectorized fast path silently fanned *uniform* n-gon files, so
  faces with more than four corners are now routed to the fallback.
  Triangles and quads keep using the fast vectorized path unchanged.

## Tests

- `test_obj_nonconvex_ngon` — loads `wallhole.obj` and asserts the area
  equals the exact sum of the source polygons' planar areas;
- `test_obj_nonconvex_ngon_synthetic` — an explicit concave pentagon;
- both no-op gracefully when no triangulation engine is installed.

Local `pytest tests/test_obj.py tests/test_loaded.py
tests/test_texture.py` passes (57 passed; the single failure is
`test_multi_nodupe`, which needs `pycollada`, unrelated). Ruff clean.

### DIFF
--- a/tests/test_obj.py
+++ b/tests/test_obj.py
@@ -105,6 +105,78 @@ def test_obj_quad():
     assert g.np.isclose(mesh.area, rec.area)
 
 
+def _polygon_area_3d(points):
+    # planar area of a 3D polygon given its ordered boundary points
+    center = points.mean(axis=0)
+    _, _, vh = g.np.linalg.svd(points - center)
+    flat = g.np.column_stack(((points - center).dot(vh[0]), (points - center).dot(vh[1])))
+    x, y = flat[:, 0], flat[:, 1]
+    return abs(0.5 * g.np.sum(x * g.np.roll(y, -1) - g.np.roll(x, -1) * y))
+
+
+def test_obj_nonconvex_ngon():
+    # regression for issue #2519: faces with more than 4 corners were
+    # triangulated with a naive fan from the first corner, which is
+    # only valid for convex polygons. for a non-convex n-gon the fan
+    # produces triangles that spill outside the polygon, inflating the
+    # surface area and corrupting the geometry.
+    #
+    # `models/wallhole.obj` contains two non-convex 11-gons; when an
+    # earcut engine is available the loaded surface area should equal
+    # the exact sum of the source polygons' planar areas.
+    engines = [k for k, exists in g.trimesh.creation._engines if exists]
+    if not engines:
+        # no triangulation engine installed: loader falls back to a
+        # fan and we can't assert the corrected area
+        return
+
+    mesh = g.get_mesh("wallhole.obj", process=False)
+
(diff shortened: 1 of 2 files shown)


mikedh:
Hey, if anywhere this should probably be in a unification of `trimesh.geometry.triangulate_quads` which actually does handle ragged sequence polygons with a fan automatically and quickly. 

Supporting non-convex polygons in the OBJ load doesn't seem like something we should do as it turns a simple index operation (fan) into a heavy plane-fit-earcut operation.

PR #2565, opened 2026-06-12
Don't require embreex on Linux/aarch64 where it has no wheel

### LINKED ISSUE #2564: Install trimesh[easy] fails because embreex is not avail for arm64/linux
When I try to install `trimesh[easy]` in docker image on MacOS on arm64 architecture, it fails because `embreex` can't be installed/build:
Likely arm64 vs linux architecture has no wheel.

Within my docker build that installs using `uv` ( the `uv.lock` file was generated on the host system MacOS/arm64 ):

```text
#5 [stage-0  1/20] FROM docker.io/library/python:3.12-slim-bookworm
#6 FROM ghcr.io/astral-sh/uv:0.11@sha256:eaa5f1a3305307aaf9e67fe2bbba1d85ebbb2d8a63bce23af21797bfafbe0f8b
....
#14 [stage-0  5/20] COPY --from=ghcr.io/astral-sh/uv:0.11 /uv /uvx /bin/
#17 [stage-0 11/20] COPY pyproject.toml uv.lock ./
#18 [stage-0 12/20] RUN python3 --version && uv --version &&   uv sync --frozen --no-install-project
#18 0.119 Python 3.12.13
#18 0.125 uv 0.11.20 (aarch64-unknown-linux-musl)
#18 0.254 Using CPython 3.12.13 interpreter at: /usr/local/bin/python3
#18 0.254 Creating virtual environment at: .venv
#18 0.263 error: Distribution `embreex==4.4.0 @ registry+https://....` can't be installed because it doesn't have a source distribution or wheel for the current platform
#18 0.263 
#18 0.263 hint: You're on Linux (`manylinux_2_36_aarch64`), but `embreex` (v4.4.0) only has wheels for the following platforms: `manylinux_2_27_x86_64`, `manylinux_2_28_x86_64`, `macosx_13_0_arm64`, `macosx_13_0_x86_64`, `win_amd64`; consider adding "sys_platform == 'linux' and platform_machine == 'aarch64'" to `tool.uv.required-environments` to ensure uv resolves to a version with compatible wheels```
````

It seems to work though with this simple Dockerfile using `pip` when I build the docker image on my MacOS/arm64 host:

```Dockerfile
FROM python:3.12-slim-bookworm
RUN uname -a
RUN pip3 install trimesh[easy]
```

yielding the following output

```text
#4 [1/3] FROM docker.io/library/python:3.12-slim-bookworm
#5 [2/3] RUN uname -a
#5 0.093 Linux buildkitsandbox 6.12.76-linuxkit #1 SMP Thu May 28 18:54:18 UTC 2026 aarch64 GNU/Linux
#6 [3/3] RUN pip3 install trimesh[easy]
#6 1.443 Collecting trimesh[easy]
....
```

gaoflow:
Fixes #2564.

`trimesh[easy]` fails to install on Linux/arm64 (e.g. `python:3.12-slim` Docker on Apple silicon) because `embreex` is pulled in unconditionally but ships no `manylinux` **aarch64** wheel, so the install falls through to a source build that needs the Embree C library and errors out.

This is a regression from the Embree 4 upgrade in 4.12.0: 4.11.x guarded the dependency with `embreex; platform_machine=='x86_64'`, and that marker was dropped to `embreex` when embreex 4.x started shipping macOS-arm64 wheels. That fixed macOS arm64 but over-broadened to Linux aarch64, which embreex still has no wheel for.

The fix narrows the guard to exactly the wheel-less platform rather than reinstating the old x86_64-only marker (which would needlessly drop the now-available macOS-arm64 acceleration):

```toml
"embreex; platform_machine!='aarch64'"
```

embreex is an optional ray-tracing accelerator, so on Linux/aarch64 trimesh simply falls back to its built-in pure-Python ray tracer.

Verified against embreex 4.4.0's published wheels (macOS arm64 + x86_64, Linux x86_64, Windows AMD64 — no Linux aarch64) and by evaluating the marker: it stays `True` on every platform that has a wheel and is `False` only on Linux/aarch64.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,8 @@ easy = [
     "rtree",
     "httpx",
     "scipy",
-    "embreex",
+    # embreex ships no Linux/aarch64 wheel, so require it everywhere except there
+    "embreex; platform_machine!='aarch64'",
     "pillow",
     "vhacdx; python_version>='3.9'",
     # old versions of mapbox_earcut produce incorrect values on numpy 2.0


binggao1230:
Looks like this landed independently on `main` in 78599e0 — `pyproject.toml` now has `embreex; platform_machine!='aarch64'`, which is exactly the change here. Closing as superseded. Thanks!

PR #2571, opened 2026-06-24
feat(exchange): add support for draco compression in GLTF

mmatl:
Adds support for the draco mesh compression extension for GLTF / GLB files via the DracoPy Python library, which has wheels for all major operating system configurations.

@mikedh Let me know what other tests you'd like added here. I loaded and re-exported a few different models from Blender and it seems to work as expected.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,7 @@ trimesh = [
 
 # this is the base extra most users will want
 easy = [
+    "DracoPy",
     "colorlog",
     "manifold3d>=2.3.0",
     "charset-normalizer",
--- a/tests/test_gltf.py
+++ b/tests/test_gltf.py
@@ -1151,6 +1151,27 @@ class GLTFTest(g.unittest.TestCase):
         mean_squared_error = ((a - b) ** 2).sum() / g.np.prod(a.shape)
         assert mean_squared_error < 10.0
 
+    def test_draco_roundtrip(self):
+        m = g.get_mesh("fuze.obj")
+        nc = m.export(file_type="glb")
+        e = m.export(file_type="glb", extension_draco=True)
+        r = g.trimesh.load_mesh(g.trimesh.util.wrap_as_stream(e), file_type="glb")
+
+        assert len(m.faces) == len(r.faces)
+        assert len(e) < len(nc)  # ensure compression worked
+
+        # compare RGBA images for the roundtripped texture
+        a = g.np.array(m.visual.material.image)
+        b = g.np.array(r.visual.material.baseColorTexture)
+        assert a.shape == b.shape
+
+        # test with normals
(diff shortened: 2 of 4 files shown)


mikedh:
Thanks for the PR! Yeah I was hoping with the extension system we could make it so GLTF extensions could add support without adding them directly to trimesh. Unfortunately some of the upstream packages have occasionally broken the trimesh install for people, and made me a little wary of expanding the dependency tree. I made a throwaway import-only draco extension as a demo: 

```
import httpx

import trimesh

# a draco-compressed model
glb_url = (
    "https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Assets/"
    "main/Models/BrainStem/glTF-Draco/BrainStem.gltf"
)


if __name__ == "__main__":
    # currently silently makes values zero, not great
    no_draco = trimesh.load_scene(glb_url, allow_remote=True, session=httpx.Client())
    print(f"no dracox:  {no_draco}\n{no_draco.bounds.round(2)}")

    # registers handlers on-import, `pip install dracox`
    import dracox  # noqa

    # populates for load
    with_draco = trimesh.load_scene(glb_url, allow_remote=True, session=httpx.Client())
    print(f"with dracox:  {with_draco}\n{with_draco.bounds.round(2)}")
```
Which outputs:
```
no dracox:  <trimesh.Scene(len(geometry)=59)>
[[0. 0. 0.]
 [0. 0. 0.]]
with dracox:  <trimesh.Scene(len(geometry)=59)>
[[-1.25 -0.   -0.34]
 [ 1.16  1.83  0.26]]
```

DracoPy does look better than my throwaway, I'm not totally opposed to adding it but I'd like to think about the dep situation and leave this open for a while. As I was writing this example I did notice/remember trimesh returns all-zero values silently when a draco mesh comes in, which is not great and I'll add a warning.

When I've looked in to draco my main objection was that it was a single C++ library, not a specification anyone could use. To be honest I was hoping draco would go away in favor of https://github.com/KhronosGroup/glTF/issues/1834 but given how that issue is me and one other person maybe that ship has sailed haha.

For exports I've been using `zstandard.compress(scene.export(file_type='glb'), 22)` which at least for me has been WAY easier to deal with on weird platforms (WASM particularly).

mmatl:
There are alternative compression schemes: https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_meshopt_compression/README.md. These come with their own libraries: https://github.com/zeux/meshoptimizer. Implementing this type of compression efficiently is generally going to be a pretty nasty C++ scheme with lots of AVX optimizations and such, so it's hard to avoid third-party libraries in general for this.

We could also put DracoPy into a separate dependency group for extensions and loaders?

mmatl:
@mikedh lmk if you need anything else from me for this.

PR #2574, opened 2026-06-28
Fixed capsule creation causing radius error

long0900:
The current way to draw linestring causes high radius error, especially with low numbers of sections.
https://github.com/mikedh/trimesh/blob/a513c0e2cfa6dfbd3121a6dfd8ceb002b9ef55d8/trimesh/creation.py#L1004-L1006
With even number of sections, theta never reaches zero, causing radius error. For example, with 8 sections:
`
[-1.57079633 -1.12199738 -0.67319843 -0.22439948  0.22439948  0.67319843
  1.12199738  1.57079633]
`
This PR fixed it by making sure theta goes to zero for each top/bottom quarter circle.

### DIFF
--- a/trimesh/creation.py
+++ b/trimesh/creation.py
@@ -981,7 +981,7 @@ def capsule(
     radius : float
       Radius of the cylinder and hemispheres
     count : (2,) int
-      Number of sections on latitude and longitude
+      Number of sections on latitude and longitude on each hemisphere
     transform : None or (4, 4) float
       Transform to apply to mesh after construction
     Returns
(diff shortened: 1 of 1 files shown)


mikedh:
Thanks for the PR! Can you add a test that fails, I don't totally understand which radius errors? And it seems like we could just do `linspace(-pi/2, pi/2, count[0] + count[0] % 2)` to avoid the multiple array allocations?

long0900:
I'm not sure what you meany about a test but here's an example capsule with radius of 1 and length of 1:
`mesh = trimesh.creation.capsule(1, 1, count=(6, 12))`

image

Issue is with a single linspace, theta never reaches 0 (it can with odd number of sections but that will cause symmetry issues too) so cos(theta) can't reach 1 so the radius of the cylinder can't fully reach the input radius when you revolve the linestring.

There must be 2 zeros at the middle of the theta array so when it's split by half, each quarter circle will have their own "cos(0)". It can't be done with a single linspace.

This is with my modification:
image

The side effect is count[0] being twice as effective as the old one, since it's the number of sections on each hemisphere instead of in total. I wanted to keep it simple and not to deal with odd number of sections like the old one.

If array reallocation is to be avoided then I guess it can be pre-allocated?

PR #2576, opened 2026-07-08
fix: handle None polygons in enclosure_tree and polygons_full

Eren845205:
## Problem

`Path2D.polygons_full` crashes with:

```
AttributeError: 'NoneType' object has no attribute 'exterior'
```

when a section consists of a single degenerate closed path.

`paths_to_polygons` documents that unrecoverable geometry produces `None`
entries ("will be None if geometry is unrecoverable"), and the multi-polygon
code path of `enclosure_tree` correctly excludes those via the bounds check
(`getattr(polygon, "bounds", [])` filters `None`). However the
single-polygon early exit returns `roots=[0]` without checking, so the
`None` reaches `polygons_full`, which calls `closed[root].exterior`.

We hit this in the wild via `Trimesh.section_multiplane` on a real
manufacturing STL voxelized at a rotated pose: one z-slice produced a single
unrecoverable circuit, and the whole pipeline died on the `AttributeError`.

## Fix (two layers)

1. **`enclosure_tree`** (root cause): the single-polygon early exit no
   longer emits a `None` polygon as a root — matching the behavior of the
   multi-polygon path, which already excludes `None`.
2. **`Path2D.polygons_full`** (defense in depth): skip root curves whose
   closed polygon is `None` and filter `None` children when building holes,
   since `polygons_closed` explicitly documents `None` entries. The
   docstring is updated accordingly (`m <= len(self.root)`).

## Tests

- `test_enclosure_tree_none`: a sub-4-vertex closed path produces `None`
  from `paths_to_polygons`; asserts the single-polygon early exit returns no
  roots (previously `roots=[0]`), and that the multi-polygon path keeps
  excluding `None` alongside a valid polygon.
- `test_polygons_full_none`: simulates what `mesh.section` produces in the
  wild (a `polygons_closed` array containing `None` next to a valid
  polygon) and asserts `polygons_full` returns the valid polygon instead of
  raising.

`tests/test_polygons.py` passes 10/10 locally. The two pre-existing failures
in `tests/test_paths.py::test_svg_arc_snap` (missing optional dependency)
and `tests/test_section.py::SliceTest::test_slice_onplane` are present on
unmodified `main` in my environment as well and are unrelated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XbfEVQhLMnkCoG755dSVZ7

### DIFF
--- a/tests/test_polygons.py
+++ b/tests/test_polygons.py
@@ -332,10 +332,12 @@ def test_polygons_full_none():
         entities=[g.trimesh.path.entities.Line([0, 1])],
         vertices=g.np.array([[0.0, 0.0], [1.0, 1.0]]),
     )
-    # used to raise AttributeError on the `None` root
+    # used to raise AttributeError on the `None` root; correspondence
+    # with `self.root` is preserved (unrecoverable entry stays `None`)
     full = path.polygons_full
-    assert len(full) == 1
-    assert g.np.isclose(full[0].area, square.area)
+    assert len(full) == len(path.root) == 2
+    assert full[0] is None
+    assert g.np.isclose(full[1].area, square.area)
 
 
 if __name__ == "__main__":
--- a/tests/test_polygons.py
+++ b/tests/test_polygons.py
@@ -275,6 +275,69 @@ def test_native_centroid():
         assert g.np.isclose(abs(area), p.area)
 
 
+def test_enclosure_tree_none():
+    # `paths_to_polygons` documents that unrecoverable geometry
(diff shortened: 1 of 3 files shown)


mikedh:
This makes sense, thanks for the PR! The `polygons_full` changing type from `Polygon | None` to `Polygon` via `append` breaks the correspondence between `.polygons_full` and `.root` though. I think we should probably not do that, but the degenerate filter in holes and test seem good.

PR #2577, opened 2026-07-08
fix: subdivide performing iterations+1 subdivisions

abhinow03:
mesh.subdivide(iterations=n) ran one extra subdivision: the recursion signaled "done" by passing iterations=None, but that's also the default that performs a single subdivision, so every explicit-iterations call got n+1. This changes the recursion to continue only while iterations > 1 and adds a regression test asserting that face counts scale by 4^n.

Before/after on trimesh.creation.box() vertex counts: iterations=1 98→26, iterations=2 386→98.


mikedh:
Makes sense, thanks for the fix!

PR #2578, opened 2026-07-16
Adds estimate_normals to PointCloud

proebrock:
Hi everybody,

This merge request adds the estimation of normal vectors to a point cloud. It also adds normals as a member variable to the PointCloud class. This is a significant change.

Let me explain my intentions. I am a lecturer and researcher at a small university of applied sciences in Switzerland (FHGR), where I work on 3D reconstructions using 2D and 3D sensors. I have used the Open3D library so far, but it has been inactive since 2023, with the last release in January 2025. For my projects, I am looking for a replacement and have found Trimesh. Trimesh is much stronger than Open3D in mesh processing, which I like very much. I also appreciate that the implementation uses plain NumPy, and that the Trimesh project tries to limit its dependencies on other libraries. However, Trimesh lacks some features regarding point clouds, in my opinion. To name a few:

* Outlier filtering
* ROI filtering
* Normal estimation based on nearest neighbors (this MR)
* Normal orientation
* Clustering point clouds (DBSCAN)
* Plane segmentation using RANSAC

I am willing to spend time adding these features to the Trimesh library because they would benefit my teaching and research projects. However, I don't know if that's the direction in which you want to move the Trimesh project. So, before we discuss the details of this MR, please check if this idea aligns with your project's overall goals.

This is my first contribution to the project, so please review it carefully.

* Is it okay to add normals as a member variable to the point cloud like this? I check if we have a normal for each point, and I normalize the normals to ensure they have a length of 1.
* I put the PCA in a helper function. Would you prefer to integrate it into the estimate_normals method, or is that okay?
* Should I add Python type hints?

In my opinion, the performance of this implementation is okay. Yes, there is a for loop. However, eliminating it would be difficult because you would need to store the neighbors of all points in memory simultaneously and perform a PCA on each set.

The unit tests pass, so that part should be fine.

Best regards

Philipp

### DIFF
--- a/tests/test_points.py
+++ b/tests/test_points.py
@@ -298,3 +298,47 @@ def test_radial_sort():
     sort = g.trimesh.points.radial_sort(points[order], origin=[0, 0, 0], normal=[0, 0, 1])
     # should have re-established original order
     assert g.np.allclose(points, sort)
+
+
+def test_estimate_normals_plane():
+    # Define plane
+    n = g.np.array((-3.0, 5.0, -7.0))
+    n = n / g.np.linalg.norm(n)
+    d = 11.0
+    # Create point cloud on plane
+    x = g.np.linspace(-10.0, 10.0, 50)
+    y = g.np.linspace(-10.0, 10.0, 50)
+    x, y = g.np.meshgrid(x, y)
+    vertices = g.np.empty((x.size, 3))
+    vertices[:, 0] = x.ravel()
+    vertices[:, 1] = y.ravel()
+    vertices[:, 2] = (n[0] * vertices[:, 0] + n[1] * vertices[:, 1] + d) / -n[2]
+    pcl = g.trimesh.points.PointCloud(vertices)
+    # pcl.show()
+
+    pcl.estimate_normals()
+
+    dot = g.np.sum(n * pcl.normals, axis=1)
+    dot = g.np.clip(dot, -1.0, 1.0)  # Otherwise rounding errors result in NaN
+    angle_errors = g.np.arccos(dot)
+    mask_close_to_zeros = angle_errors < g.np.deg2rad(1.0)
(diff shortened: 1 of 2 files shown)


mikedh:
Thanks for the PR!! Couple notes:

- I think this should be a pure function `PointCloud.estimate_normals(self, clusters, etc) -> NDArray2D[np.float64]`, not a property, and if a user wants to assign them then they can.
- I think all of the logic should be contained in one `estimate_normals` function, and be explicit about the algorithm. 

And the major point, we'd have to vectorize this and remove the per-point loop to get it to be reasonable fast in Python:
```
        for i in range(self.shape[0]):
            _, indices = kdtree.query(self.vertices[i], k=knn)
            components, _, _ = principal_component_analysis_3d(self.vertices[indices])
            normals[i] = components[2, :]
        self.normals = normals
```

Doing the neighbor query is pretty easy, the PCA might be trickier:
```
# clamp or raise to something sane
cluster_size = int(np.clip(cluster_size, 3, 10)

# get distances and indices in one call, as 2D numpy arrays
# indices will be `(len(vertices), cluster_size)` integers, always containing the self-reference
distance, indices = cKDTree(self.vertices).query(self.vertices, k=cluster_size)

# shouldn't we be weighing something by ~`distance ** -2`?
... somehow a vectorized version of PCA goes here
```

proebrock:
Thanks for taking the time to review my PR!!

I revised the code according to your comments:

* I made the method a pure function, the normals are no longer an attribute of `PointCloud`
* All logic is now contained in the method `estimate_normals`, there is no PCA method anymore
* I got rid of the loop. With the help of your suggestion it wasn't too complicated to compute the PCA in parallel for all nearest neighbor points.

mikedh:
Great thanks for the refactor!

PR #2579, opened 2026-07-21
perf(proximity): batch nearby_faces r-tree queries when rtree supports it

### LINKED ISSUE #1116: faster trimesh.proximity.closest_point for large query point sets
Thank you for this amazing library.

I would like to compute the distance of a large number of point to a relatively simple mesh.
I am using `trimesh.proximity.closest_point`, but unfortunately it is quit slow : it takes 30 second when querying 500000 points on my machine. 

One major bottleneck seem to be the line  `candidates = [list(rtree.intersection(b)) for b in bounds]`  in ` proximity.py`.
I believe this could run much faster if the loop was implemented in `c++`.  Ideally rtree.intersection would accept a list of boxes and implement the loop in c++ and return the list of list of candidates (similarly to `scipy.spatial.KDTree.query_ball_point`), but that does not seem supported by rtree.

Here is the code I use to get some timings
```
import trimesh
import time
import numpy as np

vertices = np.random.rand(20, 3)
faces = (np.random.rand(10, 3) * 20).astype(np.int)


mesh = trimesh.Trimesh(faces=faces, vertices=vertices)

nb_points_list = [1000, 10000, 100000, 500000]
durations_list = []

for nb_points in nb_points_list:
    points = np.random.rand(nb_points, 3)

    start = time.time()
    trimesh.proximity.closest_point(mesh, points)
    duration = time.time()-start
    durations_list.append(duration)
    print(f"trimesh.proximity.closest_point wit {nb_points} points took {duration} seconds")

plt.plot(nb_points_list, durations_list)
plt.xlabel("nb qurey points")
plt.xlabel("duration in seconds")
plt.show()
```
I get this timings 
trimesh.proximity.closest_point wit 1000 points took 0.11966729164123535 seconds
trimesh.proximity.closest_point wit 10000 points took 0.6350014209747314 seconds
trimesh.proximity.closest_point wit 100000 points took 6.338748455047607 seconds
trimesh.proximity.closest_point wit 500000 points took 32.05847191810608 seconds
image

It would be great if we could list the potential ways we could accelerate the function. I looked at alternative libararies available from python to do the nearest points query but did not find a good alternative yet.

dexhunter:
Fixes #1116

`nearby_faces` queried the r-tree once per point. This batches them via `intersection_v` on rtree 1.4.1+, keeping the scalar loop where that API is defective. Ordering, duplicates, and empty input are unchanged; tests added.

Quiet paired medians: nearby_faces 4.80s to 2.51s; closest_point end-to-end 3.81s to 3.19s. Outputs identical.

### DIFF
--- a/tests/test_proximity.py
+++ b/tests/test_proximity.py
@@ -232,7 +232,8 @@ class NearestTest(g.unittest.TestCase):
                 mock.patch.object(tree, method, wraps=getattr(tree, method)) as query,
             ):
                 actual = g.trimesh.proximity.nearby_faces(mesh, points)
-            assert actual == expected
+            assert len(actual) == len(expected)
+            assert all(g.np.array_equal(a, e) for a, e in zip(actual, expected))
             assert query.call_count == (1 if method == "intersection_v" else len(points))
 
     def test_candidates_empty(self):
--- a/tests/test_proximity.py
+++ b/tests/test_proximity.py
@@ -1,3 +1,5 @@
+from unittest import mock
+
 try:
     from . import generic as g
 except BaseException:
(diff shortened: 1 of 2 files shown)


mikedh:
Great find thanks for the PR!!

I think we can vectorize the index loop with `array_split` though there's probably a more graceful way of doing this:
```
    hit_ids, hit_counts = rtree.intersection_v(
        mins=bounds[:, :3], maxs=bounds[:, 3:])
    candidates = np.array_split(
        hit_ids, np.cumsum(hit_counts)[:-1])

```

dexhunter:
Thanks! Updated the batch path to use `np.array_split(hit_ids, np.cumsum(hit_counts)[:-1])`. Five alternating paired runs kept candidate and closest-point output digests identical; `nearby_faces` median improved another 19.4% versus the prior PR head. The 12 focused proximity tests and Ruff checks pass.

mikedh:
Looks good! Quick question, do we have to preserve that bad-version check, or does it fail in a way we could  just `try:` and fall back to the list comprehension?

edit: it looks like `try: ` catches it on 1.4.0 so we don't have to hardcode the version range which is nice

PR #2580, opened 2026-07-22
Add containment-aware split via solids/orient flags

mabl:
Trimesh.split() only groups faces by edge connectivity, so a closed inner cavity shell is returned as a separate mesh from its enclosing solid. Add a `solids` flag that groups watertight shells by geometric containment parity (even nesting depth = exterior, odd = cavity) so a solid and its cavities are returned as one mesh, and an `orient` flag that normalizes shell winding to positive-volume exteriors and negative-volume cavities.

Containment is evaluated with a single bidirectional ray query per shell against the parent mesh BVH, bucketed per shell via a face-label array, replicating the parity/retry semantics of ray.contains_points without building one intersector per shell. Open patches are never attached to solids and remain subject to only_watertight/repair. Defaults are unchanged so existing split() behavior is preserved.

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -4,6 +4,13 @@ except BaseException:
     import generic as g
 
 
+def _box(extents, invert=False):
+    mesh = g.trimesh.creation.box(extents=extents)
+    if invert:
+        mesh.invert()
+    return mesh
+
+
 class GraphTest(g.unittest.TestCase):
     def setUp(self):
         self.engines = []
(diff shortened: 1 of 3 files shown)


mikedh:
Thanks for the PR! Couple notes:

- Could we generalize the existing `enclosure_tree` to be 3D? The rtree broad-phase is the same, and we'd just need a `.contains` method on an intermediate representation?
- Couldn't we do the containment check on a single representative point (i.e. one vertex per shell), and get all of the benefit of the more expensive checks if the source is well constructed?
- I don't think this should have any `_internal` helper methods, and the total diff should be ~25 lines?
- I think this should have one argument, `nested: bool = False` or something, not two

PR #2585, opened 2026-07-27
Bound archive reads by chunk instead of preallocating the size cap

spark-cv:
`decompress()` caps decompressed output by handing the budget to `read()`:

```python
data = bz2.open(file_obj, mode="r").read(MAX_ARCHIVE_SIZE + 1)
```

`MAX_ARCHIVE_SIZE` is 8 GiB, and `read(n)` allocates an `n`-byte buffer up front, so the cap
is requested as a single allocation regardless of how much data the archive actually holds.
A 54-byte file is enough to request 8 GiB.

The intent is clearly the opposite — the docstring says reads "stop one byte past the budget
rather than trusting declared member sizes", which correctly avoids trusting header fields.
The cap just needs to bound the read loop instead of being passed to `read()`.

## Reproduction

```python
import bz2, io, resource, trimesh

resource.setrlimit(resource.RLIMIT_AS, (2 << 30, 2 << 30))   # 2 GiB ceiling

payload = bz2.compress(b"solid x\nendsolid x\n")
print(len(payload))                                          # 54

trimesh.load(io.BytesIO(payload), file_type="bz2")           # MemoryError
```

Without the `setrlimit` the process requests 8589934626 bytes. No decompression bomb is
needed — any minimal valid archive reaches it.

```
File "trimesh/exchange/load.py", line 313, in _load_compressed
    archive = util.decompress(file_obj=arg.file_obj, file_type=arg.file_type)
File "trimesh/util.py", line 2037, in decompress
    data = bz2.open(file_obj, mode="r").read(MAX_ARCHIVE_SIZE + 1)
MemoryError
```

| `file_type` | over-allocates |
|---|---|
| `bz2` | yes |
| `tar.gz`, `tar.bz2` | yes |
| `zip` | no — `ZipExtFile.read` grows its buffer |

Any caller that loads untrusted archives can be made to request 8 GiB per call, so a service
accepting user-uploaded meshes is denial-of-serviceable from a request measured in tens of
bytes. Memory exhaustion only — no memory corruption involved.

## Change

Adds `_read_capped()`, which reads in 1 MiB chunks and still stops one byte past the budget,
so the existing `len(data) > MAX_ARCHIVE_SIZE` checks are untouched. Peak memory now tracks
the data actually present.

All three branches route through it. `bz2` and `tar` both over-allocated; `zip` did not, but
only because `ZipExtFile.read` grows its buffer rather than preallocating — routing it
through the same helper means that behaviour is no longer incidental. Happy to drop the `zip`
hunk if you would rather keep the diff minimal.

## Verification

| | before | after |
|---|---|---|
| peak allocation, 54-byte bz2 | 8 GiB requested | ~1 MiB |
| 1 MiB cap on 4 MiB payload | returns `cap + 1` | returns `cap + 1` |
| round trip, bz2 / zip / tar.gz | ok | ok |

- Full suite: **723 passed** (720 existing + 3 added), no regressions.
- `ruff check` and `ruff format --check` clean.

## Tests

Three cases in `tests/test_util.py::ArchiveCapTest`:

- `test_small_archive_fits_in_memory` — loads the 54-byte archive in a subprocess under an
  `RLIMIT_AS` ceiling set 512 MiB above current `VmSize`. Reverting the fix fails this with
  `MemoryError` at `util.py:2037`. Linux-gated, since it needs `/proc` and `resource`.
- `test_cap_still_enforced` — overflow detection is unchanged.
- `test_round_trip` — all three archive types still decompress to their contents.

The memory probe warms imports before applying the limit; numpy and OpenBLAS reserve
gigabytes of address space at import time and would otherwise fail before reaching the code
under test.

---

Found while building an OSS-Fuzz integration for trimesh — the harness reached this within a
handful of inputs.

If you're open to it, I'd like to upstream that integration as well. It would need a
maintainer listed as `primary_contact` on the `projects/trimesh/project.yaml` entry, since
that address receives the private reports during the 90-day disclosure window. The harnesses
cover the mesh loaders and the archive paths separately and already pass OSS-Fuzz's
`check_build`. Let me know if that's something you'd want and I'll open the PR against
`google/oss-fuzz`.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 import unittest
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


spark-cv:
Heads up — both red checks look unrelated to this PR.

`ruff` is unpinned in `pyproject.toml`, and 0.16.0 (released 2026-07-23) added
RUF036, which flags three pre-existing spots:

    trimesh/inertia.py:76
    trimesh/path/path.py:734
    trimesh/visual/color.py:978

I reproduced both failures on a clean checkout of `main` with none of my changes
in it:

- `ruff check .` returns the same three errors, which is what Check Formatting
  runs.
- `make tests` fails identically at the `ruff check trimesh` step in the tests
  stage of the Dockerfile — same exit code, same three errors — so it stops
  before pytest ever runs.

The 10 jobs that actually executed my code all passed: the 8 unit-test matrix
entries, the s390x/i386 architecture tests, and corpus loading.

Happy to include the RUF036 fixes in this PR if that's useful, or leave them for
a separate one so this stays focused — whichever you prefer.

mikedh:
Thanks for the PR! Yeah I think `read_capped` was substantially over-promising, as the only honest way to resource-limit is via systemd/docker/etc. `read_capped` was suggested by an automated audit and I added it, but on reflection I think it's a bad idea and I reverted the limit to just egregious cases in https://github.com/mikedh/trimesh/pull/2568/changes/a4d3e3f1556e821fe928107f4ce38b3a9886029f. 
 
> If you're open to it, I'd like to upstream that integration as well. It would need a
> maintainer listed as primary_contact on the projects/trimesh/project.yaml 

Sure feel free to add me, or I can take a look at some point.

PR #2587, opened 2026-08-02
Fix smoothing shifting mesh coordinates (volume_constraint scaled about world origin)

### LINKED ISSUE #1070: Smoothing shifts mesh coordinates
I've noticed that mesh smoothing seems to shift the mesh in space:

```
mesh = trimesh.load_mesh('../mesh.stl')
mesh2 = mesh.copy()
mesh2 = trimesh.smoothing.filter_laplacian(mesh2)


scene = trimesh.Scene([mesh,mesh2])
scene.show()
```

Where I would think these meshes should be overlapping. The results are the same for `trimesh.smoothing.filter_taubin()` and `trimesh.smoothing.filter_humphrey()`

Screenshot from 2020-11-30 16-11-36

rootsbymenda:
filter_laplacian's volume_constraint rescaled vertices about the world origin, so any off-origin mesh drifted — the shift reported here. The rescale is now anchored at the mesh center of mass, exactly as suggested in the first comment. Two regression tests included. Behavior on non-watertight meshes stays undefined, as before. Fixes #1070.


mikedh:
Makes sense, thanks for the fix!

PR #2590, opened 2026-08-14
Raise ValueError for a PLY property line before any element

eeshsaxena:
Ran into this feeding some hand-written and partially-corrupt PLY headers through trimesh.load: a header that has a `property` line before any `element` line crashes with `UnboundLocalError: cannot access local variable 'name'` from ply.py rather than a clean error.

In `_parse_header` the `name` variable is only set inside the `element` branch, but the `property` branch immediately does `elements[name][...]`. A well-formed PLY always declares the element first, so in practice `name` is set, but a malformed header where a property comes first leaves it unbound.

I added a guard so a property with no preceding element raises `ValueError("Property defined before any element!")`, matching the `ValueError`s this same function already raises for "Not a ply file!" and "Header not terminated properly!".

Added a test in test_ply.py with a minimal bad header; it raises UnboundLocalError on main and passes with the change, and the rest of test_ply.py still passes. Found it by fuzzing the PLY loader with mutated exports of a valid mesh.

### DIFF
--- a/tests/test_ply.py
+++ b/tests/test_ply.py
@@ -272,6 +272,14 @@ class PlyTest(g.unittest.TestCase):
         p = next(iter(s.geometry.values()))
         assert p.vertices.shape == (1000, 3)
 
+    def test_property_before_element(self):
+        # a PLY header with a property line before any element declaration used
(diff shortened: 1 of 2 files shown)


mikedh:

PR #2593, opened 2026-08-17
Handle rows with zero columns in hashable_rows

eeshsaxena:
`hashable_rows` (and `unique_rows`/`group_rows` through it) divides by the column count in the bitbang path, so passing a 2D array with zero columns raises `ZeroDivisionError`. An array with no columns has every row identical, so I return a constant hash per row in that case, which makes `unique_rows` report one unique row and `group_rows` a single group. Empty arrays and the normal paths are unchanged.

### DIFF
--- a/tests/test_grouping.py
+++ b/tests/test_grouping.py
@@ -22,6 +22,18 @@ class GroupTests(g.unittest.TestCase):
         assert (inverse[:subset] == 0).all()
         assert len(unique) == count - subset + 1
 
+    def test_unique_rows_zero_columns(self):
+        # rows without any columns are all identical; this used to raise a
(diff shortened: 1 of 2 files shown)


eeshsaxena:
The only red check here is `Architecture Tests (s390x + i386)`, which timed out at the 1-hour mark. This looks like flakiness on the emulated runner rather than anything in this PR: that job normally finishes in ~20 min (see #2592, #2590, #2589), and #2591 hit the same 1-hour timeout at the same time. This change only touches `hashable_rows` and isn't architecture- or endianness-specific. Could you re-run the failed job when you have a moment? I don't have permission to rerun it myself. Thanks!

mikedh:
Thanks for the fix and test!

PR #2599, opened 2026-09-09
Close discretized circles exactly

### LINKED ISSUE #2588: 2D paths extrusion fail to produce valid meshes with trimesh >= 4.6.12 and mapbox-earcut == 1.0.1
After updating some python packages on a project, I had a annoying bug where 2d paths extrusions would stop producing valid mesh, even though I did not make any changes to the version of trimesh (4.6.12).

Running the following minimum working example would not produce any errors, however calling `mesh.show()` would produce an empty viewer. Additionally `mesh.is_volume` would return `False`.
``` python
import trimesh
path_2d = trimesh.path.creation.circle(radius=2, segments=8)
extrusion = path_2d.extrude(3)
mesh = extrusion.to_mesh()
```

Upgrading to trimesh 5.0.0 led to the same issue. However, after a non-negligible amount of time trying to debug this, I realized that bumping `mapbox-earcut` to version 2.0.0 would fix the issue.

Having some form of warning or error message would be nice in this instance. But in any case I hope this helps.

Cheers,

Tom

Miko997:
Fixes #2588's supported-Earcut residual: roundoff left circle endpoints unequal, creating collapsed cap and side faces. Snap closed arcs' final point to their first.

Validation: 76 tests passed; repository-wide Ruff and changed-file formatting passed; before/after profiling checked.

Implemented and independently reviewed locally with Codex.


Miko997:
The updated regression tests now pass across all OS/Python jobs and Docker. The architecture job stalled and was cancelled after an hour; the other 11 checks passed. Could you take a look and rerun that job when convenient?

PR #2600, opened 2026-09-10
Fix the angle returned by align_vectors

Doribelove:
Fix incorrect angles using SVD columns; remove extra π. AI-assisted. Tests: 754 passed, 4 skipped, 489 warnings. Ruff and before/after profiling completed.

```python
import numpy as np
from trimesh.geometry import align_vectors
_, angle = align_vectors([1, 0, 0], [-1, 0, 0], return_angle=True)
assert np.isclose(angle, np.pi)
```


mikedh:
Thanks for the fix and test!

Issue #2369, opened 2025-03-03
Rotation of scene not applied


PetitPhoenix:
Hello,

Still working on my stuff, I try to rotate my camera but can't get a hand on the right way to do.

As a base, I use this svg:
![Image](https://github.com/user-attachments/assets/3bf79ccb-9814-414d-965a-b8b50b21822b)

My import works fine:
```
import trimesh

svg_path = 'test_rotate3.svg'
with open(svg_path, "rb") as file:
    path = trimesh.load_path(file, file_type='svg')
print("Path loaded")
mesh = path.extrude(3)
mesh = trimesh.boolean.union(mesh)
print("Mesh extruded")
mesh.show()
```

I need to rotate by y in 180° as svg is the other way around "traditionnal viewing" but when I rotate my camera in either x or y by 180° (I want to see is on the other side, I have the exact same view:
```
import math
scene = trimesh.Scene()
scene.add_geometry(mesh)
corners = scene.bounds
r_e = trimesh.transformations.euler_matrix(
    math.radians(0),
    math.radians(180),
    math.radians(0),
    "ryxz",
)
t_r = scene.camera.look_at(corners, rotation=r_e)
scene.camera_transform = t_r
scene.show()
```

I tried also the set_camera but it has the same behaviour.

Thanks,

PetitPhoenix:
Hello all,

I wish a pleasant Christmas season to all of you :)

I looked at it a bit further and it seems that there is a mysterious 3300 factor in the angles of rotations.

Orginal svg :
![Image](https://github.com/user-attachments/assets/f0021040-506d-4090-8eab-6d38b9837961)

SVG rotated by 90°
![Image](https://github.com/user-attachments/assets/6aec4d56-56a9-49ca-837c-ecb30f129b25)

The 2 first svg imported by Trimesh (4.10.1) and a last one with the mysterious factor which makes a great 90°.
<img width="2250" height="750" alt="Image" src="https://github.com/user-attachments/assets/ca8a224e-4550-40eb-b104-96ebcdcd28b2" />

Here is the code to reproduce:
```
"""
Debug rotation: Test minimaliste pour comprendre comment trimesh interprète les rotations
Usage: Modifier ANGLE_TO_TEST pour tester différents angles
"""
from pathlib import Path
import matplotlib.pyplot as plt
import trimesh

HERE = Path(__file__).resolve().parent
output_path = HERE / "outputs" / Path(__file__).stem
[...]

PR #2602, opened 2026-09-16
Fix SVG rotation angle units and direction

guhou-hvi:
Correct SVG rotation units and direction, including pivots and composition. Addresses #2369's SVG reproduction, not its camera report. All 56 affected-module tests pass; profiling found no material slowdown. Codex wrote and tested this change; a separate Codex agent reviewed it. I reviewed and understand the changes.


mikedh:
Thanks for the fix and test!

PR #2605, opened 2026-09-20
Finish the truncated `percent` docstring in simplify_quadric_decimation

MushiSenpai:
`percent` is forwarded as `target_reduction`, so it is the fraction of faces **removed**, not kept, and the description stops mid-sentence at "for how much".

Measured on an icosphere (5120 faces): `percent=0.9` returns 510 faces, `percent=0.1` returns 4608.

Docstring only, no behaviour change.


mikedh:

